PR #23, opened 2020-03-12
Threatminer Analyzer Module

amanjiofficial:
I have tried to implement Threatminer Analyzer Module for IP/Domain/hash analysis. Please have a look and suggest appropriate changes required.  Thank you @mlodic .

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Main features:
 * Talos Reputation
 * Tor Project
 * Robtex
+* Threatminer
 
 ### Documentation
 [![Documentation Status](https://readthedocs.org/projects/intelowl/badge/?version=latest)](https://intelowl.readthedocs.io/en/latest/?badge=latest)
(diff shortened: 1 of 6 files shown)


mlodic:
Hi @amanjiofficial thank you for your contribution.

2 main things to change:
* The analyzer configuration is wrong. In this way, if you call the ThreatMiner analyzer, you would pass 'rt' as parameter to 'rt'! It still works because the API falls back to rt=1 by default but this is not the wanted behavior. You should create a different analyze configuration for each kind of information you can retrieve from ThreatMiner. Ex: "ThreatMiner_PDNS" needs rt=2, "ThreatMiner_Subdomains" needs rt=5 e so on. Before writing the code, you have to try to understand how the API works, using  a client like Postman for example or, if you are old style, `curl`.
* If you make a mistake, like a typo, please do not push other commits but fix the previous ones on your branch. Otherwise you would poison the history.

amanjiofficial:
Thank you @mlodic. I have noted both the points carefully and will soon do the changes accordingly.

mlodic:
A guide on how to squash commits: https://github.com/wprig/wprig/wiki/How-to-squash-commits.

PR #24, opened 2020-03-13
Added HoneyDB integration for all services.

Eshaan7:
This is a follow-up from https://github.com/certego/IntelOwl/pull/22. The previous branch had become messy so I decided to create a new branch that was even with the original `master` repo and work from that.

Now the HoneyDB integration can be used to call any endpoint from their [list of API services](https://honeydb.io/threats). 
Necessary information has been provided in the documentation and the `env_file_app_template` file.

Please review and suggest any changes, if required.

PS - I have renamed the analyzer `honeydb_twitter_scan` to `honeydb` in the codebase because now it is not only limited to twitter scan.

### DIFF
--- a/api_app/script_analyzers/observable_analyzers/honeydb.py
+++ b/api_app/script_analyzers/observable_analyzers/honeydb.py
@@ -15,10 +15,12 @@ def run(analyzer_name, job_id, observable_name, observable_classification, addit
                 "".format(analyzer_name, job_id, observable_name))
     report = general.get_basic_report_template(analyzer_name)
     try:
-        api_key_name = additional_config_params.get('api_key_name', 'HONEYDB_API_KEY')
-        api_id_name = additional_config_params.get('api_id_name', 'HONEYDB_API_ID')
+        api_key_name = additional_config_params.get("api_key_name", "HONEYDB_API_KEY")
+        api_id_name = additional_config_params.get("api_id_name", "HONEYDB_API_ID")
+        api_endpoint_name = additional_config_params.get("api_endpoint_name", "HONEYDB_API_ENDPOINT")
         api_key = secrets.get_secret(api_key_name)
         api_id = secrets.get_secret(api_id_name)
+        api_endpoint = secrets.get_secret(api_endpoint_name) or "nodes" # specifying `/nodes` as the default endpoint.
         if not api_key:
             raise AnalyzerRunException("no HoneyDB API Key retrieved")
         if not api_id:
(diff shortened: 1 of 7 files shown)


mlodic:
Hi @Eshaan7 the idea is good but out of the scope of the analysis of a single "observable".
Intel Owl main goal is to gather as much information as possible from different sources with a single query for a _specific_ observable or file. The harvesting of data for every HoneyDB API endpoint is useful for generic threat intelligence goals but it is _not_ for the goal of retrieving information about the single observable you analyzed. You would give back unwanted information. This means that you can integrate only the endpoints that actually retrieve information based on the input (the observable) that you give.

For this reason, you should create a single module (as you did) and _different_ configurations for the same module ([analyzer_config.json](https://github.com/certego/IntelOwl/blob/master/configuration/analyzer_config.json)), based on the available endpoints. Yara configuration could be an example of this use case.

Eshaan7:
The module is integrated in the way that at one moment (in a single query) only a single endpoint of HoneyDB is called and that endpoint can be of user's choice which they specify in the environment variable. HoneyDB has only a single endpoint for analyzing an observable (an IP) and that is the `twitter-threat-feed` which is implemented as well. But the other services do not require any observable, they output generic data. For one query/job, it does not call all the endpoints of honeydb but only the one which the user has specified.

mlodic:
Hi @Eshaan7, maybe I did not explain very well. I understood that, with this pull request, you can choose the endpoint and the analyzer does not execute _every_ available query at the same time.

There is another endpoint that can accept an IP as a parameter ([https://honeydb.io/api/lookup/<ip address>](https://honeydb.io/threats#netinfo)). This is why I asked _only_ for this one and not for the others. The problem is that I noticed that I wrote you the wrong link (https://honeydb.io/threats#nodes) that does not accept params instead of the previous one so maybe you got confused.

Also, as you said, the other endpoints output generic data. We do not want those info because they are not exclusively related to the analyzed observable. 

One of the main problems of gathering osint info is that you risk to gather just _noise_. Noise is dangerous because slow down threat intel operations

mlodic:
Another question, how did you manage to get HoneyDB credentials? The site is completely unresponsive and I can not manage to login. I'm starting to think that this project is abandoned

Eshaan7:
I signed up using my GitHub account on: https://honeydb.io/login

mlodic:
thank you, the site was just down at the moment.

Eshaan7:
New PR = #40

PR #26, opened 2020-03-14
threatminer analyzer module

amanjiofficial:
First implementation of Threatminer module. Please have a look and suggest changes @mlodic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Main features:
 * Talos Reputation
 * Tor Project
 * Robtex
+* Threatminer
 
 ### Documentation
 [![Documentation Status](https://readthedocs.org/projects/intelowl/badge/?version=latest)](https://intelowl.readthedocs.io/en/latest/?badge=latest)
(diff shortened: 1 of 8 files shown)


mlodic:
Hi @amanjiofficial, there are other problems:

-  I did not understood why you correctly joined the PDNS and the Reports Tagging in a single analyzer but, then, you created another identical analyzer for Subdomains. Could you please join both python modules in a single one?

- Also, you overwritten a previous configuration (Robtex_Forward_PDNS_Query), de facto disabling that analyzer.

amanjiofficial:
As PDNS and Reports Tagging are defined for two observable_classification(ip and domain) whereas subdomains is defined only for one observable_classification(ip). If you prefer then I will join both python modules in single one.

mlodic:
Nice observation. I found out that there is the same problem for `robtex_ip` and `robtex_rdns` modules. They implement almost the same thing but they splitted in 2 different modules. They should not.

In both cases, in the JSON configuration you can use the same `python_module` but with a different list in `observable_supported`. No need for different modules.

amanjiofficial:
Okay in that case I will create a single python module for all robtex(ip,rdns,fdns) endpoints and a single python module for threatminer(subdomains, report tagging, pdns) endpoints.

mlodic:
That would be awesome, thanks!

amanjiofficial:
I think creating a new branch will be appropriate to avoid conflicts in the code base and ensure maintained code base.

amanjiofficial:
I will fix for both robtex and

PR #28, opened 2020-03-17
Threatminer Analyzer Module

amanjiofficial:
Single python module for all endpoints of threatminer. Please have a look @mlodic .

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Main features:
 * Talos Reputation
 * Tor Project
 * Robtex
+* Threatminer
 
 ### Documentation
 [![Documentation Status](https://readthedocs.org/projects/intelowl/badge/?version=latest)](https://intelowl.readthedocs.io/en/latest/?badge=latest)
--- a/api_app/script_analyzers/observable_analyzers/threatminer.py
+++ b/api_app/script_analyzers/observable_analyzers/threatminer.py
@@ -1,0 +1,73 @@
+import traceback
+
+import requests
+from celery.utils.log import get_task_logger
+
+from api_app.exceptions import AnalyzerRunException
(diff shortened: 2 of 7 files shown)


mlodic:
Thank you.

Please adjust the code here: `threatminer.py`
line 51 -> duplicated and unused code
line 55 -> continuation line under-indented
line 63, line74 -> missing whitespace
line 72 -> rt_value could be not referenced

Please follow PEP8 code style rules. Add a PEP8 evaluator to your own IDE that can assist you in this tasks.

I'll try to update LGTM analysis configuration to let us show these mistakes next times.

mlodic:
Thank you but again,

 line 72 -> rt_value could be not referenced

If an Intel Owl user adds a new ThreatMiner configuration, the rt_value would be undefined. The result would be the crash of the program.

amanjiofficial:
In this PR there are 3 analyzer configurations for Threatminer API and each of them has seperate 'rt' parameter value associated with them. Could you please elaborate the case scenario where the program would crash.

mlodic:
That's exactly the point. You cannot rely on the `analyzer_config.json` because it is a configuration and can be changed by a user. If you have read the documentation (https://intelowl.readthedocs.io/en/latest/Installation.html#analyzers-configuration), an user can change it so, referencing the analyzer names in the module is bad and it has to be changed too.

Also, if an user adds a custom analyzer for this module in the configuration, the program would crash because that `rt_value` is not initialized. And, even if this case has not any chance to happen, it is bad code practice to "rely" on a configuration file in your code. That code is broken by itself because that variable _could_ not have been initialized. It must be changed.

amanjiofficial:
Moved to #48

PR #29, opened 2020-03-17
Robtex Analyzer

amanjiofficial:
Robtex Analyzer integrated in a single python module. Please have a look @mlodic


mlodic:
Same reasons stated here #28.

To be changed:
- Analyzers name used in the code to choose how the module works
- `uri` variable could have not been initialized

Also, the celery task name is different from the `python_module` config. This means that if you choose one of those analyzers via API, it would crash. You would have noticed it if you had at least tried one time those analyzer with the pyintelowl client or if you had run the tests.

You must run the tests.

amanjiofficial:
Moved To #50

Issue #30, opened 2020-03-17
Malware Bazaar API


mlodic:
We should leverage the new service created by Abuse.ch to create an analyzer that checks a file hash against the Malware Bazaar DB.

https://bazaar.abuse.ch/api/#query_hash

Eshaan7:
Please take a look at #32, @mlodic.

Eshaan7:
This issue can be closed now, @mlodic.

PR #32, opened 2020-03-17
Added Malware Bazaar API integration

Eshaan7:
Added integration for Malware Bazaar API. As described in Issue #30.


mlodic:
Thank you Eshaan7.

However there is a better way to develop this integration. Maybe I should note it clearly on the documentation.

In case your observable is a file hash, you should not create a "file analyzer" script because they are meant to be used with files. This means that they have to interact with real files, not hashes.
For this case, you should create an "observable analyzer" and then, add the parameter "run_hash" in the analyzer configuration. In this way you could run that analyzer for two cases:
- if you have the hash but not the sample
- if you have the sample

For a very similar example, you could check Hybrid Analysis module and configuration.
https://github.com/certego/IntelOwl/blob/master/api_app/script_analyzers/observable_analyzers/ha_get.py
https://github.com/certego/IntelOwl/blob/master/configuration/analyzer_config.json#L148
https://github.com/certego/IntelOwl/blob/master/configuration/analyzer_config.json#L195

Eshaan7:
I'll get right on it. Thanks for the review.

Eshaan7:
@mlodic Please check now.

Eshaan7:
@mlodic Please check now.

mlodic:
Thank you @Eshaan7 , the analyze is ok but could you just clean up your commits to a single one? Maybe something went wrong while pulling new updates from the origin.

Eshaan7:
New PR = #45. Thus, closing this.

PR #40, opened 2020-03-18
Added honeydb integration for ip lookup and twitter scan

Eshaan7:
Previous branch got messy and couldn't squash from previous commits. So this is a Follow up from #24. 

HoneyDB Integration has been applied, as discussed: https://github.com/certego/IntelOwl/pull/24#issuecomment-598747672


mlodic:
As stated here #29, we should not use the analyzer names like `HONEYDB_Scan_Twitter` as a conditional clause in the code because the configuration can be changed by the end user. 
You should use a custom variable in `additional_config_params` for this kind of behavior.

More importantly...you did not test the code `observable_name=="HONEYDB_Scan_Twitter"`

Eshaan7:
Moved to #51.

Issue #38, opened 2020-03-18
Incorrect analyzer name passed for HybridAnalysis Observable tests


Eshaan7:
In the [`tests/test_observables.py`](https://github.com/certego/IntelOwl/blob/master/tests/test_observables.py), For HybridAnalysis I noticed that we are passing **`HA_Get`** as the analyzer name in all 3 instances, shouldn't it be **`HybridAnalysis_Get_Observable`**, as defined in the analyzers_configuration.json ? or am I wrong? 

All the 3 instances referenced:
https://github.com/certego/IntelOwl/blob/af05c109e253100ae9a2b0087c96676bdb19dae4/tests/test_observables.py#L93
https://github.com/certego/IntelOwl/blob/af05c109e253100ae9a2b0087c96676bdb19dae4/tests/test_observables.py#L157
https://github.com/certego/IntelOwl/blob/af05c109e253100ae9a2b0087c96676bdb19dae4/tests/test_observables.py#L244

amanjiofficial:
I guess the same is with robtex tests. 

https://github.com/certego/IntelOwl/blob/fab538fb9bf192b62f68b97f6f6ab3e43ecece3f/tests/test_observables.py#L85

https://github.com/certego/IntelOwl/blob/fab538fb9bf192b62f68b97f6f6ab3e43ecece3f/tests/test_observables.py#L89

Please see @Eshaan7

mlodic:
At the moment the tests are just basic as you may have noticed. They just execute the `run` function in each module, without even passing through the `analyzer_config.json`. They just test the modules, they do not test the configuration. This is why tests work correctly.

mlodic:
I think that we could add logging in the tests to print the report results of the analyzers. In this way it would be more clear that the test are working correctly when you run one of them.
At the same time, we should give the chance to remove these logs when are unwanted: for example, when you run _all_ the tests at once.

PR #41, opened 2020-03-22
fix analyzer's names passed in tests/observables

Eshaan7:
Fix the names passed while running tests. As described in Issue #38.
I understand that atm the tests just test the modules as they execute the run function in each module, without even passing through the `analyzer_config.json`. We can work on that in another PR.


mlodic:
Thank you, merged

PR #47, opened 2020-03-23
add censys api integration for viewing by IP

paralax:
built by mimicking what's done in Shodan.


mlodic:
Thank you for your interest!

I had to made different changes because, as you may find in the documentation, the pull requests should be made against the develop branch and there were some errors in the implementation.

So I made the required changes in the develop branch and I pushed there.

paralax:
Thanks! Looks good and you found a bunch it stuff I missed. Thanks for the fast turnaround.

PR #50, opened 2020-03-24
single module for robtex api endpoints

amanjiofficial:
Followed From #29.
Robtex Analyzer integrated in a single python module
The conditions to run different endpoints of API doesn't depend on name of analyzer as suggested by @mlodic. Please review.


mlodic:
Hi, as stated by the bot, please fix the "Potentially uninitialized local variable".

Also, while refactoring the tests, you removed the `robtex_rdns` test....

amanjiofficial:
Since there is a single module for robtex, thus having different test for same observable classification is redundant unless we modify to test for different robtex endpoints. IF you still prefer then I will add test for robtex_rdns.

mlodic:
It is not a matter of preferences, you just removed the test for the `Robtex_Reverse_PDNS_Query` analyzer. Without it we cannot test if that analyzer does work or not. You have to just rename that part, not removing it.

amanjiofficial:
@mlodic suggested changes pushed.

mlodic:
You fixed the mistake correctly but the log message is wrong. You copied and pasted...

PR #51, opened 2020-03-27
HoneyDB integration added for single module

Eshaan7:
Follow up from PR #40.

**Implementation:**
Uses a single module `honeydb.py` with 2 configurations `HoneyDB_Get` and `HoneyDB_Scan_Twitter`. 

**Fixes:**
As [stated](https://github.com/certego/IntelOwl/pull/40#issuecomment-602238102) by @mlodic in the previous PR, I have used a custom variable in `additional_config_params` instead of the analyzer's name in the conditional clause.

**Test Run:**
Test run of both configurations with pyintelowl: https://i.imgur.com/8RFRJU3.jpg


mlodic:
Thank you, merged

Issue #35, opened 2020-03-17
SecurityTrials API Integration


amanjiofficial:
We should add an analyzer to extract info from Securitytrials

https://securitytrails.com/corp/api

Suggestions Please @mlodic.

mlodic:
Yes, it could be useful.

I would focus the effort on the 4 endpoints that can be used with the free license. https://securitytrails.com/corp/pricing

PR #68, opened 2020-04-12
securitytrails module

amanjiofficial:
Added Security Trial Module for IP and domain.
Reference issue : #35
Test results : [test results](https://imgur.com/a/o0BiIVo)
Please suggest changes if required @mlodic.

Security Trial has many endpoints for domain and IP analysis. I have implemented the ones that can be queried with free api. There is an endpoint [historical dns](https://docs.securitytrails.com/reference#history-dns) it has 5 different possible endpoints each for
different dns records `a` `aaaa` `mx` `ns` and others. If we go about adding all of them it would add many analyzer configurations. What is your suggestion on this @mlodic .


mlodic:
Thank you! 

If you can please add just the `A` records query for the historical DNS endpoint

amanjiofficial:
> Thank you!
> 
> If you can please add just the `A` records query for the historical DNS endpoint

It is updated. Please check

PR #75, opened 2020-05-12
Added Tag model and custom fields in admin view

Eshaan7:
Changes:
- Added a `Tag` model in database that has a ManyToMany relationship with `Job` model. This would help users to tag job objects with a label and color of their choice to categorize/group different jobs together.
- Added custom view fields and searchable fields in admin view models for `Job` and `Tag` both.
- Will be adding the option to specify tags via the pyintelowl client too.


Eshaan7:
Closed because of wrong git commit history.

Issue #82, opened 2020-05-13
add indexes for searchable fields + change data retention


Eshaan7:
Matteo Lodi: 
- [x] We could add indexes for the fields that are being used for search purposes.
Obviously, for testing and for a "classic" usage of this project this is not necessary, but to let it scale we should index the most important ones. https://docs.djangoproject.com/en/2.2/ref/models/indexes/

- [x] Also, at the moment there is a cronjob in `crons.py` that performs a cleanup of old data. The retention days are hardcoded. We could move this value to the configuration and let the intel owl users decide the retention days.

Eshaan7:
> there is a cronjob in crons.py that performs a cleanup of old data. The retention days are hardcoded. We could move this value to the configuration and let the intel owl users decide the retention days.

Added in PR - #94.

PR #94, opened 2020-06-19
Using JWT for auth using drf-simplejwt package + retention days to env variable

Eshaan7:
This PR introduces changes required to switch to JSON Web Tokens (JWT) for authentication instead of the default tokens by django-rest-framework.

- Introduces the `django-rest-framework-simplejwt` package. [Docs](https://django-rest-framework-simplejwt.readthedocs.io/).
- simplejwt package by default does not allow delete and create actions on `OutstandingToken` model from the django admin interface. Thus I have overwritten it's admin view implementing [`CustomOutstandingTokenAdmin`](https://github.com/intelowlproject/IntelOwl/pull/94/files#diff-618a96daf5cbbf7bc4d0bfd013ebc217R32) that allows these 2 operations. Related [issue](https://github.com/SimpleJWT/django-rest-framework-simplejwt/issues/258) on their GitHub.
- Refresh-access Token-pair working in brief:

   - Refresh tokens are rotated along with access tokens with lifetimes of,
   ```python
    ...
    "ACCESS_TOKEN_LIFETIME": timedelta(minutes=30),
    "REFRESH_TOKEN_LIFETIME": timedelta(hours=12),
    ...
    ```
   - The refresh token lifetime for tokens meant to be used with pyintelowl (ones created from admin interface and one with `client:pyintelowl` claim in their JWT) is set to 7 days. (can be edited from env variables)
   - Overwritten the `TokenRefreshSerializer` with [`TokenRefreshPatchedSerializer`](https://github.com/intelowlproject/IntelOwl/pull/94/files#diff-cf7d3c525eedcb42adb00bf181a930dbR54) to allow,
       - Refresh token used to obtain a new token-pair is immediately blacklisted. 
       - Stores the new refresh token to the list of Outstanding tokens immediately
       - Related [Issue](https://github.com/SimpleJWT/django-rest-framework-simplejwt/issues/25) on their GitHub.
 
    - [`flush_expired_tokens`](https://github.com/intelowlproject/IntelOwl/pull/94/files#diff-4dc50d190568ac02f59fb35fbfa59925R28): a crontab that runs every 6 hours and deletes expired tokens from `OutstandingToken` model.

- Instead of defining `Authentication` and `Permission` classes decorators on each view explicitly. I have moved it to `settings.py`. [diff](https://github.com/intelowlproject/IntelOwl/pull/94/files#diff-fa4841e2257216d729e889750faad355R80).
- [`api_app/auth.py`](api_app/auth.py) file contains 3 API endpoints, namely: `login`, `refresh-token`, and `logout`. These views have been moved from `api.py` to `auth.py`.
- `get_user_info` view is removed since it's not needed anymore.

PS: also has changes where I have replaced `django.http.JsonResponse` with `rest_framework.response.Response` class since this is [recommended by django-rest-framework](https://www.django-rest-framework.org/api-guide/responses/) and increases consistency throughout the code as well.

P.P.S: I have also added `OLD_JOBS_RETENTION_DAYS` env variable to `env_file_app_template`. For Issue #82)

### DIFF
--- a/api_app/auth.py
+++ b/api_app/auth.py
@@ -1,0 +1,137 @@
+import logging
+
+from django.contrib.auth import authenticate, login, logout
+from rest_framework.response import Response
+from rest_framework import status
+from rest_framework.exceptions import APIException
+from rest_framework_simplejwt.tokens import RefreshToken
+from rest_framework_simplejwt.views import TokenRefreshView
+from rest_framework.decorators import (
+    api_view,
+    authentication_classes,
+    permission_classes,
+)
+
+from api_app import serializers
+
+
+logger = logging.getLogger(__name__)
+
+
+""" Auth API endpoints """
+
+
+@api_view(["POST"])
+@authentication_classes([])
+@permission_classes([])
+def obtain_user_token(request):
+    """
+    REST endpoint to obtain user auth token via authentication
+
+    :param username: string
+        username of registered user
+    :param password: string
+        password of registered user
+
+    :return 202:
+        if accepted
+    :return 404:
+        if failed
+    """
+    try:
+        username = request.data["username"]
+        logger.info(f"obtain_user_token received request for '{username}'.")
+        password = request.data["password"]
+        auth_user = authenticate(username=username, password=password)
+        if auth_user:
+            if auth_user.is_superuser:
+                try:
+                    logger.info(f"headers: {request.headers}")
+                    # pass admin user's session
+                    login(request, auth_user)
+                    logger.info(f"administrator:'{username}' was logged in.")
+                except Exception:
+                    logger.exception(f"administrator:'{username}' login failed.")
+
+            refresh = RefreshToken.for_user(auth_user)
+            logger.debug(f"obtain_user_token: token created for '{username}'.")
+            # adding custom 'username' claim
+            refresh["username"] = auth_user.username
+            return Response(
+                {"refresh": str(refresh), "access": str(refresh.access_token)},
+                status=status.HTTP_202_ACCEPTED,
+            )
+        raise APIException(
+            "No such user exists or Incorrect credentials",
+            code=status.HTTP_400_BAD_REQUEST,
+        )
+
+    except APIException as e:
+        logger.exception(f"obtain_user_token exception: {e}")
+        return Response({"error": str(e)}, status=e.status_code)
+
+
+class CustomTokenRefreshView(TokenRefreshView):
+    """
+    REST endpoint that returns new `token` object consisting of
+    `access` and `refresh` fields
+    given a valid `refresh` token.
+
+    :methods_allowed:
+        POST
+
+    :param request.data:
+        JSON[refresh]
+
+    :return 201:
+        if accepted and created new token
+    :return 400:
+        if failed
+    """
+
+    serializer_class = serializers.TokenRefreshPatchedSerializer
+
+
+@api_view(["POST"])
+def perform_logout(request):
+    """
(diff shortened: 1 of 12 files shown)


api_app/auth.py
```diff
@@ -0,0 +1,128 @@
...
+    :return 404:
+        if failed
+    """
+    try:
+        username = request.data["username"]
+        logger.info(f"obtain_user_token received request for '{username}'.")
+        password = request.data["password"]
+        auth_user = authenticate(username=username, password=password)
+        if auth_user:
+            # ---- Todo ----
```
Eshaan7, line 47: This change is yet to finalize.

api_app/auth.py
```diff
@@ -0,0 +1,128 @@
...
+        if failed
+    """
+    try:
+        logger.info(f"perform_logout received request from {str(request.user)}")
+        recvd_refresh_token = request.data["refresh"]
+        refresh = RefreshToken(recvd_refresh_token)
+        try:
+            # ---- TODO ----
+            # logout user, this is only meant for the admin user.
+            logout(request)
```
Eshaan7, line 113: same as above.

api_app/auth.py
```diff
@@ -0,0 +1,128 @@
...
+            refresh["username"] = auth_user.username
+            return Response(
+                {"refresh": str(refresh), "access": str(refresh.access_token)},
+                status=status.HTTP_202_ACCEPTED,
+            )
+        raise APIException("No such user exists or Incorrect credentials")
+
+    except APIException as e:
+        logger.exception(f"obtain_user_token exception: {e}.")
+        return Response({"error": e}, status=status.HTTP_404_NOT_FOUND)
```
mlodic, line 68: if APIException triggers, then it would not be JSONSerializable and would trigger a 500 response. By casting the error as string maybe we solve this problem. In any case, we should add a more generic Exception clause to catch the unexpected. (However I noticed that the error ("No such user exists or Incorrect credentials") is not correctly shown by the client even with the previous adjustment)

api_app/auth.py
mlodic: for me this is the best for the admin user experience

api_app/auth.py
```diff
@@ -0,0 +1,128 @@
...
+            # ---- Todo ----
+            # should or should not login the admin ?
+            # verify with user-agent ?
+            logger.info(f"headers: {request.headers}")
+            if auth_user.is_superuser:
+                logger.info(f"administrator:'{username}' was logged in.")
+                login(request, auth_user)
+            # ---------------------------
+
+            logger.debug(f"obtain_user_token token created for '{username}'.")
```
mlodic, line 56: maybe should this log be one line below?

api_app/auth.py
```diff
@@ -0,0 +1,128 @@
...
+        recvd_refresh_token = request.data["refresh"]
+        refresh = RefreshToken(recvd_refresh_token)
+        try:
+            # ---- TODO ----
+            # logout user, this is only meant for the admin user.
+            logout(request)
+            # --------------
+            # Attempt to blacklist the given refresh token
+            refresh.blacklist()
+        except AttributeError:
```
mlodic, line 117: could you add a log with the same info you wrote as a comment?

Eshaan7:
Update:
- user can specify pyintelowl's token lifetime in `env_file_app_template`.
- JSON serializable error fixed.
- logging added as requested in review.
- added `OLD_JOBS_RETENTION_DAYS` env variable to `env_file_app_template`. (For Issue #82)

mlodic:
Thank you again! Merged!

PR #103, opened 2020-07-01
modified PEframe to use the new flask-shell2http package

Eshaan7:
1. modified PEframe to use the new [flask-shell2http](https://github.com/Eshaan7/Flask-Shell2HTTP) package.
2. I have modified the testing suite to reduce some redundant code duplicacy.
   - Same `MockResponse` class was earlier defined at many places, I have moved it to the `test_api.py` and imported in the 
`test_files.py` and `test_observables.py`. 
   - The other mock response class has that does nothing been renamed to `MockResponseNoOp` to specify no-operation. This is also now defined in `test_api.py` and imported into `test_observable.py`.

### DIFF
--- a/api_app/script_analyzers/file_analyzers/peframe.py
+++ b/api_app/script_analyzers/file_analyzers/peframe.py
@@ -1,5 +1,6 @@
 import requests
 import traceback
+import json
 import logging
 import time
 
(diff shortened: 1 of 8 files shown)


mlodic:
good job! Thank you!

PR #155, opened 2020-08-19
Fix Thug never getting past the is_test check

colbyprior:
- The ThugUrl class would set the `is_test` argument to either True or False.
- The `_docker_run` function was only checking whether argument `is_test` is set or not.
- This resulted in the Thug analysers never being run, they always return the empty json array.

I added a simple merge request to also check to see if the `is_test` argument is set to True and it fixed the issue.


mlodic:
Nice catch! Thank you! This will be shipped on the next release number!

PR #159, opened 2020-08-20
[Help required] Added IntelX.io observable source

ideaengine007:
Hi,

I added the IntelX source for getting intel on observables like emails, domains, IPs, phone numbers. I am facing an issue when I am  trying to fetch results using IntelX.

I set the API key and tested the logic of intelx.py and I can verify that the request is happening as the Developer dashboard shows a reduction of available queries (eg. 1000 queries available initially. After running the query, the count reduces to 999, i.e., the API call is successful but the response is throwing a 500 response code.

The flow is as follows:

POST the search term
Response of POST request returns the 'id' (uuid) associated with the search term
GET the results by using the uuid returned in the above POST request

Initially, I thought that some time is required for the query to complete. The response of POST request might take some time in returning the UUID, but even after doing a `sleep`, the issue persists.


ideaengine007:
The json file and all other edits were not in the request.

PR #168, opened 2020-08-23
Urlscan

RamnathKumar181:


api_app/script_analyzers/observable_analyzers/urlscan.py
```diff
@@ -0,0 +1,81 @@
...
+        api_key_name = additional_config_params.get("api_key_name", "URLSCAN_KEY")
+        self.visibility = additional_config_params.get("visibility", "public")
+        self.__api_key = secrets.get_secret(api_key_name)
+
+    def __submit_query(self):
+        data = {"url": self.observable_name, "visibility": self.visibility}
+        uri = "api/v1/scan/"
+        try:
+            response = requests.post(
+                self.base_url + uri, headers=self.headers, data=json.dumps(data)
```
mlodic, line 23: It is better to use the `json` parameter instead of `data`. Check the docs in `requests`

api_app/script_analyzers/observable_analyzers/urlscan.py
```diff
@@ -0,0 +1,81 @@
...
+            except requests.RequestException as e:
+                raise AnalyzerRunException(e)
+
+        elif self.analysis_type == "submit_result":
+            token = self.__submit_query()
+            time.sleep(10)
+            result = self.__poll_for_result(token, self.headers)
+        else:
+            raise AnalyzerRunException(
+                "not supported observable"
```
mlodic, line 76: the message is wrong and come from a copy and paste...

api_app/script_analyzers/observable_analyzers/urlscan.py
```diff
@@ -0,0 +1,81 @@
...
+    def __poll_for_result(self, url, headers):
+        max_tries: int = 10
+        poll_distance: int = 2
+        result = {}
+        for chance in range(max_tries):
+            time.sleep(poll_distance)
+            resp = requests.get(url, headers=headers)
+            resp_json = resp.json()
+            status = resp_json.get("status", None)
+            if status == "done":
```
mlodic, line 39: as stated in URLScan docs, I think you should make this check based on the status code and not from the keyword "status"

`The response to the API call will give you the scan ID and API endpoint for the scan, you can use it to retrieve the result after waiting for a short while. Until the scan is finished, the URL will respond with a HTTP 404 status code.`

tests/test_observables.py
```diff
@@ -478,6 +509,16 @@ def test_mb_get(self, mock_get=None, mock_post=None):
         ).start()
         self.assertEqual(report.get("success", False), True)
 
+    def test_urlscan_hash(self, mock_get=None, mock_post=None):
```
mlodic, line 512: about tests, this a good step forward. But there are 2 problems:

1 - you have to test **both** the analyzers you created. In this way you are just testing the first analyzer. (also the analyzer name is wrong: "Urlscan" != "UrlScan_Search")

2 - "UrlScan_Submit_Result" supports only urls so its testing suite can be put here, in the URL section. Otherwise "UrlScan_Search" supports all the 4 types of observables and, like you did, you are duplicating the code. You should put the testing suite of this analyzer here https://github.com/intelowlproject/IntelOwl/blob/master/tests/utils.py#L25. In this way you just write the test one time but it will be applied for all the observables types

api_app/script_analyzers/observable_analyzers/urlscan.py
```diff
@@ -0,0 +1,81 @@
...
+from api_app.script_analyzers import classes
+from intel_owl import secrets
+
+
+class UrlScan(classes.ObservableAnalyzer):
+    base_url: str = "https://urlscan.io/"
+
+    def set_config(self, additional_config_params):
+        self.analysis_type = additional_config_params.get("urlscan_analysis", "search")
+        api_key_name = additional_config_params.get("api_key_name", "URLSCAN_KEY")
```
mlodic, line 14: URLSCAN_KEY still missing in the env_file_template

mlodic:
Again, please read the docs before sending PR. https://intelowl.readthedocs.io/en/latest/Contribute.html#create-a-pull-request 

You have to squash all commits into a single one. Commits like "fixed" or "merge" are not allowed.

api_app/script_analyzers/observable_analyzers/urlscan.py
RamnathKumar181: Sorry about this! I keep changing it in the env_file_app but forget to do the same in the template. Have fixed this and pushed again!

api_app/script_analyzers/observable_analyzers/urlscan.py
RamnathKumar181: Have changed the error message. Please let me know if this is fine.

api_app/script_analyzers/observable_analyzers/urlscan.py
RamnathKumar181: Change `data=json.dumps(data)` to `json=data`. Thanks!

api_app/script_analyzers/observable_analyzers/urlscan.py
RamnathKumar181: Changed poll_for_result. It follows the procedure mentioned in URLScan docs.

tests/test_observables.py
RamnathKumar181: Hey @mlodic , I have changed the tests. The file is able to pass the tests if I run the tests by passing the test parameters as a standalone application. I am still not able to debug why it's failing in the Travis build. Could you please help me with this. Sorry for bothering you!

mlodic:
Wait please @RamnathKumar181. You should not push and push again in the branch where the PR is opened, otherwise your changes will be continuously reflected yo the PR. While you are doing development changes, you should create another branch in your fork and push there the changes. Then, you should only push into `RamnathKumar181:urlscan` (this branch) only when your work is finished. One single time. Otherwise you are spamming. Thanks

api_app/script_analyzers/observable_analyzers/urlscan.py
```diff
@@ -0,0 +1,80 @@
...
+    def __poll_for_result(self, url, headers):
+        max_tries: int = 10
+        poll_distance: int = 2
+        result = {}
+        for chance in range(max_tries):
+            time.sleep(poll_distance)
+            resp = requests.get(url, headers=headers)
+            resp_json = resp.json()
+            status = resp_json.get("status", None)
+            if status == 404:
```
mlodic, line 38: this is still wrong. you have to check for the status code, not the "status" keyword.

Eshaan7: Status code is accessible via `resp.status_code`

Eshaan7:
@RamnathKumar181 any update on this ? Are you going to continue developing this ?

Issue #174, opened 2020-08-28
Integrate FireEye SpeakEasy


mlodic:
New tool: https://www.fireeye.com/blog/threat-research/2020/08/emulation-of-malicious-shellcode-with-speakeasy.html

PR #177, opened 2020-08-31
Added spekeasy as file analyzer

0ssigeno:
Solves issue #174

### DIFF
--- a/intel_owl/tasks.py
+++ b/intel_owl/tasks.py
@@ -14,6 +14,7 @@ from api_app.script_analyzers.file_analyzers import (
     strings_info,
     rtf_info,
     signature_info,
+    speakeasy_emulation,
     peframe,
     thug_file,
     capa_info,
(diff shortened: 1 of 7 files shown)


Eshaan7:
Thankyou for your contribution. LGTM.

requirements.txt
```diff
@@ -79,4 +79,5 @@ djangorestframework-guardian==0.3.0
 flake8==3.8.2 
 black==19.10b0
 quark-engine==20.8
-git+git://github.com/DissectMalware/XLMMacroDeobfuscator.git@c78b9e443e2667399470462fd863ebe4e2b8c978
\ No newline at end of file
+git+git://github.com/DissectMalware/XLMMacroDeobfuscator.git@c78b9e443e2667399470462fd863ebe4e2b8c978
+git+git://github.com/fireeye/speakeasy.git@e5f7bcf0f1ddf10614f32e76b5d322a5387185db
```
Eshaan7, line 83: Is there a particular reason you used the GitHub url ? [Speakeasy is available on PyPi](https://pypi.org/project/speakeasy-emulator/) and it would be better to grab it from there.

intel_owl/tasks.py
```diff
@@ -581,6 +582,15 @@ def signatureinfo_run(
     ).start()
 
 
+@shared_task(soft_time_limit=40)
```
Eshaan7, line 585: I see that you've kept a lower time limit and added a catch block for `SoftTimeLimitExceeded`. I haven't used Speakeasy myself, but I think in certain cases the analysis might take a few minutes. So maybe we should change this to around 120.

requirements.txt
0ssigeno: To be honest I didn't know it was provided in pypi. If you think is a better solution, will make this change.

intel_owl/tasks.py
0ssigeno: I don't think that speakeasy, with the number of api that have right now, will ever need more than 40 second. But yeah, increasing has no cons, so I will update it.

Issue #145, opened 2020-08-13
Checkdmarc analyzer


mlodic:
We could leverage this tool to extract SPF and DMARC records for a domain:

https://pypi.org/project/checkdmarc/

PR #187, opened 2020-09-09
added checkdmarc analyzer

goodlandsecurity:
#145


goodlandsecurity:
i think i need some help with why the checks are failing as they don't appear to be from the files I added/edited. 

i made sure to follow the guide here: https://intelowl.readthedocs.io/en/latest/Contribute.html#create-a-pull-request

Eshaan7:
@goodlandsecurity Thank you for your contribution. It's only a linter check that's failing. Please fix the merge conflicts and then run this `black . --exclude "migrations|venv"` command.

api_app/script_analyzers/observable_analyzers/checkdmarc.py
```diff
@@ -0,0 +1,37 @@
...
+
+            dmarc_info = stdout.decode("utf-8"), stderr
+
+            dmarc_str = dmarc_info[0]
+
+            dmarc_json = json.loads(dmarc_str)
+
+            return dmarc_json
+
+        except OSError as e:
```
Eshaan7, line 31: The try-catch block here is redundant. The `run` function is itself wrapped into a try-catch block inside the `ObservableAnalyzer` class.

api_app/script_analyzers/observable_analyzers/checkdmarc.py
```diff
@@ -0,0 +1,37 @@
...
+        try:
+            process = subprocess.Popen(
+                [self.check_command, self.observable_name],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            )
+            process.wait()
+            stdout, stderr = process.communicate()
+
+            dmarc_info = stdout.decode("utf-8"), stderr
```
Eshaan7, line 23: A better way would be,
```
if stderr:
  raise Exception(stderr.decode("utf-8"))
dmarc_str = stdout.decode("utf-8")
```

Eshaan7 (changes requested):
Really glad you followed all steps as given in the docs. Just a few things I'd like to change..

goodlandsecurity:
Of course! This is only the second project I've gotten to contribute to, but I am hopeful to add a few more analyzers though! I removed the try-catch block and used shutil to check if the required binary exists before running the analyzer

Eshaan7 (approved):

Eshaan7:
Thank you for your contribution! The checkdmarc analyzer will be available with the next release which we will drop soon.

Issue #186, opened 2020-09-08
Unpacme analyzer


mlodic:
Reference: https://api.unpac.me

PR #191, opened 2020-09-15
UnpacMe Api implementation

0ssigeno:
#186


api_app/script_analyzers/file_analyzers/unpac_me.py
```diff
@@ -0,0 +1,100 @@
...
+            )
+            status = self._get_status(unpac_id)
+            if status == "fail":
+                logger.error(f"md5 {self.md5} job {self.job_id} analysis has failed")
+                raise AnalyzerRunException("failed analysis")
+            if status != "complete":
+                logger.info(
+                    f"md5 {self.md5} job {self.job_id} id {unpac_id} status {status}"
+                )
+                continue
```
Eshaan7, line 49: Unnecessary 'continue' statement

api_app/script_analyzers/file_analyzers/unpac_me.py
```diff
@@ -0,0 +1,100 @@
...
+            raise AnalyzerRunException(
+                f"md5 {self.md5} job {self.job_id} function upload id not in response"
+            )
+        return response["id"]
+
+    def _get_status(self, unpac_me_id) -> str:
+        response = self._req_with_checks(f"{self.private}/status/{unpac_me_id}")
+        return response.json().get("status", False)
+
+    def _get_report(self, unpac_me_id) -> dict:
```
Eshaan7, line 98: This should be `-> "Dict"` or we can do `from typing import Dict` and then `-> Dict`

env_file_app_travis
```diff
@@ -27,6 +27,7 @@ CENSYS_API_SECRET=test
 ONYPHE_KEY=test
 GREYNOISE_API_KEY=test
 INTELX_API_KEY=test
+UNPACK_ME_KEY=test
```
Eshaan7, line 30: This should be `UNPAC_ME_API_KEY`

tests/test_files.py
```diff
@@ -186,6 +187,25 @@ def test_yara_exe(self):
         ).start()
         self.assertEqual(report.get("success", False), True)
 
+        def test_unpackme_exe(self):
+            additional_params = {
```
Eshaan7, line 191: `additional_params` is not needed here

configuration/analyzer_config.json
```diff
@@ -786,6 +786,17 @@
     "description": "check if a domain is in the last Tranco ranking top sites list",
     "python_module": "tor_run"
   },
+  "Unpac_Me": {
+    "type": "file",
+    "supported_filetypes": ["application/x-dosexec"],
+    "description": "UnpacMe unpacker",
+    "python_module": "unpac_me_run",
+    "additional_config_params": {
+      "api_key_name": "UNPAC_ME_KEY",
```
Eshaan7, line 795: This should be `UNPAC_ME_API_KEY`

api_app/script_analyzers/file_analyzers/unpac_me.py
```diff
@@ -0,0 +1,100 @@
...
+        # interval b/w HTTP requests when polling
+        self.poll_distance = 5
+
+    def run(self):
+        if not self.__api_key:
+            raise AnalyzerRunException(
+                f"No API key retrieved with name: {self.api_key_name}"
+            )
+        API_KEY_STRING = "Key %s" % self.__api_key
+        self.headers = {"Authorization": API_KEY_STRING}
```
Eshaan7, line 32: Just change this to `self.headers = {"Authorization": "Key %s" % self.__api_key}` for simplicity

api_app/script_analyzers/file_analyzers/unpac_me.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def set_config(self, additional_config_params):
+        self.api_key_name = additional_config_params.get(
+            "api_key_name", "UNPAC_ME_API_KEY"
+        )
+        private = additional_config_params.get("private", False)
+        self.private = "private" if private else "public"
+        self.__api_key = secrets.get_secret(self.api_key_name)
+        # max no. of tries when polling for result
+        self.max_tries = additional_config_params.get("max_tries", 100)
```
Eshaan7, line 22: 100 max_tries is too much. Should change this to around 30.

Eshaan7 (changes requested):
Thanks for your contribution! Just needs a few changes/fixes.

EDIT: Also I changed the target branch to `develop` which has latest changes. So you may need to resolve merge conflicts.

api_app/script_analyzers/file_analyzers/unpac_me.py
0ssigeno: The continue is required.
Othewise we will try to retrieve the value from a report that is not already present.
Is not a pass.

api_app/script_analyzers/file_analyzers/unpac_me.py
0ssigeno: Why is better to use
 
```
from typing import Dict
```
?
Is not the same at the end?

Eshaan7:
Thanks for the quick changes. The travis CI build is failing. Please see the docs on [testing](https://intelowl.readthedocs.io/en/latest/Contribute.html#pass-linting-and-tests). Briefly,
- You need to run the psf/black formatter. See [here](https://github.com/intelowlproject/IntelOwl/pull/187#issuecomment-690601198).
- The test case for unpacme is failing. You will need to debug this on the local to see at which point it's failing. In the `env_file_app_travis`, make `DISABLE_LOGGING_TEST=False` to see the error logs.

tests/test_files.py
```diff
@@ -194,15 +202,11 @@ def test_yara_exe(self):
         ).start()
         self.assertEqual(report.get("success", False), True)
 
-    def test_unpackme_exe(self):
-
+    @mock_connections(patch("requests.get", side_effect=mocked_unpacme_get()))
+    @mock_connections(patch("requests.post", side_effect=mocked_unpacme_post()))
```
Eshaan7, line 206: remove the (), it should just be `side_effect=mocked_unpacme_post)`

tests/test_files.py
```diff
@@ -194,15 +202,11 @@ def test_yara_exe(self):
         ).start()
         self.assertEqual(report.get("success", False), True)
 
-    def test_unpackme_exe(self):
-
+    @mock_connections(patch("requests.get", side_effect=mocked_unpacme_get()))
```
Eshaan7, line 205: same here

0ssigeno:
Unfortunely my local test does not report the same errors as travis, so it will take more time to debug this.
I'm sorry.

api_app/script_analyzers/file_analyzers/unpac_me.py
```diff
@@ -0,0 +1,99 @@
...
+        logger.info(f"md5 {self.md5} job {self.job_id} uploaded id {unpac_id}")
+        for chance in range(self.max_tries):
+            time.sleep(self.poll_distance)
+            logger.info(
+                f"unpacme polling, try n.{chance + 1}."
+                f" job_id {self.job_id}. starting the query"
+            )
+            status = self._get_status(unpac_id)
+            if status == "fail":
+                logger.error(f"md5 {self.md5} job {self.job_id} analysis has failed")
```
Eshaan7, line 43: redundant `logger.error` line. Same error is already logged inside AnalyzerRunException.

api_app/script_analyzers/file_analyzers/unpac_me.py
```diff
@@ -0,0 +1,99 @@
...
+                raise AnalyzerRunException("failed analysis")
+            if status != "complete":
```
Eshaan7, line 45: Here, the loop does not stop even if we get status="continue".
You should check as `elif status == "complete"` and then put a `break` inside, otherwise it will keep continuing on its own.

Eshaan7: This could probably be the reason why the tests are failing.

0ssigeno: Mhh no.
If fail is returned, we have an error, and the exception is raised.
If something that is not COMPLETE or FAIL, it means that is still unpacking, and the cycle continue.
If COMPLETE is retrieved, the two "if" are not triggered and the return is called.

api_app/script_analyzers/file_analyzers/unpac_me.py
0ssigeno: Yeah, you are right

api_app/script_analyzers/file_analyzers/unpac_me.py
0ssigeno: This code on the local version of IntelOwl works, and an output is retrieved correctly. Is just travis ¯\_(ツ)_/¯

Eshaan7 (approved):

env_file_app_travis
```diff
@@ -28,6 +38,7 @@ ONYPHE_KEY=test
 GREYNOISE_API_KEY=test
 INTELX_API_KEY=test
 WHOISXMLAPI_KEY=test
+UNPAC_ME_API_KEY=test
```
Eshaan7, line 41: There was a typo in this name and that's why the travis build was failing.

Eshaan7:
Once again, thank you for your contribution! :D

env_file_app_travis
0ssigeno: I was never, ever, going to find this. Thank you.

PR #196, opened 2020-09-24
Xlm Macro deobfuscation

0ssigeno:
Unfortunately there is a malware campaign in Italy that is using xls obfuscated macro. 
The last version of https://github.com/DissectMalware/XLMMacroDeobfuscator, is able to deobfuscate the payload, allowing to correctly identify IOC.
image


requirements.txt
```diff
@@ -80,5 +80,5 @@ djangorestframework-guardian==0.3.0
 flake8==3.8.2 
 black==19.10b0
 quark-engine==20.8
-git+git://github.com/DissectMalware/XLMMacroDeobfuscator.git@c78b9e443e2667399470462fd863ebe4e2b8c978
+git+git://github.com/DissectMalware/XLMMacroDeobfuscator.git@89fbce0c87014a4b5a22c1aef09c8b3ea9bf16c0
```
Eshaan7, line 83: Can you replace this with the pip module ? https://pypi.org/project/XLMMacroDeobfuscator/

api_app/script_analyzers/file_analyzers/xlm_macro_deobfuscator.py
```diff
@@ -0,0 +1,62 @@
...
+    def decrypt(self, xlmpassword=""):
+        cmd = [
+            "xlmdeobfuscator",
+            "--file",
+            self.filepath,
+            "--export-json",
+            "result.json",
+        ]
+        pwd_cmd = ["--password", xlmpassword] if xlmpassword else []
+        subprocess.run(cmd + pwd_cmd, stdout=subprocess.PIPE).stdout.decode("utf-8")
```
Eshaan7, line 54: We don't need to use subprocess here. We can use it as a python module. Please see https://github.com/DissectMalware/XLMMacroDeobfuscator#library

Eshaan7 (changes requested):

requirements.txt
0ssigeno: Unfortunely the pip version is 0.1.4, and to deobfuscate the latest macro, 0.1.5 is required

mlodic: Yes, XLMMacroDeobfuscator is a very useful too but very experimental so we should try to get the latest version that we were able to properly test. That commit is ok imho and solves the problem that @0ssigeno stated

api_app/script_analyzers/file_analyzers/xlm_macro_deobfuscator.py
0ssigeno: Yeah, you are right

api_app/script_analyzers/file_analyzers/xlm_macro_deobfuscator.py
```diff
@@ -0,0 +1,56 @@
...
+        results = {}
+        try:
+            for password in self.passwords_to_check:
+                results = self.decrypt(password)
+                if results:
+                    break
+            if not results:
+                results["error"] = "Can't decrypt with current passwords"
+        except SoftTimeLimitExceeded as e:
+            error_message = (
```
Eshaan7, line 30: I missed this before. But if it's not too much trouble, just one more change.
Line 30-36 can be replaced with `self._handle_base_exception("Soft Time Limit Exceeded error")`

Eshaan7: Awesome. Thanks!

Eshaan7:
Thankyou for your contribution!

PR #204, opened 2020-10-03
[WIP] generic celery task for all analyzer modules

Eshaan7:
- [x] need to replace `python_module` attr for each entry in `analyzer_config.json`.
- [x] update contributing docs
- [x] move `soft_time_limit` param to entries in `analyzer_config.json`. (PR #210)
- [ ] add a note in release about retro-compatibility.
- [x] add a default `soft_time_limit` value as well.

recommended: Squash and merge


mlodic:
smart idea to simplify the code! :)  I just wanted to remember to update "Contribute" docs at the end :P

mlodic:
I was thinking about retrocompatibility. Some people could have created a custom `analyzer_config.json` with maybe different analyzers, customized on their needs. By adding `soft_time_limit` in the configuration we give people the power to change that limit without affecting the python code, but, at the same time, it could break forks or customizations. Is it possible to add a default value of `soft_time_limit` in the code? In this way, if that new configuration is missing, the analyzers would still work.

Eshaan7:
>  Is it possible to add a default value of soft_time_limit in the code? In this way, if that new configuration is missing, the analyzers would still work ? 

Yeah, correct. I'll add a default value for the same. Also, will add a note in the next release coz there's a possibility that when users do `git pull` it may show merge conflicts if they have customized the analyzer_config file.

PR #206, opened 2020-10-05
little changes

mlodic:
* adjusts to urlscan analyzer
* removed unecessary `docker-compose-for-tests-no-angular.yml`
* tweaks in docs
* added `Yara_Scan_Stratosphere` to leverage these [rules](https://github.com/stratosphereips/yara-rules)
* upgraded Speakeasy to last version


env_file_app_travis
```diff
@@ -39,7 +39,7 @@ GREYNOISE_API_KEY=test
 INTELX_API_KEY=test
 WHOISXMLAPI_KEY=test
 UNPAC_ME_API_KEY=test
-URLSCAN_API_KEY=
+URLSCAN_API_KEY=test
```
Eshaan7, line 42: this should be left blank for the URLscan test to complete successfuly

mlodic: ah, why?

Eshaan7: The `URLScan_Search` analyzer works without API key but the submit_result requires API key. If we give dummy api key, then search one fails

Eshaan7: I commented the `submit_result` variant of URLscan for this reason only and that's why the test is failing ig

mlodic: I think that this is because it is not mocked `requests.Session.get`

mlodic: yes, that was the cause. Now it works.

Issue #212, opened 2020-10-09
Simplifying CONTRIBUTIONS to IntelOwl


NerdSec:
Hi Team,

First of all, great work on putting this out. I have just started to test `IntelOwl` and it looks promising. Have managed to do something similar previously, albeit without a UI. Would love to contribute to this repo and improve it, :small_airplane: 

I just have two suggestions:

1. Simplify the directory structure
a. Currently too many files are cluttered into the main repo and it is difficult to understand which of the folders are actually django modules.
b. I think it makes sense to separate configurations, dockerfiles, and the actual code in a much more cleaner layout.
2. Avoid using root user in the `intelowl` images.
a. Adding a non-privileged user and running all the processes with that user should help reducing risk.

There are some other suggestions which I would prefer to raise as PR's but I will wait before I create a branch.

mlodic:
Hi,

thank you for your interest in the project!

About your suggestions, feel free to open a PR with those improvements. 

We are here to help you if necessary.

NerdSec:
Sure @mlodic 

Could not identify the use of the following folders:
1. [integrations](https://github.com/NerdSec/IntelOwl/tree/develop/integrations)
2. [configuration](https://github.com/NerdSec/IntelOwl/tree/develop/configuration)

What are these used for or by? Have opened a draft PR. Could you please take a look and let me know your thoughts? :slightly_smiling_face:

Eshaan7:
- The `integrations/` folder containes code, dockerfiles and compose files for optional analyzers which run in their own container. Kind of a micro-services pattern.
- The `configuration/` folder contains files which are mounted as docker volumes to the `uwsgi` docker service, they include configurable settings.

Eshaan7:
> Simplify the directory structure

Done in: #277, #98

PR #215, opened 2020-10-11
Modifying the directory structure for extensibility

NerdSec:
## Motivation

Trying to segregate the files being used so that it is easier to contribute, extend & maintain in the long run. The directory structure being used is inspired from coolacid's [MISP](https://github.com/coolacid/docker-misp) repo.

## Checks

- [x] Move the docker contents to a separate folder
- [x] Move the compose files in a single folder
- [ ] Try to use a single compose file to build all images if feasible by using [contexts](https://github.com/coolacid/docker-misp/blob/master/build-docker-compose.yml)
- [ ] Modify the dotfiles in the project repo to account for the structure change
- [x] Modified the `Dockerfile` to run as the `www-data` user
- [ ] Identify if the [dockerentrypoint.sh](https://github.com/intelowlproject/IntelOwl/blob/master/docker_entrypoint.sh) can be added to the image itself

Targeting #212


Eshaan7 (commented):
I think it's better to rename the files like `docker/compose/tests.yml`, docker/compose/travis.yml`, docker/compose/traefik.yml`, and the generic one could be `prod.yml`

Eshaan7 (commented):
`docker/compose/**docker_entrypoint.sh**` could be named `**entrypoint.sh**`

Eshaan7:
- I think we should not move `intel_owl` directory to under `docker/` directly. We should use the `docker/` directory to store compose files, Dockerfile(s) but not the code.
- `docker/compose/docker_entrypoint.sh` could be named **`entrypoint.sh`**
- better to rename the files like `docker/compose/tests.yml`, docker/compose/travis.yml`, docker/compose/traefik.yml`, and the generic one could be `prod.yml`

mlodic:
I would just like to underline that the rework should bring a real and strong improvement. Otherwise there are only cons in changing all the structure:
* we would lose all the files history...that is pretty bad in my opinion
* some forks could break really bad when they would pull the changes and we could bring more disappointment than improvement.

NerdSec:
I agree with the suggestions from both @Eshaan7 and @mlodic.

Maybe moving the intel_owl directories may not make sense, especially for forks. The problem is, right now it is difficult to grasp that the Django App consists of two modules which are `api_app` and `intel_owl`. Granted, that a little poking around could get you there, but it is not easy to get to. Secondly, the docker-compose files and their requirements are cluttering the root directory structure.

I plan to contribute the swarm and kubernetes helm chart for deploying this in prod, how do we accommodate that in the current directory structure? Adding extra folders to the repo, makes it seem as if it is all the part of the Django app.

In the long term, would it be logical to ship this as a module as well? Maybe for folks with an existing Django Celery deployments? In such a scenario, there does not seem to be any structure to the code base. Please feel free to point out something that did not make sense or you feel is incorrect. Happy to discuss! :slightly_smiling_face:

NerdSec:
Given that, this is a breaking and major change, which I do not deny. However, we could work out at least work out a structure or decide if we would probably move some things to a separate repo or maybe change a few things here itself.

NerdSec:
@Eshaan7  Fair enough, will work on this over the weekend!

mlodic:
I am about to close this because it has become obsolete.

PR #219, opened 2020-10-12
Added phishtank

RamnathKumar181:
Added phishtank api: https://phishtank.org/api_info.php

Could you please take a look, @mlodic @Eshaan7


docs/source/conf.py
```diff
@@ -13,7 +13,7 @@
 # import os
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
-import sphinx_rtd_theme
+# import sphinx_rtd_theme
```
RamnathKumar181, line 16: Commented on this line due to the failure of flake8 on this file:
image

mlodic: hey, hi again. You did right to check flask before submitting the PR. However, in this case, this import is required to create the documentation properly. Travis build would still work with that because the `docs` folder is ignored in `.dockerignore`. To avoid to get this error again, I suggest you to modify `.flake8.yml` and add `docs` in the exclusion pattern because we do not want to lint those files too. Then, you can add again that import. Thank you!

mlodic:
Could you also update the documentation as stated in the instructions?

api_app/script_analyzers/observable_analyzers/phishtank.py
```diff
@@ -0,0 +1,39 @@
...
+        }
+        session = requests.Session()
+        api_key = secrets.get_secret(self.api_key_name)
+        if not api_key:
+            logger.warning(f"{self.__repr__()} -> Continuing w/o API key..")
+            print("Was here")
+        else:
+            data["app_key"] = api_key
+            # Change to phishtank/username once you add the API key
+            headers = {"User-Agent": "phishtank/ramnath"}
```
mlodic, line 31: you should change this user-agent to a more generic "IntelOwl"

api_app/script_analyzers/observable_analyzers/phishtank.py
```diff
@@ -0,0 +1,39 @@
...
+            "api_key_name", "PHISTANK_API_KEY"
+        )
+
+    def run(self):
+        result = {}
+        data = {
+            "url": base64.b64encode(self.observable_name.encode("utf-8")),
+            "format": "json",
+        }
+        session = requests.Session()
```
mlodic, line 23: I do not think that "Session" is required this time because you have to do only a single request.

api_app/script_analyzers/observable_analyzers/phishtank.py
```diff
@@ -0,0 +1,39 @@
...
+        api_key = secrets.get_secret(self.api_key_name)
+        if not api_key:
+            logger.warning(f"{self.__repr__()} -> Continuing w/o API key..")
+            print("Was here")
```
mlodic, line 27: remove debugging prints please

configuration/analyzer_config.json
```diff
@@ -420,6 +420,16 @@
     "python_module": "peframe_run",
     "description": "Perform static analysis on Portable Executable malware and malicious MS Office documents"
   },
+  "Phistank_Check_URL": {
```
mlodic, line 423: No need call it so long, just "Phishtank" is enough

api_app/script_analyzers/observable_analyzers/phishtank.py
RamnathKumar181: For some reason, requests return an empty dictionary, if not used with an API. "Session" is a workaround.

env_file_app_travis
```diff
@@ -40,6 +40,7 @@ INTELX_API_KEY=test
 WHOISXMLAPI_KEY=test
 UNPAC_ME_API_KEY=test
 URLSCAN_API_KEY=test
+PHISTANK_API_KEY=test
```
mlodic, line 43: typo

api_app/script_analyzers/observable_analyzers/phishtank.py
```diff
@@ -0,0 +1,37 @@
...
+from intel_owl import secrets
+import base64
+
+logger = logging.getLogger(__name__)
+
+
+class Phishtank(ObservableAnalyzer):
+    def set_config(self, additional_config_params):
+        self.api_key_name = additional_config_params.get(
+            "api_key_name", "PHISTANK_API_KEY"
```
mlodic, line 14: typo

api_app/script_analyzers/observable_analyzers/phishtank.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+
+    def run(self):
+        result = {}
+        data = {
+            "url": base64.b64encode(self.observable_name.encode("utf-8")),
+            "format": "json",
+        }
+        session = requests.session()
```
mlodic, line 23: still not needed

Eshaan7:
To resolve the merge conflicts, remove the function you defined in `tasks.py` and add the soft_time_limit as an attribute to the analyzer config you made for phishtank in `analyzer_config.json`

Eshaan7:
@RamnathKumar181. We will be releasing a new version of IntelOwl in a day or two so it will be appreciated if you can make the required changes ASAP. Thankyou!

api_app/script_analyzers/observable_analyzers/phishtank.py
RamnathKumar181: Hey, yes. I tried the above. The issue also exists when we use no API key. In this scenario, header would be an empty dictionary.

tests/test_observables.py
```diff
@@ -474,6 +475,16 @@ def test_circl_pdns(self, mock_get=None, mock_post=None, sessions_get=None):
         ).start()
         self.assertEqual(report.get("success", False), True)
 
+    def test_phishtank(self, *args):
+        report = phishtank.Phishtank(
+            "Phistank",
```
mlodic, line 480: typo

api_app/script_analyzers/observable_analyzers/phishtank.py
mlodic: Could you please give us proof about your statement?

configuration/analyzer_config.json
```diff
@@ -420,6 +420,17 @@
     "python_module": "peframe_run",
     "description": "Perform static analysis on Portable Executable malware and malicious MS Office documents"
   },
+  "Phistank": {
+    "type": "observable",
+    "observable_supported": ["url"],
+    "external_service": true,
+    "soft_time_limit": 100,
+    "description": "Check if url is verified in Phishtank API",
+    "python_module": "ObservableAnalyzer.Phishtank",
```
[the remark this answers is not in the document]
RamnathKumar181: Yes, I have tried this. If you want, I can change it to phishtank.Phishtank

Eshaan7:
The reason why it returns an empty response without using `requests.Session` is because the `requests.post` is mocked inside the test suite i.e. no HTTP request is made but rather mocked to give an empty `{}` response. This is done as to not bash the service's API everytime the tests run among other reasons.

@RamnathKumar181 You should check the analyzer by running it via the web interface. The reason you didn't test it via the web interface is also why https://github.com/intelowlproject/IntelOwl/pull/219#discussion_r505617810 this mistake was not spotted.  

So I advise you to kindly drop the `requests.Session` and just use `requests.post` and try to run the same via the web interface.

configuration/analyzer_config.json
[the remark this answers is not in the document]
Eshaan7: #229 - I have opened a new issue regarding this. We should find a way to detect such mistakes in the API testing.

api_app/script_analyzers/observable_analyzers/phishtank.py
Eshaan7: Please see my comment below which explains this behavior.

RamnathKumar181:
Hey @Eshaan7, I have made the changes and switched it back to `requests.post` instead of `session.post`.  Hope this works!
Thanks a lot for your time!

api_app/script_analyzers/observable_analyzers/phishtank.py
```diff
@@ -0,0 +1,38 @@
...
+        data = {
+            "url": base64.b64encode(self.observable_name.encode("utf-8")),
+            "format": "json",
+        }
+        api_key = secrets.get_secret(self.api_key_name)
+        if not api_key:
+            logger.warning(f"{self.__repr__()} -> Continuing w/o API key..")
+        else:
+            data["app_key"] = api_key
+            headers = {"User-Agent": "phishtank/IntelOwl"}
```
Eshaan7, line 29: shoudln't we send this header with or w/o api key, in both cases ? any particular reason why is this been sent only when there's an API key ?

Eshaan7: everything else, looks good. Can you please just make this little change and then I'd merge.

RamnathKumar181: That makes sense. Only issue is, the particular username IntelOwl needs to be created if we want to pass the 'User-Agent' field. I will make the change and push it immediately, but there might be errors since the username does not exist.

PR #232, opened 2020-10-19
Using Django-Rest-Durin for auth

Eshaan7:
- Updated IntelOwl-ng docker tag to `v1.6.2`
- Using the [Django-Rest-Durin](https://github.com/Eshaan7/django-rest-durin) library to handle tokens and auth.
- Removed `--time-limit=1000` flag from celery commands inside docker-compose files and instead added a default setting of 30 minutes in the configuration present inside `api_app/celery.py`


api_app/migrations/0003_auto_20201020_1406.py
```diff
@@ -0,0 +1,27 @@
...
+from django.db import migrations
+from datetime import timedelta
+
+
+def create_default_clients(apps, schema_editor):
+    # We can't import the Person model directly as it may be a newer
+    # version than this migration expects. We use the historical version.
+    Client = apps.get_model("durin", "Client")
+    # for pyintelowl, custom token_ttl
+    Client.objects.update_or_create(name="pyintelowl", token_ttl=timedelta(days=365))
```
mlodic, line 12: is it possible to customize this time to live? from the environment variables? I would set the default to "no expiration" if possible. In this way the pyintelowl token for a user would mimic a classic Token

Eshaan7: We can make it customizable, yes through an environment variable. I'll do that. 
However, I am struggling to have an option for "no expiration" because that introduces too many if conditions (hence bugs) which are quite hard to test. We could change the value to  2, 5 or even 10 years, which essentially is "no expiration" I guess.

mlodic: Yes you are right, a 100 years value I think is enough :P Where will IntelOwl be in 2120? :thinking:

Eshaan7: In 100 years ? I just wish I am not still working on authentication. xD

mlodic: no well, jokes aside, I think that we could set >=10 years like you mentioned for this so it is like "no expiration"

api_app/admin.py
```diff
@@ -28,36 +28,46 @@ class TagAdminView(GuardedModelAdmin):
...
-        return super(CustomTokenAdmin, self).add_view(request)
+        return super(CustomAuthTokenAdmin, self).add_view(request)
+
+    def has_change_permission(self, *args, **kwargs):
+        return False
+
+    def save_model(self, request, obj, form, change):
+        client = Client.objects.get(name="pyintelowl")
+        # lgtm [py/unused-local-variable]
+        obj = AuthToken.objects.create(user=obj.user, client=client)
```
mlodic, line 63: LGTM points that this variable is unused.

mlodic:
Great job for this. Let's merge into develop. Could you please squash the "f" commit before?

Eshaan7:
> Great job for this. Let's merge into develop. Could you please squash the "f" commit before?

Not able to do it for some reason :/. Maybe we can just squash and merge ?

api_app/admin.py
Eshaan7: it's used by django admin internally, that's why I put the `# lgtm...` ignore flag

PR #235, opened 2020-10-22
added googlesb special fields

drosetti:
Restored Googlesb extra data removed during DNS analyzer refactoring


mlodic:
Thank you for fixing it

PR #243, opened 2020-11-04
Added dnstwist analyzer

m0mosenpai:
2 main issues so far:

1. dnstwist is not able to find dictionaries as such (directly running `dnstwist` in shell. I had to clone their source repo (which has a dictionaries folder) and then run it from inside the folder to make it work. It needs absolute path to dictionary. I'm not sure where the dictionaries folder is, when the program is downloaded via PyPi. We might need to store the dictionaries separately.

2. Even though I implemented the `additional_config_params`, the approach could be better. Would like to know how I can make it better.

### DIFF
--- a/api_app/script_analyzers/observable_analyzers/dnstwist.py
+++ b/api_app/script_analyzers/observable_analyzers/dnstwist.py
@@ -1,0 +1,61 @@
+import subprocess
+import json
+import logging
+from shutil import which
+
+from api_app.exceptions import AnalyzerRunException
+from api_app.script_analyzers import classes
+
+logger = logging.getLogger(__name__)
+
+
+class DNStwist(classes.ObservableAnalyzer):
+    command: str = "dnstwist"
+    dictionary_base_path: str = "/opt/deploy/dnstwist-dictionaries/"
+
+    def set_config(self, additional_config_params):
+        self._ssdeep = additional_config_params.get("ssdeep", False)
+        self._mxcheck = additional_config_params.get("mxcheck", False)
+        self._tld = additional_config_params.get("tld", False)
+        self._tld_dict = additional_config_params.get("tld_dict", "abused_tlds.dict")
+
+    def run(self):
+        if not which(self.command):
+            self.report["success"] = False
+            raise AnalyzerRunException("dnstwist not installed!")
+
+        args = [self.command, "--registered", "--format", "json"]
+        final_report = {}
+
+        if self._ssdeep:
+            args.append("--ssdeep")
+            final_report["ssdeep"] = True
+        if self._mxcheck:
+            args.append("--mxcheck")
+            final_report["mxcheck"] = True
+        if self._tld:
+            args.append("--tld")
+            final_report["tld"] = True
+            args.append(self.dictionary_base_path + self._tld_dict)
+
+        args.append(self.observable_name)
+
+        process = subprocess.Popen(
+            args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        stdout, stderr = process.communicate()
+        logger.warning(stderr.decode("utf-8"))
+        final_report["error"] = stderr.decode("utf-8")
+
+        dns_report = stdout.decode("utf-8"), stderr
+        dns_str = dns_report[0]
+
+        try:
+            data = json.loads(dns_str)
+            final_report["data"] = data
+        except ValueError as e:
+            raise AnalyzerRunException(e)
+
+        return final_report
--- a/tests/test_observables.py
+++ b/tests/test_observables.py
@@ -32,6 +32,7 @@ from api_app.script_analyzers.observable_analyzers import (
     checkdmarc,
     urlscan,
     phishtank,
+    dnstwist,
 )
 from .mock_utils import (
     MockResponse,
@@ -387,6 +388,17 @@ class DomainAnalyzersTests(
             self.observable_name,
             self.observable_classification,
             {},
+        ).start()
+
+        self.assertEqual(report.get("success", False), True)
+
+    def test_dnstwist(self, mock_get=None, mock_post=None):
+        report = dnstwist.DNStwist(
+            "DNStwist",
+            self.job_id,
+            self.observable_name,
+            self.observable_classification,
+            {"tld": True, "mxcheck": True, "ssdeep": True},
         ).start()
 
         self.assertEqual(report.get("success", False), True)
--- a/api_app/script_analyzers/yara_repo_downloader.sh
+++ b/api_app/script_analyzers/yara_repo_downloader.sh
@@ -49,3 +49,6 @@ svn export https://github.com/quark-engine/quark-engine/tags/v20.08/quark/rules 
 
 # chown directories
 chown -R www-data:www-data /opt/deploy/yara /opt/deploy/quark-rules
+
+# Clone dictionaries for dnstwist analyzer
+svn export https://github.com/elceef/dnstwist/tags/20201022/dictionaries dnstwist-dictionaries
--- a/configuration/analyzer_config.json
+++ b/configuration/analyzer_config.json
@@ -163,6 +163,18 @@
       "api_version": 2
     }
   },
+  "DNStwist": {
+    "type": "observable",
+    "observable_supported": [
+      "domain",
+      "url"
+    ],
+    "external_service": true,
+    "description": "scans for potentially malicious permutations of a domain name",
+    "python_module": "dnstwist.DNStwist",
+    "additional_config_params": {
+    }
+  },
   "Doc_Info": {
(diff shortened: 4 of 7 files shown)


api_app/script_analyzers/observable_analyzers/dnstwist.py
```diff
@@ -0,0 +1,49 @@
...
+            args.append(self._tld_dict)
+
+        args.append(self.observable_name)
+
+        process = subprocess.Popen(
+            args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        process.wait()
```
mlodic, line 42: I think that this line is not required since `communicate` still waits for the end of the process.

tests/test_observables.py
```diff
@@ -461,6 +473,17 @@ def test_robtex_fdns(self, mock_get=None, mock_post=None):
         ).start()
         self.assertEqual(report.get("success", False), True)
 
+    def test_dnstwist(self, mock_get=None, mock_post=None):
```
mlodic, line 476: to avoid duplication of code could you please move this code only on `utils.py` under the `CommonTestCases_url_domain` class? In this way you just need to write the test once

mlodic (changes requested):
Thank you very much for your contribution! It seems really clear and neat. Could you also add the proper documentation? In the `Contribute` docs there are details on where you can add the required information

mlodic:
About your points:
1. Ok so you mean that in this draft the `abused_tlds.dict` cannot be found?
An idea could be to put the file with abused TLDs in the `configuration` directory of the project and also add that file in the bind mounted volumes in docker (see how `analyzer_config.json` is managed inside the docker-compose-*.yml files). In this way IntelOwl users will also be able to configure the TLDs as they prefer.

2. Honestly I do not know how we could make this better ATM. Surely there could be improvement in the management of the optional analyzer configuration but it is more a "project" problem than an problem of this analyzer/PR.

api_app/script_analyzers/observable_analyzers/dnstwist.py
```diff
@@ -0,0 +1,49 @@
...
+        stdout, stderr = process.communicate()
+
+        dns_report = stdout.decode("utf-8"), stderr
+        dns_str = dns_report[0]
+        dns_json = json.loads(dns_str)
```
mlodic, line 47: You could check the value in `stderr` and trigger an `AnalyzerRunException` in the case is populated.

Also, maybe you could verify that the output is supported JSON. Otherwise it would crash unexpectedly

m0mosenpai:
> About your points:
> 
> 1. Ok so you mean that in this draft the `abused_tlds.dict` cannot be found?
>    An idea could be to put the file with abused TLDs in the `configuration` directory of the project and also add that file in the bind mounted volumes in docker (see how `analyzer_config.json` is managed inside the docker-compose-*.yml files). In this way IntelOwl users will also be able to configure the TLDs as they prefer.

Shall I add the entire `dictionaries` folder in `configuration`? For `abused_tlds.dict` and other dictionaries too.

Eshaan7:
> An idea could be to put the file with abused TLDs in the configuration directory

Maybe we can add command which clones that folder via `svn` inside the `yara_repo_downloaded.sh`, the same way it's done for quark-engine rules ?

m0mosenpai:
> EDIT:
> 
> > It needs absolute path to dictionary.
> 
> It should be: `~/.local/lib/python3.7/site-packages/dnstwist/dictionaries`
> 
> Since we have the absolute path, we wouldn't need to download the dictionaries folder seperately.

I actually tried that and found no directory named `dnstwist` in `~/.local/lib/python3.6/site-packages` (for me). There is just a `dnstwist.py` file there.

Eshaan7:
I see. So add this command:

```bash
svn export https://github.com/elceef/dnstwist/tags/<insert_tag_here>/dictionaries` dnstwist-dictionaries
``` 

inside [`yara_repo_downloader.sh`](https://github.com/intelowlproject/IntelOwl/blob/develop/api_app/script_analyzers/yara_repo_downloader.sh).

Then in the code you can specify the path like `intelowl/configuration/dnstwist-dictionaries`

m0mosenpai:
Please review the changes made so far. I added the docs as well. Let me know if there is anything I need to change.

As for build issues, I have run tests for `dnstwist` separately and along with all other tests in one go. Both times the tests were successful but `yara_scan` is failing for some reason, both locally and in the latest build attempt on github.

@mlodic in regards to one of your suggestions to check `stderr` and raising an exception,
`--ssdeep` and `--geoip` give a "missing module error"  so the final results account for only `--mxcheck` and `--tld` working. For that reason, I have not checked `stderr` separately (if it's populated or not) and raised `AnalyzerRunException` since these error messages were causing the tests to fail, even though rest of the arguments were working fine and returning a valid json.
I'm not sure what/if there is a way around this, or some way I can get those arguments to work.

One last thing, testing takes a lot of time for `dnstwist` due to the program itself taking a lot of time to generate the relevant permutations of the url/domain. It probably slows down the entire testing quite a bit. Thought I should mention this.

mlodic:
@m0mosenpai thank you for your work and your detailed comment. It really helps a lot. We are almost done.

About the `yara` problem, this is related to a new addition to an external Yara repo that IntelOwl imports automatically: https://github.com/Neo23x0/signature-base/commits/master/yara/gen_icon_anomalies.yar. He fixed the problem last days so the next run of travis should find no problems.

About the check on `stderr` I agree with you. If some module is broken but the others not, it is right that the analysis do not fail. However we could anyway log that information in the results. In this way people will know what and if something went wrong. I suggest to change the result dictionary (what the `run` method returns) in something like:
```
{
"data":  {<dnstwist_results>}
"error": "<stderr>",
"ssdeep": <if set>,
<the_other_available_options>
}
```
You could also add a simple `logger.warning(<error>)`. In this way we would have that info logged outside the results too.

About the `geoip` not working: `Missing GeoIP module or database (geographical location not available)`. I think this is not extremely important. We could just drop the support for this option ATM.

Also `ssdeep` is not working because `dnstwist` package was not built to install dependencies.... :man_facepalming: so we should add `ppdeep>=20200505` to our requirements.txt and it should work.

To reduce the time of the Travis tests, I would kindly ask you to execute the test function only for `domain` type (please put it under `DomainAnalyzersTests`). In this way we would not not test for `urls`. However the behavior is really the same so I think that the extra time is not worth it.

Ultimately I would ask you to squash your commits into a single one. In the docs you can find a step-by-step section to do it correctly.

api_app/script_analyzers/observable_analyzers/dnstwist.py
```diff
@@ -0,0 +1,58 @@
...
+from shutil import which
+
+from api_app.exceptions import AnalyzerRunException
+from api_app.script_analyzers import classes
+
+
+class DNStwist(classes.ObservableAnalyzer):
+    command: str = "dnstwist"
+    dictionary_base_path: str = "/opt/deploy/dnstwist-dictionaries/"
+    final_report = {}
```
mlodic, line 13: care, this is a variable of the class. I would put this directly inside the `run` function.

api_app/script_analyzers/observable_analyzers/dnstwist.py
```diff
@@ -0,0 +1,58 @@
...
+
+        args.append(self.observable_name)
+
+        process = subprocess.Popen(
+            args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        stdout, stderr = process.communicate()
+        logging.warning(stderr.decode("utf-8"))
```
mlodic, line 46: Does this work? It would be better if you create an instance of `Logger` outside the function: `logger = logging.getLogger(__name__)` and then call `logger`. You can use as an example the `cymru.py` module.

mlodic (changes requested):
thank you!, just the last little changes and then I merge!

api_app/script_analyzers/observable_analyzers/dnstwist.py
m0mosenpai: Oh okay sure. May I ask why ? I don't think I know when to make a variable of the class and when not to.

mlodic: in this way, that variable is shared among all the instances of the DNStwist class. The report must be different between each execution, so it can't be here. (https://www.digitalocean.com/community/tutorials/understanding-class-and-instance-variables-in-python-3). Otherise the other strings you put are correctly placed there because they are the same in each execution.

mlodic (approved):

m0mosenpai:
Thank you for guiding me through this first open source contribution of mine! Looking forward to contribute more.

mlodic:
You are welcome, feel free to contribute more to the project!

PR #248, opened 2020-11-12
Increase file_name max_length

korrosivesec:
Increase the file_name max length to 64 to allow for samples where the file name is the SHA256 hash of the sample.


mlodic:
Thank you for that PR. Fun fact is that I had the same problem yesterday :sweat_smile: I increased the max_length value even further to avoid other similar problems

Eshaan7:
We would still have the issue when the analysis fails because the user unknowingly submitted a filename that was too large.
For cases where the submitted file's filename is more than 64 characters, maybe we can truncate it to first 60 characters + file extension on the API side only before saving it on the filesysten or in the database ? That would prevent the users from any need of debugging.

mlodic:
That would solve the problem definitely. However I am thinking about possible issues when looking for the data. I mean, people do expect that the filename is the same as the one they submitted, so they could be confused if they do not find the same value as before. This could be even an harder problem for the ElasticSearch integration that allows advanced searches. People would not find the filename

Eshaan7:
You make a valid point. For now, I'll just add small hint/text to the interface just below the file upload form stating "filename shouldn't exceed 64 characters".

mlodic:
I incremented the value to 512

PR #253, opened 2020-11-15
added triage analyzer

m0mosenpai:
Have added file analyzer part of triage so far. I had a few doubts though:

1. In [triage_scan.py line 63](https://github.com/m0mosenpai/IntelOwl/blob/triage_analyzer/api_app/script_analyzers/file_analyzers/triage_scan.py#L63), I have used `/samples/{id}/events` endpoint to make sure the submitted sample has finished being processed so the subsequent requests to generate reports don't return a `report not available` error in their reponse. I'm not sure if this is ok or is it better to send periodic requests to check if the report has been generated?

2. In `supported_filetypes` in `analyzer_config.json`, I have added most of the supported ones mentioned in their api docs but there were a few i couldn't find a clear mime type for. Have skipped those for now.

3. I only added a test function in `EXETests` section in `test_files.py` for now. Since there are many file types supported, I was wondering if there is a common way to test all of them together or should I add separate tests for each filetype?


configuration/analyzer_config.json
```diff
@@ -920,6 +920,39 @@
...
+      "application/internet-shortcut"
+
+    ],
+    "description": "leverage triage sandbox environment to scan various files",
+    "python_module": "triage_scan.TriageScanFile",
+    "requires_configuration": true,
+    "additional_config_params": {
+      "api_key_name": "TRIAGE_KEY",
+      "endpoint": "public"
+    }
```
mlodic, line 954: "soft_time_limit" parameter is missing here. Default is "300s". In your experience with the APIs, how much time is required usually to perform an analysis? Consider that this could vary. Please could you try with some of the provided files in the "test_files.zip" and paste here the effective time + the report data?

api_app/script_analyzers/file_analyzers/triage_scan.py
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        logger.info(f"triage md5 {self.md5} sending sample for analysis")
+        response = requests.post(
+            self.base_url + "samples", headers=headers, files=files
+        )
+        sample_id = response.json()["id"]
+
+        for _try in range(self.max_tries):
+            if response.status_code != 200:
```
mlodic, line 52: Could you please explain the purpose of this loop?

the first iteration would check the previous request status code. It seems an error. Is this wanted? Otherwise, you could manage this with `break` and the check at the end of the cycle.

api_app/script_analyzers/file_analyzers/triage_scan.py
```diff
@@ -0,0 +1,89 @@
...
+                logger.info(f"triage md5 {self.md5} polling for result try #{_try + 1}")
+                result_url = response.json().get("result_url", "")
+                response = requests.get(self.base_url + result_url)
+                response.raise_for_status()
+
+        if response.status_code != 200:
+            raise AnalyzerRunException("max retry attempts exceeded")
+
+        # Event stream is opened. Updates till the task is completed
+        requests.get(self.base_url + f"samples/{sample_id}/events", headers=headers)
```
mlodic, line 63: this is ok I think

api_app/script_analyzers/file_analyzers/triage_scan.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # Get static report of sample
+        static_report = requests.get(
+            self.base_url + f"samples/{sample_id}/reports/static",
+            headers=headers,
+        )
+        static_report_json = static_report.json()
+        final_report["static_report"] = static_report_json
+
+        return final_report
```
mlodic, line 89: could you please provide screenshot or paste of some example reports?

mlodic (changes requested):
thank you for the PR.

> I only added a test function in EXETests section in test_files.py for now. Since there are many file types supported, I was wondering if there is a common way to test all of them together or should I add separate tests for each filetype?

It is enough a single file type as you did. The point is to test the analyzer, not the files.

EDIT: I also think that you may need to mock the Triage responses accordingly to let the test work. Please see how Cuckoo or VT tests were mocked.

api_app/script_analyzers/file_analyzers/triage_scan.py
m0mosenpai: I ran the test on a random sample i took from the publicly available ones. Summary and the other reports might have some redundant stuff. I found these endpoints in their docs and just compiled all the results. Will change/ remove things not needed.
[Tree format from the GUI](https://imgur.com/a/BvdtVgX)
[Raw](https://imgur.com/a/ln0XstJ)

configuration/analyzer_config.json
m0mosenpai: image

I put the test function in both EXEs and DLLs in this case as well as testing some random files on the GUI. Time to finish varied between ~160s and ~185s

api_app/script_analyzers/file_analyzers/triage_scan.py
m0mosenpai: This was intended for retrying sending requests incase response wasn't 200. I took reference from a few other analyzers for this. I can remove it if it's not necessary.

mlodic: Ok, good idea but the request below is different from the previous one. Is this intended too?

api_app/script_analyzers/file_analyzers/triage_scan.py
mlodic: Yes, I was thinking the same. My concern is to avoid duplicated data in the result to avoid confusion.

At a first glance it seems that the "summary" data are replicated in the "overview". So we can remove it I think.

What about the data in the "static" report? are they replicated in the "overview" too?

configuration/analyzer_config.json
```diff
@@ -920,6 +920,39 @@
     "python_module": "tor.Tor",
     "soft_time_limit": 30
   },
+  "Triage_Scan": {
+    "type": "file",
+    "supported_filetypes":[
```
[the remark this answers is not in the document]
mlodic: They support almost anything so I would leave this empty

mlodic (changes requested):

configuration/analyzer_config.json
mlodic: ok so we could put the soft_time_limit explicitly to 300 anyway. Otherwise the risk is that if we change the default value we change this too, creating a possible unwanted behavior

api_app/script_analyzers/file_analyzers/triage_scan.py
m0mosenpai: I removed summary and added task specific reports. I noticed on the wesbite, that the analyzed report contains 4 or more parts : `overview`, `static_report`, `<task_1>`, `<task_2>` and so on as seen below. So i added task wise report alongside overview and static in the final json. All of them share some basic common info but have their own unique analysis so I have kept all of them and removed only summary.

image

api_app/script_analyzers/file_analyzers/triage_scan.py
m0mosenpai: Oh no sorry that was a mistake. I made some changes to the way this is handled. Please have a look.

m0mosenpai:
> EDIT: I also think that you may need to mock the Triage responses accordingly to let the test work. Please see how Cuckoo or VT tests were mocked.

I have been testing locally so far using the key. Will look into mocking and commit.

api_app/script_analyzers/file_analyzers/triage_scan.py
```diff
@@ -0,0 +1,97 @@
...
+        logger.info(f"triage md5 {self.md5} sending sample for analysis")
+        response = requests.post(
+            self.base_url + "samples", headers=headers, files=files
+        )
+
+        _try = 0
+        while _try < self.max_tries and response.status_code != 200:
+            time.sleep(self.poll_distance)
+            logger.info(f"triage md5 {self.md5} polling for result try #{_try + 1}")
+            response = requests.post(
```
mlodic, line 54: ok but this is still duplicated. you could rewrite this better with a single requests.post

api_app/script_analyzers/file_analyzers/triage_scan.py
mlodic: ok, thank you for explaining this! 

Anyway I would change this to avoid too much data in the report. The best option could be to make this configurable. For instance, by default this analyzer could provide only the "overview" report. And, with a specific parameter, the analysis report could be "complete" with the static and the tasks reports too.

m0mosenpai: Added an option in `additional_config_params`  for this.

api_app/script_analyzers/file_analyzers/triage_scan.py
m0mosenpai: I made some changes to that and also restructured the code a bit.

mlodic (approved):

mlodic:
Merged! :rocket:

m0mosenpai:
> Merged! 🚀

Oh thank you! I was yet to add the docs and squash the commits though . And this is only the file analyzer part. Shall I make another PR for the observable part ? I'll add the docs for both in one go.

mlodic:
Yeah, that was my fault, I did not notice that. Now it's late, I won't mess with the commit history. My bad. :/

And yes please if you can create another PR it would be good

PR #254, opened 2020-11-16
search on bazaar.abuse.ch via google

theguly:
it's possible to use google search to achieve a similar result


mlodic:
I did not exactly understand what you tried to do here. Is that a question? In that case please open an issue, not a PR.

theguly:
sorry, wrong tree. was a PR against dev that added a feature to search malwarebazaar through google

PR #255, opened 2020-11-16
query abuse.ch via google, to search also IP domain and urls

theguly:
using mb_google analyzer added with this PR it's now possible to use google search to achieve a similar result

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,8 @@ WORKDIR $PYTHONPATH
 RUN pip3 install --no-cache-dir --compile -r requirements.txt
 # install elasticsearch-dsl's appropriate version as specified by user
 RUN pip3 install --no-cache-dir django-elasticsearch-dsl==${ELASTICSEARCH_DSL_VERSION}
+# install google for mb_google
+RUN pip3 install --no-cache-dir google
 
 COPY . $PYTHONPATH
 
--- a/configuration/analyzer_config.json
+++ b/configuration/analyzer_config.json
@@ -355,6 +355,18 @@
     "python_module": "ha_get.HybridAnalysisGet",
     "soft_time_limit": 30
   },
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -26,6 +26,8 @@ WORKDIR $PYTHONPATH
 RUN pip3 install --no-cache-dir --compile -r requirements.txt
 # install elasticsearch-dsl's appropriate version as specified by user
 RUN pip3 install --no-cache-dir django-elasticsearch-dsl==${ELASTICSEARCH_DSL_VERSION}
+# install google for mb_google
+RUN pip3 install --no-cache-dir google
```
mlodic, line 30: this should be moved in `requirements.txt`

configuration/analyzer_config.json
```diff
@@ -355,6 +355,18 @@
     "python_module": "ha_get.HybridAnalysisGet",
     "soft_time_limit": 30
   },
+  "IPInfo": {
```
mlodic, line 358: I think something went wrong with a local merge. I suggest you to start again from a clean `develop` branch and add your analyzer on top of it

tests/test_observables.py
```diff
@@ -534,6 +535,16 @@ def test_mb_get(self, mock_get=None, mock_post=None):
         ).start()
         self.assertEqual(report.get("success", False), True)
 
+    def test_mb_google(self, mock_get=None, mock_post=None):
```
mlodic, line 538: this should be put in `utils.py` under the `CommonTestCases_ip_url_domain` class

mlodic (changes requested):
Thank you for your interest in the project! Please follow the suggestions I made to create a clean PR. Please also add the related documentation (follow the doc). It is a matter of few lines.

PR #256, opened 2020-11-16
query abuse.ch via google, to search also IP, domain, and url

theguly:

using mb_google analyzer added with this PR it's now possible to use google search to achieve a similar result


mlodic:
Please follow [this](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-squash-commits-to-a-single-one) to squash commits into a single one. This is to maintain a clean history. Then I think we are done.

mlodic:
I can't add this id you have not run `black` and `flake` as stated in the docs

PR #258, opened 2020-11-17
Added a new validation check. 

0ssigeno:
The object should not have more than 200 nested levels. Otherwise we can have problems to store it in dbs.

### DIFF
--- a/api_app/script_analyzers/classes.py
+++ b/api_app/script_analyzers/classes.py
@@ -57,16 +57,30 @@ class BaseAnalyzerMixin(metaclass=ABCMeta):
         In most cases, this would be overwritten.
         """
 
-    def _validate_result(self, result):
+    def _validate_result(self, result, level=0, max_recursion=190):
         """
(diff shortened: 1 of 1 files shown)


api_app/script_analyzers/classes.py
```diff
@@ -57,16 +57,21 @@ def set_config(self, additional_config_params):
...
+    def _validate_result(self, result, level=0, max_recursion=190):
         """
-        function to validate result, allowing to store inside postgres without errors
+        function to validate result, allowing to store inside postgres without errors.
+        If the character \u0000 is present in the string, postgres will throw an error
+        If an integer is bigger than max_int, Mongodb is not capable to store and will throw an error.
+        If we have more than 200 recursion levels, every encoding will throw a maximum_nested_object exception
         """
+        if level == max_recursion:
+            return None
```
mlodic, line 68: could you add a log stating that this happened?

mlodic (changes requested):
squash please?

mlodic (approved):

PR #259, opened 2020-11-17
Upgraded yara to 4.0.2

0ssigeno:
Checked if the internal error 30 is raised. [The error 30 means too many matches](https://github.com/VirusTotal/yara/blob/2289bbabb539045d18200a5d7c49fca6e4866d06/libyara/include/yara/error.h#L75), but it is not possible to attach a callback to check which match triggers. Or at least I wasn't able to.


mlodic:
could you please fix codefactor?

mlodic:
Nice, thank you!

Issue #217, opened 2020-10-12
Add https://github.com/sbousseaden/YaraHunts rules


mlodic:
We could add the following rules (https://github.com/sbousseaden/YaraHunts) to the default Yara rules available

Issue #214, opened 2020-10-09
Add reversinglabs yara rules


mlodic:
We could add the following rules (https://github.com/reversinglabs/reversinglabs-yara-rules) to the default Yara rules available.

PR #260, opened 2020-11-18
More yara repositories

0ssigeno:
Should solve issue #214, #217


mlodic (changes requested):
Could you also add them in the `Usage` section? Thank you!

PR #271, opened 2020-11-26
Merged newly added FLOSS analyzer with Capa and PEFrame

CITIZENDOT:
Currently, It is based on Linux binary. I placed it in project folder. I'm not sure where to place it instead. And additional options for `floss` needed to give in `runtime_config`. 

And I didn't added tests to it, Since I didn't understand how to add tests for binary type of analyzer.
Suggestions?

**Edit**:
Floss

**Edit**:

#### Previous Routes:
---------
- Capa: http://capa:4002/capa
- PEFrame: http://peframe:4000/peframe

#### Now:
-------
- Capa: [http://static_analyzers:4002/capa](http://static_analyzers:4002/capa)
- PEFrame: [http://static_analyzers:4002/capa](http://static_analyzers:4002/capa)
- FLOSS: [http://static_analyzers:4002/floss](http://static_analyzers:4002/floss)

Logs of above analyzers are stored in `/var/log/intel_owl/static_analyzers` directory.
Open Ports for FLASK (now): `4001`, `4002`, `4003`, `4004`

---


configuration/analyzer_config.json
```diff
@@ -229,6 +229,15 @@
     "python_module": "file_info.FileInfo",
     "soft_time_limit": 30
   },
+  "Floss": {
+    "type": "file",
+    "description": "Static Analysis of Binaries",
```
mlodic, line 234: I would say something more useful for the user. Ex: "advanced strings extraction from FireEye"

mlodic (changes requested):
A lot of things are missing from this PR. Have you checked the docs?

Also we cannot embed binaries like you did. The binary should be downloaded from the source at the time of the build. 
I would aggregate this analyzer in one of the already existing docker-based analyzers. In particular, the `capa` one.
This is because both `capa` and `floss` are static file analyzers from Fireeye and both needs to be downloaded from the source.

So, I would strongly suggest to check how the `capa` analyzer is actually implemented, follow the docs about the `docker-based` analyzers and try to add this analyzer in this way. Otherwise it would be not clean to embed this binary directly inside the main docker image.

configuration/analyzer_config.json
CITIZENDOT: Sorry, I wasn't aware of this type of analyzers, So, I directly kept a file And Sure. I'll make the changes.

configuration/analyzer_config.json
```diff
@@ -229,6 +229,15 @@
...
+    "python_module": "floss.Floss",
+    "soft_time_limit": 30,
```
Eshaan7, line 236: are you sure this time limit is enough ? did you try to run the binary locally against any file and get an approximation of average time the process takes ? a value like `120` might be more suitable

CITIZENDOT: No, I am changing it to 500, Since `capa` also have the same.

CITIZENDOT:
@mlodic, There isn't json output in floss. But It can log the output into a Json file. If I log output into JSON, What should be the temporary JSON file? Am I going in the correct approach?

mlodic:
> @mlodic, There isn't json output in floss. But It can log the output into a Json file. If I log output into JSON, What should be the temporary JSON file? Am I going in the correct approach?

The same problem was solved for `box-js` by @Eshaan7. Please use the related docker-based analyzer as an example.

Eshaan7:
Python provides a library for dealing with temporary directories and temporary files. That's what you should use.

CITIZENDOT:
Thanks @mlodic , @Eshaan7 , Everything is working now.

.env
```diff
@@ -17,7 +17,7 @@ COMPOSE_FILE=docker-compose.yml
 ###### For Tests or local development ######
 #COMPOSE_FILE=docker-compose-for-tests.yml
 # To run all additional integrations in development
-#COMPOSE_FILE=docker-compose-for-tests.yml:./integrations/docker-compose-for-tests.peframe.yml:./integrations/docker-compose-for-tests.thug.yml:./integrations/docker-compose-for-tests.capa.yml:./integrations/docker-compose-for-tests.boxjs.yml
+COMPOSE_FILE=docker-compose-for-tests.yml:./integrations/docker-compose-for-tests.capa.yml:./integrations/docker-compose-for-tests.floss.yml
```
mlodic, line 20: you forgot to do not push the change on this file

mlodic (changes requested):
A new docker-based analyzer is unecessary.

| Also we cannot embed binaries like you did. The binary should be downloaded from the source at the time of the build.
I would aggregate this analyzer in one of the already existing docker-based analyzers. In particular, the capa one.
This is because both capa and floss are static file analyzers from Fireeye and both needs to be downloaded from the source.

.env
CITIZENDOT: Oh, I'll update.

mlodic:
We should also set a cap in the found strings like we did in `strings_info.py`

mlodic:
There is no need to push every time here. Please fork your personal repo and push in the `develop` repo ogni when you have finished

CITIZENDOT:
I'm forgetting somethings.😅. By the way, Should I consider Codefactor warning severe?

mlodic:
> Sorry. I'm forgetting somethings.sweat_smile. By the way, Should I consider Codefactor warning severe?

I can silence it. The important thing is to be sure that the temporary file will be deleted in `any` case. Some generic `try except` clause can help

Eshaan7:
Make sure to update the optional analyzers section in the docs.

mlodic:
Yesterday, with the new release of `pip` (20.3)  they started to force the new method of resolve conflicts that they announced months ago. Unfortunately for your PR this means that it won't work until we solve that problem.

First problem is:
```
The conflict is caused by:
    The user requested colorama==0.3.7
    peepdf 0.4.2 depends on colorama==0.3.7
    quark-engine 20.11 depends on colorama
    androguard 3.4.0a1 depends on colorama>=0.4.1
```
Peepdf is quite an old tool but still effective, while quark-engine is a super recent tool. Obviously something bad happens when both try to use the same package...

The other one I noticed by building the actual IntelOwl version is this one:
```
matplotlib 3.3.3 requires pillow>=6.2.0, but you'll have pillow 3.2.0 which is incompatible.
```
Even for this case Pillow is forced to v3.2.0 to allow to properly install `peepdf`.

So, I do not want to remove `peepdf` so my first solution is to work on peepdf to let it work with recent versions of those packages and publish a new version of it on pip.

Every other faster idea is welcome

CITIZENDOT:
> Make sure to update the optional analyzers section in the docs.
> (https://intelowl.readthedocs.io/en/develop/Advanced-Usage.html#optional-analyzers)

Shall I point PEFrame reference [here](https://intelowl.readthedocs.io/en/latest/Contribute.html#integrating-a-docker-based-analyzer) to Box-Js?

mlodic:
I just changed the pointer to `peepdf` to a custom version of mine to avoid dependency problems. I have restarted the Travis build too.

This problem would mean that also new installation of IntelOwl from the master version would fail because we did not pin the `pip` version. I am about to release a fix version

.env
```diff
@@ -12,15 +12,15 @@ INTELOWL_TAG_VERSION=v1.9.0
 ###### Default (Production) ######
 COMPOSE_FILE=docker-compose.yml
 # To run all additional integrations in production
-#COMPOSE_FILE=docker-compose.yml:./integrations/docker-compose.peframe.yml:./integrations/docker-compose.thug.yml:./integrations/docker-compose.capa.yml:./integrations/docker-compose.boxjs.yml
+#COMPOSE_FILE=docker-compose.yml:./integrations/docker-compose.thug.yml:./integrations/docker-compose-for-tests.exe_analyzers.yml:./integrations/docker-compose.boxjs.yml
```
Eshaan7, line 15: In this line, it should be `docker-compose.exe_analyzers.yml`, without `for-tests` part

Eshaan7: Meanwhile, I also noticed that we are missing the `docker-compose.apk.yml` entry. on line 20 and 23 for the test compose files as well. We should add this in another commit.

mlodic: The last thing was because I found unresolved problems with the logs on a clean startup. So I disabled that container for the moment. I did not open any specific issue, while I should have did it

CITIZENDOT: @mlodic, Could you explain more about your problem with logs? I too encountered one at the start. 

It's log directory wasn't creating in `/var/log/intel_owl`. I don't know why this happened, I created in manually by entering into the docker shell. Is this you faced?

But from the next clean build (Deleted local volumes of previous), it worked fine.

Still, I didn't understand why it behaved like that. 

Is this what you faced?

Eshaan7: Yes, this problem happens sometimes. It's something to do with permissions which gets changed on changing the log directories and files.

integrations/exe_analyzers/Dockerfile
```diff
@@ -0,0 +1,49 @@
+FROM python:3.8.4-slim
+
+ENV PROJECT_PATH /opt/deploy
+ENV LOG_PATH /var/log/intel_owl/exe_analyzers
+
+RUN mkdir -p $LOG_PATH
```
Eshaan7, line 6: Doesn't this directory gets created in the main IntelOwl dockerfile ? Is there a reason you added this command here ?

integrations/exe_analyzers/entrypoint.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
 
 touch ${LOG_PATH}/gunicorn_access.log ${LOG_PATH}/gunicorn_errors.log
-chown -R capa-user:capa-user ${LOG_PATH}
-su capa-user -s /bin/bash
+chown -R exe_analyzers-user:exe_analyzers-user ${LOG_PATH}
+su exe_analyzers-user -s /bin/bash
 exec gunicorn 'app:app' \
```
Eshaan7, line 6: Missing options like,

```
    --access-logfile ${LOG_PATH}/gunicorn_access.log \
    --error-logfile ${LOG_PATH}/gunicorn_errors.log
```

.env
CITIZENDOT: So, Shall I add both changes into one? It's a small change and not needed of one extra commit, IMO

Eshaan7: It's a small change but we should still have it in a different commit/PR since it's not related to changes introduced in this particular pull request and then it's also easier to track.

integrations/exe_analyzers/Dockerfile
CITIZENDOT: Oh, I should remove it actually,Thanks for pointing out. While testing several methods to resolve it, I tried this once. Yeah, In the first build it didn't get created. From next time, It's working fine. I'll remove this line now.

mlodic:
Could we also call the new container `static_analyzers`? I know this a detail but I think it is more appropriate

CITIZENDOT:
> Could we also call the new container `static_analyzers`? I know this a detail but I think it is more appropriate

I Agree. `static_exe_analyzers`? Too long?

mlodic:
The point is that they can not analyze only `.exe` files but the thing that they share is that they analyze files statically. 

However I just changed the string format in optional docker containers (see [commit](https://github.com/intelowlproject/IntelOwl/commit/4ba80753479920a90e912d8002911f6873064930), please resolve the conflicts locally and add milliseconds in the merged container too. 

Thank you in advance

mlodic:
Actually my last change was an error, I reverted it. You should not have merge conflicts anymore

CITIZENDOT:
Is it necessary to squash commits now? I thought, changes like, Peepdf and version changes should be visible and shouldn't go hidden inside this commit.

mlodic:
I will squash all commits anyway when merging to maintain a clear history

docs/source/Advanced-Usage.md
```diff
@@ -26,20 +26,21 @@ table, th, td {
     <th>Description</th>
   </tr>
   <tr>
-    <td>PEframe</td>
-    <td><code>PEframe_Scan</code></td>
-    <td>performs static analysis on Portable Executable malware and generic suspicious file</td>
+    <td>EXE Analyzers</td>
```
mlodic, line 29: this should be updated with the new name

mlodic (commented):
ok so should we consider this ready to be merged?

CITIZENDOT:
Yeah it's almost done , I'll update the name and squash it now.

mlodic (approved):

Eshaan7:
The codefactor alert was an old alert only so I put in the ignored list again.

.env
```diff
@@ -7,20 +7,20 @@
 
 ### DO NOT CHANGE THIS VALUE !!
 ### It should be updated only when you pull latest changes off from the 'master' branch of IntelOwl.
-INTELOWL_TAG_VERSION=v1.9.0
+INTELOWL_TAG_VERSION=v1.9.1
```
Eshaan7, line 10: why is this part of this PR ? the version number is `v1.9.1` in `develop` branch already.

.github/CHANGELOG.md
```diff
@@ -2,6 +2,24 @@
 
 [**Upgrade Guide**](https://intelowl.readthedocs.io/en/latest/Installation.html#update-to-the-most-recent-version)
 
+## [v1.9.1](https://github.com/intelowlproject/IntelOwl/releases/tag/v1.9.1)
```
Eshaan7, line 5: same goes for this.

configuration/intel_owl_nginx_http
```diff
@@ -9,6 +9,8 @@ server {
 
     server_name intel_owl.com;
 
+    server_tokens off;
```
Eshaan7, line 12: why this change ?

docs/source/conf.py
```diff
@@ -22,7 +22,7 @@
 author = "Matteo Lodi"
 
 # The full version, including alpha/beta/rc tags
-release = "1.9.0"
+release = "1.9.1"
```
Eshaan7, line 25: same, shouldn't be part of this PR. I think there is need to pull and merge once.

integrations/docker-compose.static_analyzers.yml
```diff
@@ -5,9 +5,9 @@ version: '3'
 # All additional integrations should be added following this format only.
 
 services:
-  capa:
-    image: intelowlproject/intelowl_capa:${INTELOWL_TAG_VERSION}
-    container_name: intelowl_capa
+  static_analyzers:
+    image: intelowlproject/intelowl_static_analyzers:${INTELOWL_TAG_VERSION}
```
Eshaan7, line 9: /reminder: need to create a repository on dockerhub for this new docker container and stop autobuild of the deprecated ones.

Eshaan7: @mlodic can you create a new repo on dockerhub with name `intelowl_static_analyzers`, then I'll configure the auto-builds.

Eshaan7: disabled auto-build on dockerhub for capa and peframe repos.

mlodic: done!

configuration/intel_owl_nginx_http
mlodic: this was a commit of mine in `develop` to remove the Nginx version banner

intel_owl/settings.py
```diff
@@ -219,7 +219,6 @@
     "formatters": {
         "stdfmt": {
             "format": "%(asctime)s - %(name)s - %(funcName)s - %(levelname)s - %(message)s",
-            "datefmt": "%Y-%m-%d %H:%M:%S",
```
Eshaan7, line 222: :/

.env
CITIZENDOT: I pulled the changes. Hence🤷‍♂️

mlodic: a `git` rebase would have worked better I think

integrations/docker-compose.static_analyzers.yml
Eshaan7: configured. :+1:

.env
CITIZENDOT: Actually, I did a mistake in the start. I should have created another branch and then keep rebasing from develop. 😒. Hurriedly, I worked on develop branch and so, pulls are reflecting directly here.


Sorry, I'll create branches now on.

intel_owl/settings.py
mlodic: this is a commit of mine too. This is to add milliseconds in logs

.env
CITIZENDOT: Should I make any changes now?

PR #274, opened 2020-11-30
Added 3 different celery queues for analyzers.

0ssigeno:
It is possible to select the queue inside analyzer_config.json.
Added eventlet for the I/O queues.

### DIFF
--- a/intel_owl/settings.py
+++ b/intel_owl/settings.py
@@ -145,7 +145,7 @@ CELERY_WORKER_HIJACK_ROOT_LOGGER = False
 # these two are needed to enable priority and correct tasks execution
 CELERY_TASK_ACKS_LATE = True
 CELERY_WORKER_PREFETCH_MULTIPLIER = 1
-CELERY_TASK_DEFAULT_QUEUE = "analyzers_queue"
+CELERY_QUEUES = os.environ.get("CELERY_QUEUES", "default").split(",")
 # this is to avoid RAM issues caused by long usage of this tool
 CELERY_WORKER_MAX_TASKS_PER_CHILD = 200
 # value is in kilobytes
--- a/.env
+++ b/.env
@@ -11,6 +11,8 @@ INTELOWL_TAG_VERSION=v1.9.0
 
 ###### Default (Production) ######
 COMPOSE_FILE=docker-compose.yml
+##### Multi-Queue (Production) #####
+COMPOSE_FILE=docker-compose-multi-queue.yml
 # To run all additional integrations in production
 #COMPOSE_FILE=docker-compose.yml:./integrations/docker-compose.peframe.yml:./integrations/docker-compose.thug.yml:./integrations/docker-compose.capa.yml:./integrations/docker-compose.boxjs.yml
 
--- a/api_app/script_analyzers/general.py
+++ b/api_app/script_analyzers/general.py
@@ -1,12 +1,12 @@
 import logging
 from celery.execute import send_task
 
-from intel_owl import settings
 from api_app.exceptions import (
     AnalyzerConfigurationException,
     AnalyzerRunException,
 )
 from api_app.helpers import generate_sha256
+from intel_owl.settings import CELERY_QUEUES
 from .utils import (
     set_job_status,
     set_failed_analyzer,
@@ -46,7 +46,16 @@ def start_analyzers(
             adjust_analyzer_config(
                 runtime_configuration, additional_config_params, analyzer
             )
+            # get celery queue
+            queue = ac.get("queue", "default")
+            if queue not in CELERY_QUEUES:
+                logger.error(
+                    f"Analyzer {analyzers_to_execute} has a wrong queue."
+                    f" Setting to default"
+                )
+                queue = "default"
             # construct arguments
+
             if is_sample:
                 # check if we should run the hash instead of the binary
                 run_hash = ac.get("run_hash", False)
@@ -98,7 +107,7 @@ def start_analyzers(
             send_task(
                 "run_analyzer",
                 args=args,
-                queue=settings.CELERY_TASK_DEFAULT_QUEUE,
+                queue=queue,
                 soft_time_limit=stl,
             )
 
--- a/configuration/analyzer_config.json
+++ b/configuration/analyzer_config.json
@@ -103,7 +103,8 @@
     "supported_filetypes": ["application/x-dosexec"],
     "python_module": "capa_info.CapaInfo",
     "description": "capa detects capabilities in executable files",
-    "soft_time_limit": 500
+    "soft_time_limit": 500,
+    "queue": "long"
   },
   "Censys_Search": {
     "type": "observable",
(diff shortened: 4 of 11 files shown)


Eshaan7:
Is the performance improvement worth the 2 extra containers and added complexity?

0ssigeno:
Eventlet was designed to work better with I/O bound threads. The vast majority of our analyzers is I/O bound, since we are using API in many cases. Unfortunely, it is not always the case. For example the docker analyzers or the static file analyzer.
Moreover, we have a huge difference in timeouts between analyzers. Meaning that we could, and we actually did, block the entire IntelOwl if many long task are scheduled. This way we can at least provide the results of the fast analyzers, even if the long queue is full. 

The number one cons of this implementation, is that yes, IntelOwl will require more resources to be run, since two more containers will be created. A solution for this problem can be to provide another `docker-compose` production file, allowing the users to choose the architecture that better describe their needs.

I thought about putting eventlet as the default queue. Again, our vast majority of analyzers is I/O bound, so probably is a good idea. The only problem are the tests. The default queue has been tested till this day in a production environment, and it does it job quite well. We have no data if using eventlet will cause secondary effect that a short time testing is not able to detect (i.e. memory leaks and whatever).

0ssigeno:
I will wait your thoughts on the matter before completing the PR for travis.

mlodic:
I think that `eventlet` is absolutely worth to be experimented. I also agree with @Eshaan7 concerns about actual IntelOwl users. Lot of them just use IntelOwl "the easy way" and do not need this change at all. On the other hand, groups that would like to make IntelOwl scale appropriately would probably benefit from a change like that. So ATM IMHO it is better to have this new feature as an optional configuration. We could explain the benefits + additional configuration required in the "Advanced Usage" section of the docs. Thoughts?

0ssigeno:
I'm fine with that. Will add the `eventlet` queue as default everywhere and will provide an optional `docker-compose` for a multi-queue IntelOwl.

intel_owl/settings.py
```diff
@@ -145,7 +146,7 @@
 # these two are needed to enable priority and correct tasks execution
 CELERY_TASK_ACKS_LATE = True
 CELERY_WORKER_PREFETCH_MULTIPLIER = 1
-CELERY_TASK_DEFAULT_QUEUE = "analyzers_queue"
+CELERY_QUEUES = os.environ.get("CELERY_QUEUES", "default").split(".")
```
Eshaan7, line 149: I think `CELERY_QUEUES` variable should be defined in the `env_file_app`, right ?

Eshaan7 (changes requested):

intel_owl/settings.py
0ssigeno: I get your point. The problem is that we should then provide two different `env_file_app`, one for the `multi-queue`, one for the standard. Since I personally don't like to provide to users too many different files, I decided to use the `environment` keyword inside the docker-compose-multi-queue.yaml to describe this behaviour.

0ssigeno: Well, thinking about it a little bit more, you are probably right. Splitting the environment variables is not a good practice.
Maybe I can add something like this inside the `env_file_app`, what do you think?
```
...
CELERY_QUEUES="default"
# USE THIS FOR THE MULTI-QUEUE
# CELERY_QUEUES="default.local.long"
...
```

Eshaan7: ok, my bad actually; I didn't notice that you put it inside the compose file itself, I thought it was missing. Then I think your intial implementation was correct because there's no point in leaving this upto the user to decide. 

Just one thing, instead of `default.local.long` can you change it to `default,local,long` and then split on the basis of `,` ?

Eshaan7: ok, my bad actually; I didn't notice that you put it inside the compose file itself, I thought it was missing. Then I think your intial implementation was correct because there's no point in leaving this upto the user to decide since it is tied with the compose file they choose.

Just one thing, instead of `default.local.long` can you change it to `default,local,long` and then split on the basis of `,` ?

0ssigeno: Yeah, sure

Eshaan7:
Another thought: with this implementation, the user would have 2 options: `docker-compose.yml` and `docker-compose-multi-queue.yml`, now what if they choose the `traefik` compose file ? then they can only use single queue but not the multi queue. We could create another compose file for `traefik` + `multi-queue` but this is not ideal because too many compose files multiplicating in parallel directions.

Maybe we can totally remove the celery worker services from all base docker compose files (production, test, travis, traefik) and create 2 standalone files like `celery.single-queue.yml` and `celery.multi-queue.yml` ? and then use the `.env` to join them like we do for optional analyzer services ?

mlodic:
Yes, at this point we should completely rework all the docker-compose files to leverage the override feature (https://docs.docker.com/compose/extends/).

This would reduce also all the problems related to the duplication of code in those files

Eshaan7:
The `extends` keyword was removed after compose `v2.1` but we are using v3 of docker-compose in all the compose files. ([source](https://docs.docker.com/compose/extends/#extending-services)).

What I am suggesting is something similar to [this comment](https://github.com/moby/moby/issues/31101#issuecomment-309071230).

In any case, you are right. We do need to rework all compose files and also the directory structure.

Eshaan7:
I also found this [comment](https://github.com/docker/compose/pull/7588#issuecomment-708390804) which says we can use `extends` keyword in v3 as well but the same is not updated in docker's docs. Will need to experiment with this and see for ourselves.

mlodic:
Sorry maybe I was misunderstood with that link. I did not mean to refer to `extend` but just to the override feature between docker-compose files.

Example, we can avoid to set this in each compose file:
```
  rabbitmq:
    image: library/rabbitmq:3.8-alpine
    container_name: intel_owl_rabbitmq
```
We can create a "base" docker-compose file that contains the basic configuration. 
What I meant is basically your idea but applied to all the services, not only the ones related to celery.

Eshaan7:
>  What I meant is basically your idea but applied to all the services, not only the ones related to celery.
Ah, I see. Yes, definitely what I meant as well. So, let's create a `compose.common.yml` file with the services like `rabbit-mq`, `postgres` which are common in all compose files.

Eshaan7:
> we can totally remove the celery worker services from all base docker compose files (production, test, travis, traefik) and create 2 standalone files like `celery.single-queue.yml` and `celery.multi-queue.yml` ? and then use the `.env` to join them like we do for optional analyzer services. I think this would also be more maintainable in the long run

We should do this for celery in this pull request only, will you be able to do this @0ssigeno ?

mlodic:
yep, I agree, one step at a time

0ssigeno:
Do you prefer to split the configuration file before merge this PR?
If so, I can work on that in a separated PR

Eshaan7:
I think we can do it in a seperate PR after merging this one. I say this because this commit/PR Is already very big as of now. You can fix the LGTM alert and mark the PR ready for review.

PR #278, opened 2020-12-02
Stratosphere blacklist

### LINKED ISSUE #216: Add Stratosphere prioritzed blacklists
These researchers created a public and very smart IP blocklists. Blog: https://www.stratosphereips.org/attacker-ip-prioritization-blacklist

We could create a new analyzer for these lists (all 3). 

We should also create an "updater" function that, daily, get the new lists (see this example: [Talos analyzer](https://github.com/intelowlproject/IntelOwl/blob/master/api_app/script_analyzers/observable_analyzers/talos.py)). 
We should try to avoid the flood of updating requests to their site caused by this updater. This means that I suggest to add an environment variable that allows people to activate this analyzer (and its updates) manually, only if they want to use it. By default, both analyzer and updater should be disabled and the analyzer, if executed, should show a message explaining that the analyzer must be activated manually.

Rishabh-Kumar-07:
# Description

Adding new analyzer to cross-reference the IP from blacklist provided by Stratosphere Labs.

## Fixes
#216 

## Type of change

- [+] New feature (non-breaking change which adds functionality).

# Checklist

- [+] The pull request is for the branch develop
- [+] I changed the file [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md) and [ReadMe](https://github.com/intelowlproject/IntelOwl/blob/master/README.md) if I added a new analyzer. 
- [+] I have added tests in the [Tests](https://github.com/intelowlproject/IntelOwl/blob/master/tests) folder. 
- [+] The tests gave 0 errors.
- [+] `Black` gave 0 errors.
- [+] `Flake` gave 0 errors.
- [+] I squashed the commits into a single one.

# Real World Example

image
image
image

### DIFF
--- a/api_app/script_analyzers/observable_analyzers/stratosphere.py
+++ b/api_app/script_analyzers/observable_analyzers/stratosphere.py
@@ -1,7 +1,7 @@
 import os
 import logging
-import traceback
 import requests
+import itertools
 
 from api_app.exceptions import AnalyzerRunException
 from api_app.script_analyzers import classes
@@ -24,56 +24,48 @@ class Stratos(classes.ObservableAnalyzer):
         ip = self.observable_name
         result = {
             "last24hrs": False,
-            "last24hrs_rating": "_",
+            "last24hrs_rating": "",
             "new_attacker": False,
-            "new_attacker_rating": "_",
+            "new_attacker_rating": "",
             "repeated_attacker": False,
-            "repeated_attacker_rating": "_",
+            "repeated_attacker_rating": "",
         }
 
         self.check_dataset_status()
 
-        with open(db_loc0, "r") as f:
-            db = f.read()
-        db_list = db.split("\n")
-        count = 0
-        for ip_tuple in db_list:
-            if count < 2:
-                count += 1
-                continue
-            else:
-                if ip in ip_tuple:
-                    ip_rating = ((ip_tuple.split(","))[2]).strip()
-                    result["last24hrs"] = True
-                    result["last24hrs_rating"] = ip_rating
+        with open(db_loc0, "r") as f0, open(db_loc1, "r") as f1, open(
+            db_loc2, "r"
+        ) as f2:
+            db0 = f0.read()
+            db1 = f1.read()
+            db2 = f2.read()
 
-        with open(db_loc1, "r") as f:
-            db = f.read()
-        db_list = db.split("\n")
-        count = 0
-        for ip_tuple in db_list:
-            if count < 2:
-                count += 1
-                continue
-            else:
-                if ip in ip_tuple:
-                    ip_rating = ((ip_tuple.split(","))[2]).strip()
-                    result["new_attacker"] = True
-                    result["new_attacker_rating"] = ip_rating
+        db_list0 = db0.split("\n")
+        db_list1 = db1.split("\n")
+        db_list2 = db2.split("\n")
 
-        with open(db_loc2, "r") as f:
-            db = f.read()
-        db_list = db.split("\n")
         count = 0
-        for ip_tuple in db_list:
+        for (ip0, ip1, ip2) in itertools.zip_longest(db_list0, db_list1, db_list2):
             if count < 2:
                 count += 1
                 continue
             else:
-                if ip in ip_tuple:
-                    ip_rating = ((ip_tuple.split(","))[2]).strip()
-                    result["repeated_attacker"] = True
-                    result["repeated_attacker_rating"] = ip_rating
+                if ip0:
+                    if ip in ip0:
+                        ip_rating = ((ip0.split(","))[2]).strip()
+                        result["last24hrs"] = True
+                        result["last24hrs_rating"] = ip_rating
+                if ip1:
+                    if ip in ip1:
+                        ip_rating = ((ip1.split(","))[2]).strip()
+                        result["new_attacker"] = True
+                        result["new_attacker_rating"] = ip_rating
+
+                if ip2:
+                    if ip in ip2:
+                        ip_rating = ((ip2.split(","))[2]).strip()
+                        result["repeated_attacker"] = True
+                        result["repeated_attacker_rating"] = ip_rating
 
         return result
 
@@ -112,7 +104,7 @@ class Stratos(classes.ObservableAnalyzer):
             logger.info("ended download of dataset from stratosphere")
 
         except Exception as e:
-            traceback.print_exc()
+            logger.debug("Traceback %s", exc_info=True)
             logger.exception(e)
 
         db_location = [db_loc0, db_loc1, db_loc2]
@@ -128,7 +120,8 @@ class Stratos(classes.ObservableAnalyzer):
         dt_object = datetime.fromtimestamp(timestamp)
 
         if (
-            today.day == dt_object.day
+            dt_object.hour > 3
+            and today.day == dt_object.day
             and today.month == dt_object.month
             and today.year == dt_object.year
         ):
--- a/api_app/script_analyzers/observable_analyzers/stratosphere.py
+++ b/api_app/script_analyzers/observable_analyzers/stratosphere.py
@@ -1,0 +1,140 @@
+import os
+import logging
+import traceback
+import requests
+
+from api_app.exceptions import AnalyzerRunException
+from api_app.script_analyzers import classes
+from intel_owl import settings
+from datetime import datetime, date
+
+logger = logging.getLogger(__name__)
+
+db_name0 = "stratos_ip_blacklist_last24hrs.csv"
+db_name1 = "stratos_ip_blacklist_new_attacker.csv"
+db_name2 = "stratos_ip_blacklist_repeated_attacker.csv"
+
+db_loc0 = f"{settings.MEDIA_ROOT}/{db_name0}"
+db_loc1 = f"{settings.MEDIA_ROOT}/{db_name1}"
+db_loc2 = f"{settings.MEDIA_ROOT}/{db_name2}"
+
+
+class Stratos(classes.ObservableAnalyzer):
+    def run(self):
+        ip = self.observable_name
+        result = {
+            "last24hrs": False,
+            "last24hrs_rating": "_",
+            "new_attacker": False,
+            "new_attacker_rating": "_",
+            "repeated_attacker": False,
+            "repeated_attacker_rating": "_",
+        }
+
+        self.check_dataset_status()
+
+        with open(db_loc0, "r") as f:
+            db = f.read()
+        db_list = db.split("\n")
+        count = 0
+        for ip_tuple in db_list:
+            if count < 2:
+                count += 1
+                continue
+            else:
+                if ip in ip_tuple:
+                    ip_rating = ((ip_tuple.split(","))[2]).strip()
+                    result["last24hrs"] = True
+                    result["last24hrs_rating"] = ip_rating
+
+        with open(db_loc1, "r") as f:
+            db = f.read()
+        db_list = db.split("\n")
+        count = 0
+        for ip_tuple in db_list:
+            if count < 2:
+                count += 1
+                continue
+            else:
+                if ip in ip_tuple:
+                    ip_rating = ((ip_tuple.split(","))[2]).strip()
+                    result["new_attacker"] = True
+                    result["new_attacker_rating"] = ip_rating
+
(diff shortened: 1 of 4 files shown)


api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,140 @@
...
+            with open(db_loc2, "w") as f:
+                f.write(r.content.decode())
+
+            if not os.path.exists(db_loc0 or db_loc1 or db_loc2):
+                raise AnalyzerRunException("failed extraction of stratosphere dataset")
+
+            logger.info("ended download of dataset from stratosphere")
+
+        except Exception as e:
+            traceback.print_exc()
```
mlodic, line 115: this print should be used only for debugging purposes

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,140 @@
...
+        timestamp = os.path.getctime(db_loc0)
+        dt_object = datetime.fromtimestamp(timestamp)
+
+        if (
+            today.day == dt_object.day
+            and today.month == dt_object.month
+            and today.year == dt_object.year
+        ):
+            logger.info("Dataset is up to date")
+        else:
```
mlodic, line 136: This is ok but the problem is that these blacklists are updated usually some hours after the start of the day (UTC). This would mean that at the start of the new day, this would download the blacklist of the day before because it was not already updated.
If possible, I would add a reasonable hour when we expect those blacklists are already been updated and force the updater there. I think that 4 am UTC could be a reasonable date. (I saw that they seem to update those lists between 1 am and 2 am)

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,140 @@
...
+            "last24hrs_rating": "_",
+            "new_attacker": False,
+            "new_attacker_rating": "_",
+            "repeated_attacker": False,
+            "repeated_attacker_rating": "_",
+        }
+
+        self.check_dataset_status()
+
+        with open(db_loc0, "r") as f:
```
mlodic, line 36: please try to avoid the duplication of code for this

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class Stratos(classes.ObservableAnalyzer):
+    def run(self):
+        ip = self.observable_name
+        result = {
+            "last24hrs": False,
+            "last24hrs_rating": "_",
+            "new_attacker": False,
+            "new_attacker_rating": "_",
```
mlodic, line 29: imho it is better an empty string than a placeholder to allow possible integrations to work better (empty string == boolean False). In this way the other boolean fields are not useful anymore because it is possible to check whether the IP was found or not just by checking the rating string

mlodic (changes requested):
good start, some adjusts are needed

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,140 @@
...
+            logger.info("starting download of dataset from stratosphere")
+
+            base_url = "https://mcfp.felk.cvut.cz"
+            mid_url = "/publicDatasets/CTU-AIPP-BlackList/Todays-Blacklists/"
+            url0 = base_url + mid_url + "AIP_blacklist_for_IPs_seen_last_24_hours.csv"
+            priority_url = "AIP_historical_blacklist_prioritized_by_"
+            url1 = base_url + mid_url + priority_url + "newest_attackers.csv"
+            url2 = base_url + mid_url + priority_url + "repeated_attackers.csv"
+
+            p = requests.get(url0, verify=False)
```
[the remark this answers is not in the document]
Rishabh-Kumar-07: No it is not working with validation.

api_app/script_analyzers/observable_analyzers/stratosphere.py
Rishabh-Kumar-07: I have changed it to print only during debugging.

api_app/script_analyzers/observable_analyzers/stratosphere.py
Rishabh-Kumar-07: Changed it to verify whether the file is downloaded before or after 4 am. If downloaded before 4 am it would force the updater to download again.

api_app/script_analyzers/observable_analyzers/stratosphere.py
Rishabh-Kumar-07: Changed the code to have minimum duplication.

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,133 @@
...
+            "last24hrs_rating": "",
+            "new_attacker": False,
+            "new_attacker_rating": "",
+            "repeated_attacker": False,
+            "repeated_attacker_rating": "",
+        }
+
+        self.check_dataset_status()
+
+        with open(db_loc0, "r") as f0, open(db_loc1, "r") as f1, open(
```
mlodic, line 36: you have to use a specific method for all this code of the function from now on to avoid duplication of code and increment the readability. This is extremely important

api_app/script_analyzers/observable_analyzers/stratosphere.py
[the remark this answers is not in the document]
mlodic: yes it seems there are some problems in their certificate configuration. I have just contacted them and see if they can solve it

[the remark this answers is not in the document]
mlodic: however also this part of code contains duplicated code and should be changed

mlodic (changes requested):

api_app/script_analyzers/observable_analyzers/stratosphere.py
Rishabh-Kumar-07: I have changed to do this checking in a function. Please check once.

api_app/script_analyzers/observable_analyzers/stratosphere.py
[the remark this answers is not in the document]
Rishabh-Kumar-07: I have checked. The problem with certificate configuration is still there. Created a function to remove duplicated code._

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,123 @@
...
+
+        ans = self.check_in_list(db_loc2, ip)
+        if(ans[0]):
+            result["repeated_attacker"] = True
+            result["repeated_attacker_rating"] = ans[1]
+
+        return result
+
+    def download_dataset(self, url, db_loc):
+        p = requests.get(url, verify=False)
```
mlodic, line 70: until they fix their problems, we'll keep it like that and we are fine. 

However could you please add a comment over this to remember *why* we are checking the certificate? Thank you

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,123 @@
...
+
+class Stratos(classes.ObservableAnalyzer):
+    def check_in_list(self,dataset_loc, ip):
+        with open(dataset_loc, "r") as f:
+            db = f.read()
+
+        db_list = db.split("\n")
+
+        count = 0
+        for ip_tuple in db_list:
```
mlodic, line 29: you could use "enumerate" to write this in a more pythonic way

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,123 @@
...
+db_loc1 = f"{settings.MEDIA_ROOT}/{db_name1}"
+db_loc2 = f"{settings.MEDIA_ROOT}/{db_name2}"
+
+
+class Stratos(classes.ObservableAnalyzer):
+    def check_in_list(self,dataset_loc, ip):
+        with open(dataset_loc, "r") as f:
+            db = f.read()
+
+        db_list = db.split("\n")
```
mlodic, line 26: could you write an example of what we should expect from this list as a comment? That would help future contributors

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,123 @@
...
+        count = 0
+        for ip_tuple in db_list:
+            if count < 2:
+                count += 1
+                continue
+            else:
+                if ip in ip_tuple:
+                    ip_rating = ((ip_tuple.split(","))[2]).strip()
+                    return(True, ip_rating)
+        return(False, "")
```
mlodic, line 37: You could write `return False, ""` and extract it `found, rate = self.check_in_list(db_loc0, ip)`.

Or better, you could avoid to use the boolean value at all. The empty string is already "False".

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def run(self):
+        ip = self.observable_name
+        result = {
+            "last24hrs": False,
```
mlodic, line 42: boolean values are unecessary as I have already told...

mlodic (changes requested):

api_app/script_analyzers/observable_analyzers/stratosphere.py
Rishabh-Kumar-07: Mentioned the comment.

api_app/script_analyzers/observable_analyzers/stratosphere.py
Rishabh-Kumar-07: I have removed the boolean value.

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -20,53 +20,40 @@
...
-            else:
-                if ip in ip_tuple:
-                    ip_rating = ((ip_tuple.split(","))[2]).strip()
-                    return(True, ip_rating)
-        return(False, "")
+        for ip_tuple in enumerate(db_list):
+            if ip_tuple[0] >= 2:
+                if ip in ip_tuple[1]:
+                    ip_rating = ((ip_tuple[1].split(","))[2]).strip()
+                    return(ip_rating)
```
mlodic, line 33: parenthesis are not required here in `return`

api_app/script_analyzers/observable_analyzers/stratosphere.py
```diff
@@ -20,53 +20,40 @@
...
-        ans = self.check_in_list(db_loc2, ip)
-        if(ans[0]):
-            result["repeated_attacker"] = True
-            result["repeated_attacker_rating"] = ans[1]
+        #Checks the IP in last24hrs attacker list.
+        result["last24hrs_rating"] = self.check_in_list(db_loc0, ip)
+        #Checks the IP in new attacker list.
+        result["new_attacker_rating"] = self.check_in_list(db_loc1, ip)
+        #Checks the IP in repeated attacker list.
+        result["repeated_attacker_rating"] = self.check_in_list(db_loc2, ip)
```
mlodic, line 51: this is really better 👍🏻

mlodic (changes requested):
The `black` problem in Travis + the last mini adjust and I guess we are done! 

With the recent improvements the code looks really much cleaner.

api_app/script_analyzers/observable_analyzers/stratosphere.py
Rishabh-Kumar-07: Thanks for all the advices. Will keep them in mind for next PR.

mlodic:
Please `git rebase develop` to get recent changes on Continuous Integration management (Travis dropped support for OS projects so we had to move on). In this way, checks will trigger again with the new configuration. Thank you

if you notice Travis CI check is missing :(

Rishabh-Kumar-07:
> Please `git rebase develop` to get recent changes on Continuous Integration management (Travis dropped support for OS projects so we had to move on). In this way, checks will trigger again with the new configuration. Thank you
> 
> if you notice Travis CI check is missing :(

PR #280, opened 2020-12-04
Moved stringsifter analyzer to Static Analyzers container

CITIZENDOT:
# Description

Also added ranking options to FLOSS

## Type of change

# Checklist

- [ ] I changed the file [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md) and [ReadMe](https://github.com/intelowlproject/IntelOwl/blob/master/README.md) if I added a new analyzer. 
- [X] `Black` gave 0 errors.
- [X] `Flake` gave 0 errors.

# Real World Example

```json
"additional_config_params": {
     "max_no_of_strings": {
       "stack_strings": 1000,
       "static_strings": 1000,
       "decoded_strings": 1000
     },
     "rank_strings": {
      "stack_strings": false,
      "static_strings": false,
      "decoded_strings": false
     }
},
```
Additional Config for FLOSS is changed as above.

`static_analyzers` docker currently hosts 5 Analyzers.
- Capa_Info
- PEframe_Scan
- Floss
- Strings_Info_Classic
- Strings_Info_ML


api_app/script_analyzers/file_analyzers/floss.py
```diff
@@ -25,4 +36,23 @@ def run(self):
             "callback_context": {"read_result_from": fname},
         }
         req_files = {fname: binary}
-        return self._docker_run(req_data, req_files)
+        result = self._docker_run(req_data, req_files)
+        result["exceeded_max_number_of_strings"] = {}
+
+        self.url = "http://static_analyzers:4002/stringsifter"  # For Ranking
```
Eshaan7, line 42: define this at the very top just under the class name

requirements.txt
```diff
@@ -68,7 +68,6 @@ scikit-learn==0.21.3
 six==1.13.0
 smmap2==2.0.5
 sqlparse==0.2.4
```
Eshaan7, line 70: Here, removing `stringsifter` won't be enough. We should also remove the entries for all it's sub-dependencies like numpy, etc. Should use something like pipdeptree (https://pypi.org/project/pipdeptree/) to inspect.

.env
```diff
@@ -10,7 +10,7 @@
 INTELOWL_TAG_VERSION=v1.9.1
 
 ###### Default (Production) ######
-COMPOSE_FILE=docker-compose.yml
+# COMPOSE_FILE=docker-compose.yml
```
Eshaan7, line 13: revert all changes in this file

api_app/script_analyzers/file_analyzers/floss.py
CITIZENDOT: Initially, FLOSS url should be used, After getting list of strings, then URL needs to be changed. However, It can be moved to the start of `run()` function.

Eshaan7: why not just declare 2 different variables at the top of the class then

CITIZENDOT: `self._docker_run()` extracts url from `self.url`.

Initial Run of `self.docker_run()` is required for FLOSS strings. 

Second Run is for rank_strings. (Had to change URL before running this).

Eshaan7:
Can you fix the codefactor alerts https://www.codefactor.io/repository/github/intelowlproject/intelowl/pull/280 ?
when doing `popen`, we should include the full-path of the binaries and not the relative path as a security measure.

api_app/script_analyzers/file_analyzers/strings_info.py
```diff
@@ -24,58 +27,32 @@ def set_config(self, additional_config_params):
...
+            ]
+            req_data = {"args": args, "timeout": self.timeout}
+            result = self._docker_run(req_data)
+        elif len(result) > self.max_no_of_strings:
+            result = [s for s in result[: self.max_no_of_strings]]
+            result = {
+                "data": result,
+                "exceeded_max_number_of_strings": True,
+            }  # Normally returns List, But now returns dict
+        return result
```
mlodic, line 58: Could we have the same kind of output in *any* case? Users or integrations would expect always the same format. In this way we do not risk to have people confused

api_app/script_analyzers/file_analyzers/strings_info.py
```diff
@@ -24,58 +27,32 @@ def set_config(self, additional_config_params):
...
+                "rank_strings",
+                "--limit",
+                str(self.max_no_of_strings),
+                "--strings",
+                json_dumps(result),
+            ]
+            req_data = {"args": args, "timeout": self.timeout}
+            result = self._docker_run(req_data)
+        elif len(result) > self.max_no_of_strings:
+            result = [s for s in result[: self.max_no_of_strings]]
```
mlodic, line 53: I think you forgot to use *max_chars_for_string* attribute. This is extremely important to reduce the possible output

mlodic (changes requested):
good job, we are almost done I guess

api_app/script_analyzers/file_analyzers/strings_info.py
CITIZENDOT: Yeah, Sure

Eshaan7:
@CITIZENDOT if there's no change left, I think we can go ahead and merge this.

Eshaan7 (approved):

PR #279, opened 2020-12-03
Used docker overwrite feature

0ssigeno:
# Description
Docker allows to overwrite services, and we did not used this feature before
## Fixes
Please add related issues.
#277 
## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue).

# Checklist

- [x] The pull request is for the branch develop
- [x] The tests gave 0 errors.
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.
- [x] I squashed the commits into a single one.

PR #283, opened 2020-12-08
Added triage observable analyzer

m0mosenpai:
# Description

Added triage analyzer for scanning observables - searching through existing reports via URL or Hash query; or submitting a sample for analysis via a URL. Also added documentation for both triage observable and file analyzers.

## Fixes
closes #142 

## Type of change
- [x] New feature (non-breaking change which adds functionality).

# Checklist

- [x] The pull request is for the branch develop
- [x] I changed the file [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md) and [ReadMe](https://github.com/intelowlproject/IntelOwl/blob/master/README.md) if I added a new analyzer. 
- [x] I have added tests in the [Tests](https://github.com/intelowlproject/IntelOwl/blob/master/tests) folder. 
- [x] The tests gave 0 errors.
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.
- [x] I squashed the commits into a single one.

# Real World Example
1. `submit` config paramenter - submitting file via url (similar report generated as triage file analyzer)
image

2. `search` config parameter - searching for reports via hash.
image
50 results (default) (max 200) in same format as above followed by a `next` key. I haven't implemented pagination using `next` as of now. A bit confused on how to go about it since next contains a date-time stamp of sorts. 
image

3. `search` config parameter - searching for reports via URL.
I couldn't find a good example of this (a malicious url) but report should be similar to the hash case.

PR #282, opened 2020-12-08
Added InQuest Analyzers

CITIZENDOT:
# Description

Added InQuest Analyzer

## Fixes
closes #200 

## Type of change

- [X] New feature (non-breaking change which adds functionality).

# Checklist

- [X] The pull request is for the branch develop
- [X] I changed the file [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md) and [ReadMe](https://github.com/intelowlproject/IntelOwl/blob/master/README.md) if I added a new analyzer. 
- [X] I have added tests in the [Tests](https://github.com/intelowlproject/IntelOwl/blob/master/tests) folder. 
- [X] The tests gave 0 errors.
- [X] `Black` gave 0 errors.
- [X] `Flake` gave 0 errors.
- [ ] I squashed the commits into a single one.

# Real World Example
InQuest

PS: Most of the work is done by @i-karen. I did few edits suggested in that PR.

PR #289, opened 2020-12-10
docker/-compose refactoring + start.py script with docs

### LINKED ISSUE #98: Divide requirements.txt into development.txt and production.txt
We could separate the `requirements.txt` file into small files such as development/local.txt and requirements/production.txt.

_Originally posted by @Andrew-Chen-Wang in https://github.com/intelowlproject/IntelOwl/issues/95#issuecomment-647153922_

### LINKED ISSUE #277: Divide common docker services into diff. compose files
>  What I meant is basically your idea but applied to all the services, not only the ones related to celery.

Ah, I see. Yes, definitely what I meant as well. So, let's create a `compose.common.yml` file with the services like `rabbit-mq`, `postgres` which are common in all compose files and similarly for celery.

_Originally posted by @Eshaan7 in https://github.com/intelowlproject/IntelOwl/pull/274#issuecomment-737187816_

0ssigeno:
# Description
Moved docker files in a subdirectory
Created start.py
Fixed docker integrations files
Added documentation

## Fixes
extended from #279 
closes #277
closes #98
## Type of change

# Checklist

- [x] I changed the file [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md) and [ReadMe](https://github.com/intelowlproject/IntelOwl/blob/master/README.md) if I added a new analyzer. 
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.


docs/source/Tests.md
```diff
@@ -19,7 +19,7 @@
 ### Launch tests
     
 - Start the app using the docker-compose test file. In this way, you would launch the code in your environment and not the last official image in Docker Hub:
-`docker-compose -f docker-compose-for-tests.yml up`
+`docker-compose -f docker/default.yml -f docker/test.yml up`
```
Eshaan7, line 22: correction: should be: `docker/test.override.yml`

start.py
```diff
@@ -0,0 +1,92 @@
...
+    context_settings=dict(
+        ignore_unknown_options=True,
+        allow_extra_args=True,
+    )
+)
+@click.argument("mode", required=True, type=click.Choice(["prod", "test", "ci"]))
+@click.argument(
+    "docker_command", required=True, type=click.Choice(["build", "up", "down"])
+)
+@click.option("--traefik", default=False, is_flag=True)
```
Eshaan7, line 39: really think we should provide some help with each option. For example, `help="Uses the `traefik.override.yml` compose file"` and similarly for other options.

mlodic:
I think we should merge this after having pushed #282, #290 and #283. New PRs will need to be based from this refactor

requirements.txt
```diff
@@ -67,8 +67,6 @@ quark-engine==20.12
 git+https://github.com/DissectMalware/XLMMacroDeobfuscator.git@89fbce0c87014a4b5a22c1aef09c8b3ea9bf16c0
 speakeasy-emulator==1.4.8
 sphinx_rtd_theme==0.5.0
-flake8==3.8.4
-black==20.8b1
 pre-commit==2.9.2
```
Eshaan7, line 70: sphinx_rtd_theme and pre-commit are also part of testing/development

docker/docker_entrypoint.sh
```diff
@@ -6,4 +6,6 @@
 python manage.py migrate                  # Apply database migrations
 python manage.py collectstatic --noinput  # Collect static files
 
+bash api_app/script_analyzers/yara_repo_downloader.sh
```
Eshaan7, line 9: should revert this change. The yara/quark-engine rules are to be kept constant for a particular IntelOwl release (otherwise, there are bugs). If we make this script part of entrypoint, then it gets executed each time the container is started thus slowing it down by a lot.

mlodic: that's right... the idea was to reduce the size of the base image + reduce the build time of the image.

I would like to avoid to load the rules in any case at the start. I think they should be loaded only the first time they are requested and only if the users want to use them. Another motivation: the number of Yara repositories is growing a lot so I do not think they should be managed in the base image. It does not scale. To solve this we would need an additional work that is maybe out of the scope of the PR.

We could just create an issue for this and revert this change. Thoughts?

Eshaan7: > For now, we could just create an issue for this and revert this change.

I agree.

> To solve this we would need an additional work that is maybe out of the scope of the PR.

Yeah

requirements.txt
0ssigeno: You don't have a `.readthedocs.yml` file. 
If i move the sphinx import, the readthedocs should not build anymore.

Eshaan7:
Finally, I think everything is satisfactory. :D
Very nice work, @0ssigeno. Thank you for your patience through every iteration. 

Before we merge this PR, there's still 1 change left. After we have merged the other open PRs, we would need to modify the `docker/env_file_app_template` and `docker/env_file_app_ci`  to add entries for the newly added variables and cross-check that no variable is left.

0ssigeno:
I did not found any problem during the merge, so I think everything is fine. If you have time, you can doublecheck the environ variables, just to be sure.
If problems for readthedocs arise after the merge, I will try to fix them asap.

.readthedocs.yml
```diff
@@ -0,0 +1,24 @@
...
+  image: latest
+
+# Build all formats
+formats: all
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.7
+  install:
+    - requirements: docs/requirements.txt
```
Eshaan7, line 18: this is set to `docs/requirements.txt`

docs/source/requirements.txt
```diff
@@ -1 +1 @@
-# This requirements.txt is intentionally left blank for RTD build
```
Eshaan7, line 1: the path here is `docs/source/requirements.txt`. I'd prefer to move this file to `docs/requirements.txt`.

PR #290, opened 2020-12-10
Added Manalyze Analyzer | #264

### LINKED ISSUE #264: New analyzer: Manalyze
A static analyzer for PE executables : https://github.com/JusticeRage/Manalyze

CITIZENDOT:
# Description

Added Manalyze Analyzer

## Fixes
closes #264 

## Type of change

# Checklist

- [X] I changed the file [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md) and [ReadMe](https://github.com/intelowlproject/IntelOwl/blob/master/README.md) if I added a new analyzer. 
- [X] `Black` gave 0 errors.
- [X] `Flake` gave 0 errors.

# Real World Example
Manalyze

### DIFF
--- a/api_app/script_analyzers/file_analyzers/manalyze.py
+++ b/api_app/script_analyzers/file_analyzers/manalyze.py
@@ -1,0 +1,28 @@
+from api_app.helpers import get_binary
+from api_app.script_analyzers.classes import FileAnalyzer, DockerBasedAnalyzer
+
+
+class Manalyze(FileAnalyzer, DockerBasedAnalyzer):
+    name: str = "Manalyze"
+    url: str = "http://static_analyzers:4002/manalyze"
+    # interval between http request polling
+    poll_distance: int = 10
+    # http request polling max number of tries
+    max_tries: int = 60
+    # here, max_tries * poll_distance = 10 minutes
+    timeout: int = 60 * 9
+    # whereas subprocess timeout is kept as 60 * 9 = 9 minutes
+
+    def run(self):
+        # get binary
+        binary = get_binary(self.job_id)
+        # make request data
+        fname = str(self.filename).replace("/", "_").replace(" ", "_")
+        args = [f"@{fname}", "--output", "json"]
+        req_data = {
+            "args": args,
+            "timeout": self.timeout,
+        }
+        req_files = {fname: binary}
+        report = self._docker_run(req_data, req_files)
+        return report.values()
--- a/configuration/analyzer_config.json
+++ b/configuration/analyzer_config.json
@@ -422,6 +422,13 @@
     },
     "soft_time_limit": 30
   },
+  "Manalyze": {
+    "type": "file",
+    "supported_filetypes": ["application/x-dosexec"],
+    "description": "A static analyzer for PE files",
+    "python_module": "manalyze.Manalyze",
+    "soft_time_limit": 720
+  },
   "MISP": {
     "type": "observable",
     "observable_supported": [
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ license terms.
 [IntelX](https://intelx.io/terms-of-service),
 [Speakeasy](https://github.com/fireeye/speakeasy),
 [Checkdmarc](https://github.com/domainaware/checkdmarc)
+[Manalyze](https://github.com/JusticeRage/Manalyze)
 
 ## Acknowledgments
 
--- a/docs/source/Advanced-Usage.md
+++ b/docs/source/Advanced-Usage.md
@@ -27,13 +27,15 @@ table, th, td {
   </tr>
   <tr>
     <td>Static Analyzers</td>
-    <td><code>PEframe_Scan</code>, <code>Capa_Info</code>, <code>Floss</code></td>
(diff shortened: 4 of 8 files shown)


api_app/script_analyzers/file_analyzers/manalyze.py
```diff
@@ -0,0 +1,28 @@
...
+from api_app.script_analyzers.classes import FileAnalyzer, DockerBasedAnalyzer
+
+
+class Manalyze(FileAnalyzer, DockerBasedAnalyzer):
+    name: str = "Manalyze"
+    url: str = "http://static_analyzers:4002/manalyze"
+    # interval between http request polling
+    poll_distance: int = 10
+    # http request polling max number of tries
+    max_tries: int = 60
```
Eshaan7, line 11: 60 requests is a big number. We should change this to about 10. In your testing, did you find manalyze to take a lot of time processing a binary ?

configuration/analyzer_config.json
```diff
@@ -422,6 +422,12 @@
     },
     "soft_time_limit": 30
   },
+  "Manalyze": {
+    "type": "file",
+    "description": "A static analyzer for PE files",
+    "python_module": "manalyze.Manalyze",
+    "soft_time_limit": 500
```
Eshaan7, line 429: we should decrease the `max_tries` and then the `soft_time_limit` as well

api_app/script_analyzers/file_analyzers/manalyze.py
```diff
@@ -0,0 +1,28 @@
...
+    def run(self):
+        # get binary
+        binary = get_binary(self.job_id)
+        # make request data
+        fname = str(self.filename).replace("/", "_").replace(" ", "_")
+        args = [f"@{fname}", "--output", "json"]
+        req_data = {
+            "args": args,
+            "timeout": self.timeout,
+            "callback_context": {"read_result_from": fname},
```
Eshaan7, line 25: no need for the `callback_context` parameter here

api_app/script_analyzers/file_analyzers/manalyze.py
CITIZENDOT: No, It instantly gave the results. But it may take time on bigger binaries. I'm not sure what limit should I keep there.

api_app/script_analyzers/file_analyzers/manalyze.py
CITIZENDOT: Oh, I copy-pasted from other analyzer. Will change it now.

configuration/analyzer_config.json
CITIZENDOT: Okay, I am changing `max_tries` to 10 and `soft_time_limit` to 100 and `timeout` to 60.

api_app/script_analyzers/file_analyzers/manalyze.py
Eshaan7: I guess that's fair then. `60` requests is the worst case scenario.

configuration/analyzer_config.json
Eshaan7: `soft_time_limit` should be made equal to about 12 minutes. Since, the timeout is 9 minutes and total polling time is 10 minutes.

CITIZENDOT: I didn't quite understand it. I should decrease the `max_tries`, Right?. Could you give exact values of `soft_time_limit`, `max_tries` and `timeout` please?

Eshaan7: `soft_time_limit=720`; every other value would remain same as before

Eshaan7: >  Okay, I am changing max_tries to 10 and soft_time_limit to 100 and timeout to 60.

ah, just saw this. No, keep the values same as before.

CITIZENDOT: Yeah, I only changed `soft_time_limit` to 720.

Eshaan7:
need to add `Manalyze` analyzer info the optional analyzer's under advanced usage section of docs as well.

CITIZENDOT:
> need to add `Manalyze` analyzer to the optional analyzer's table under advanced usage section of docs as well.

I see that, I have to add `StringInfo_Classic` and `ML` too in that table. Shall I add in this PR? And We forgot to add `apk-analyzers` in `.env` file.

Eshaan7:
> I see that, I have to add StringInfo_Classic and ML too in that table. Shall I add them in this PR? And We forgot to add apk-analyzers in .env file.

Can u do both of this in a separate PR if it's not too much hassle ?

Eshaan7:
You may add the `StringsInfo_*` as part of this PR. We can leave the apk analyzers, as that will be overwritten as part of the other PR where we are refactoring docker.

mlodic (changes requested):

Eshaan7:
LGTM. Just one thing that caught my eye, in the output screenshot I see the results are stored inside a JSON object with key name equal to temporary filepath that is used in the static analyzers container. Can you modify the key to make it equal to just the filename of the original file that the user uploaded?

Eshaan7:
LGTM. Just one thing that caught my eye, in the output screenshot I see the results are stored inside a JSON object with key name equal to temporary filepath that is used in the static analyzers container. I would prefer to not have that key there, but just the nested results.
You will have to do something like:

```python
self.report = self.report[self.report.keys()[0]]
```

Issue #281, opened 2020-12-06
New analyzer: emailrep.io + "generic" classification


theguly:
hi there, i was about to starting an integration with emailrep.io when i found that observable classification doesn't let me to query for an email object.

maybe you're not interested in doing query against email address?

in case you are, do you think you could add that classification to js definition (i'm reluctant to dig into docker build script and minimized javascript, if possible...)?

thanks

mlodic:
You are right, ATM IntelOwl does not support emails as observables...there are already existing integrations that would benefit of this change (ZoomEye,...) but it is not yet supported. 

This is a recurrent request (analyze new types of observables) so I think that it is time to understand how to manage this new possible feature. 

I'd like to mention @Eshaan7 because we already talked about this. It would be nice if the Angular Interface would dynamically support new types of observables when we add new ones. Server side I guess we can just add new ones without particular problems. Also I do not think that PyIntelOwl client would be affected in any case too. Thoughts?

Eshaan7:
Don't really think the dynamic way would work although adding a new observable type on the angular client would take little effort. Same goes for pyintelowl. For both clients, we just need to add an entry for new the observable type and it's corresponding regular expression.

But as discussed before, I think we should add a new observable type called `generic` which would be a placeholder for observable types like email, city, phone number, location, etc. Basically, only certain analyzers allow any kind of observable so if user chose `generic` as the type, we just pass the value to the analyzer and see what happens.

@theguly so I suggest you to add `generic` to the `observables_supported` parameter. Soon, we will add support for the same.

mlodic:
Yes, in this way we would have the chance to analyze *everything* and we won't have to think about it again.

My only concern is one. For instance, one analyzer would like to change its behavior based on the observable classification (ex: emails and phone number use different endpoints): that specific analyzer would have to perform the "classification" of what kind of observable arrived on its own, while IntelOwl could have already done it automatically. At the moment, this is an edge case so I guess we can ignore my concern and go on like you proposed.

theguly:
ok i have my side done, waiting for generic classification to be available to submit a PR

thanks

mlodic:
I just added support for "generic" type of observables in the `develop` branch of `pyintelowl`. Please @theguly push your code/PR as a draft so we can start to test it with pyintelowl

PR #291, opened 2020-12-11
integrate a generic analyzer: emailrep.io

theguly:
closes https://github.com/intelowlproject/IntelOwl/issues/281

i'm quite sure tests hasn't been placed in the right way, but i think it could be handled after your integration


CITIZENDOT:
I think it's better to have a `test_generic.py` file, for these kinds of analyzers. Just my opinion.

api_app/script_analyzers/observable_analyzers/emailrep.py
```diff
@@ -0,0 +1,42 @@
+import requests
+
+from api_app.exceptions import AnalyzerRunException
+from api_app.script_analyzers import classes
+from intel_owl import secrets
+
+
+class EMAILREP(classes.ObservableAnalyzer):
```
Eshaan7, line 8: Can you rename this to `EmailRep` instead ?

api_app/script_analyzers/observable_analyzers/emailrep.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def run(self):
+        """
+        API key is not mandatory, emailrep supports requests with no key:
+        a valid key let you to do more requests per day.
+        therefore we're not checking if a key has been configured.
+        """
+
+        headers = {
+            "User-Agent": "Intelowl",
```
Eshaan7, line 23: rename to: `IntelOwl v2`

tests/utils.py
```diff
@@ -288,3 +289,19 @@ def test_thug_url(self, mock_get=None, mock_post=None):
             additional_params,
         ).start()
         self.assertEqual(report.get("success", False), True)
+
+class CommonTestCases_generic(metaclass=ABCMeta):
```
Eshaan7, line 293: move this class to `test_observables.py` file with name like `class GenericAnalyzersTest(TestCase):`

configuration/analyzer_config.json
```diff
@@ -227,6 +227,17 @@
     "soft_time_limit": 30,
     "queue": "local"
   },
+  "Emailrep": {
```
Eshaan7, line 230: rename this to `EmailRep`; same change in the docs

api_app/script_analyzers/observable_analyzers/emailrep.py
Eshaan7: then, you'd need to change module name in the analyzer_config as well

Eshaan7:
> I think it's better to have a test_generic.py file, for these kinds of analyzers. Just my opinion.

`generic` is to represent different classes of observables only, so we can just use the existing `test_observables.py` module only.

api_app/script_analyzers/observable_analyzers/emailrep.py
```diff
@@ -0,0 +1,42 @@
...
+    def run(self):
+        """
+        API key is not mandatory, emailrep supports requests with no key:
+        a valid key let you to do more requests per day.
+        therefore we're not checking if a key has been configured.
+        """
+
+        headers = {
+            "User-Agent": "IntelOwl v2",
+            "Key": self.__api_key,
```
Eshaan7, line 24: in case the `self.__api_key` is `None` (case when it's not defined by user), does passing a null API key to the URL does not give error ?

tests/utils.py
Eshaan7: also add this at the top of this new class,

```python
    @staticmethod
    def get_params():
        return {
            "source": "test",
            "is_sample": False,
            "observable_name": os.environ.get(
                "TEST_GENERIC", "email@example.com"
            ),
            "observable_classification": "generic",
            "force_privacy": False,
            "analyzers_requested": ["test"],
        }

```

env_file_app_ci
```diff
@@ -43,6 +43,7 @@ URLSCAN_API_KEY=test
 PHISHTANK_API_KEY=test
 ZOOMEYE_KEY=test
 TRIAGE_KEY=test
+EMAILREP_KEY=test
```
Eshaan7, line 46: nice! on line no. 54, can you add this `TEST_GENERIC=email@example.com`

Eshaan7:
make sure to run black and flake8 linters ([docs](https://intelowl.readthedocs.io/en/latest/Contribute.html#pass-linting-and-tests)) before pushing final changes.

api_app/script_analyzers/observable_analyzers/emailrep.py
CITIZENDOT: No, It did not give any errors with `None`.

theguly: as far as i tested, an empty key doesn't give any error

tests/utils.py
theguly: done

theguly:
note to self: squash all these small commit when we're done with testing

env_file_app_ci
Eshaan7: @theguly

tests/utils.py
```diff
@@ -17,6 +17,7 @@
     intelx,
     urlscan,
     mb_google,
+    emailrep,
```
Eshaan7, line 20: import not needed

tests/test_observables.py
```diff
@@ -554,3 +554,32 @@ def test_cymru_get(self, mock_get=None, mock_post=None):
...
+        "is_sample": False,
+        "observable_name": os.environ.get(
+            "TEST_GENERIC", "email@example.com"
+        ),
+        "observable_classification": "generic",
+        "force_privacy": False,
+        "analyzers_requested": ["test"],
+    }
+
+class CommonTestCases_generic(metaclass=ABCMeta):
```
Eshaan7, line 571: this should be `class GenericAnalyzersTest(TestCase):`

tests/test_observables.py
```diff
@@ -554,3 +554,32 @@ def test_cymru_get(self, mock_get=None, mock_post=None):
             {},
         ).start()
         self.assertEqual(report.get("success", False), True)
+
+@staticmethod
+def get_params():
```
Eshaan7, line 559: this should be inside the `GenericAnalyzersTest` class

tests/test_observables.py
Eshaan7: add a new function inside this class,

```python
    def setUp(self):
        params = self.get_params()
        params["md5"] = hashlib.md5(
            params["observable_name"].encode("utf-8")
        ).hexdigest()
        test_job = Job(**params)
        test_job.save()
        self.job_id = test_job.id
        self.observable_name = test_job.observable_name
        self.observable_classification = test_job.observable_classification
```

Eshaan7:
@theguly, I think everything looks good. Can you please squash all commits into one and run black and flake8 linters (docs) before pushing final changes.

Eshaan7:
@theguly Thankyou so much for your contribution!

PR #295, opened 2020-12-17
Fix log directory with docker integrations

0ssigeno:
# Description
I can't wrap my head around on why it happens but creating the log directories for the analyzers_integration inside the Dockerfile it is not effective. I think that this happens because we are attaching there a volume in the docker-compose file, but I'm not sure about the motivation. Nevertheless, we had `No such file or directory` happening. The solution that I provide is to force the creation of the log directory *inside* each entrypoint of each integration. Moreover, I think is cleaner to remove the creation of directories that the user may not use from the main Dockerfile.

## Type of change

## Steps to reproduce
- Run IntelOwl
- Delete directory /var/log/intel_owl/static_analyzer
- Build and run again intelowl
-  image
 

## Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,9 +8,8 @@ ENV ELASTICSEARCH_DSL_VERSION 7.1.4
 ENV watch_logs_cmd "watch -n1 tail -n10 /var/log/intel_owl/django/api_app.log"
 
 RUN mkdir -p ${LOG_PATH} \
-    ${LOG_PATH}/django ${LOG_PATH}/uwsgi \
-    ${LOG_PATH}/thug ${LOG_PATH}/static_analyzers ${LOG_PATH}/box-js \
-    ${LOG_PATH}/apk_analyzers \
+    ${LOG_PATH}/django \
+    ${LOG_PATH}/uwsgi \
     /opt/deploy/files_required /opt/deploy/yara /opt/deploy/configuration
 
 RUN apt-get update \
--- a/integrations/apk_analyzers/entrypoint.sh
+++ b/integrations/apk_analyzers/entrypoint.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-
+mkdir -p ${LOG_PATH}
 touch ${LOG_PATH}/gunicorn_access.log ${LOG_PATH}/gunicorn_errors.log
 chown -R apk-user:apk-user ${LOG_PATH}
 su apk-user -s /bin/bash
--- a/integrations/box-js/entrypoint.sh
+++ b/integrations/box-js/entrypoint.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-
+mkdir -p ${LOG_PATH}
 touch ${LOG_PATH}/gunicorn_access.log ${LOG_PATH}/gunicorn_errors.log
 chown -R boxjs-user:boxjs-user ${LOG_PATH}
 su boxjs-user -s /bin/sh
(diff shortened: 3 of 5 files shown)


Eshaan7:
> The solution that I provide is to force the creation of the log directory inside each entrypoint of each integration. 

This is what I initially did too when we first introduced the concept of optional containers. IIRC, it resulted in some issues with `insufficient permission` since we are creating a log directory and log files at a path which is then mounted to the named volume: `generic_logs`. 

Did you face any error of the same sort ?

Eshaan7:
> Steps to reproduce

Yes, it makes sense that this error would arise if we delete any log directory because the creation of these folders is part of the docker image's build steps and _not_ the entrypoint. 

As I pointed out the issue with permissions in my previous comment, if we cannot circumvent that then perhaps, it might be helpful to move the creation of these log folders to the main `entrypoint.sh` except for the individual ones.

0ssigeno:
I am not able to reproduce your error. Can you try

> > The solution that I provide is to force the creation of the log directory inside each entrypoint of each integration. Moreover, I think is cleaner to remove the creation of directories that the user may not use from the main Dockerfile.
> 
> This is what I initially did too when we first introduced the concept of optional containers. IIRC, it resulted in some issues with `insufficient permission` since we are creating a log directory and log files at a path which is then mounted to the named volume: `generic_logs`.
> 
> Did you face any error of the same sort ?

Unfortunately (or fortunately, it depends) I am not able to reproduce your issue

0ssigeno:
> > Steps to reproduce
> 
> Yes, it makes sense that this error would arise if we delete any log directory because the creation of these folders is part of the docker image's build steps and _not_ the entrypoint.
> 
> If we cannot circumvent the issue with permissions (the one I pointed in my previous comment) then perhaps, it might be helpful to move the creation of these log folders to the main `entrypoint.sh` and not the individual ones (like you have implemented in this PR).

Building the image again( we are building, not retrieving the image from dockerhub here) should not create an error if we removed the directories. Even adding `--no-cache` to force docker correctly rebuild the entire container is not enough to not have this error.

Eshaan7:
> I can't wrap my head around on why it happens

I feel the same way after reading,

>  Building the image again( we are building, not retrieving the image from dockerhub here) should not create an error if we removed the directories.

So I think we are left with 2 options:
 a) either create all log directories in the main `entrypoint.sh` (to avoid permissions issues)
 b) or force the creation of the log directory inside each entrypoint of each integration. (which you currently implemented)
 
In terms of cleanliness and overall better modularity, option B seems to win however I am critical about the permission issue since I remember that issue was quite frustrating for @mlodic and I both.

0ssigeno:
Agree on the testing, any feedback is welcome!

mlodic:
Hi guys, I have just tried with a clean setup with both the old version and this new version and got no problems in both cases.

Previously, I had the same problem @0ssigeno faced with those `static_analyzers` logs, once upgrading IntelOwl from an old version (without those logs) and a new one (with those logs). 

I guess that your (@Eshaan7 and @0ssigeno) environments are somehow different and this is why the issue is not easy to replicate. We (me and @Eshaan7) got similar problems some months ago and was really difficult to find out what was the real issue.

Anyway, I agree that creating the logs directories in the main Dockerfile could be the source of all these problems so we should change this at least. Also I guess that splitting the configuration for each analyzers (@0ssigeno implementation) is the best way to avoid conflicts of any kind because they are indeed separated. So I would try with this way for the next release and see how it goes.

Thoughts? (so we can close this aged PR)

PR #300, opened 2020-12-21
Fix strings_info.py

0ssigeno:
# Description
We had two cases that were not checked:
- If too many strings are sent to `Strings_info_ml`, the program crashes. So we check them
- If self.rank_strings==False AND len(result) < self.max_no_of_strings, result is still a list. Meaning that result.get("data", []) will crash the analyzer.

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

# Real World Example
image
image

### DIFF
--- a/api_app/script_analyzers/file_analyzers/strings_info.py
+++ b/api_app/script_analyzers/file_analyzers/strings_info.py
@@ -38,7 +38,9 @@ class StringsInfo(FileAnalyzer, DockerBasedAnalyzer):
         }
         req_files = {fname: binary}
         result = self._docker_run(req_data, req_files)
-
+        exceed_max_strings = len(result) > self.max_no_of_strings
+        if exceed_max_strings:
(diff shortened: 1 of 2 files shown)


api_app/script_analyzers/file_analyzers/strings_info.py
```diff
@@ -38,7 +38,9 @@ def run(self):
         }
         req_files = {fname: binary}
         result = self._docker_run(req_data, req_files)
-
+        exceed_max_strings = len(result) > self.max_no_of_strings
+        if len(result) > self.max_no_of_strings:
```
CITIZENDOT, line 42: You can replace the line 
```
if len(result) > self.max_no_of_strings:
```
 with 
```
if exceed_max_strings:
```

0ssigeno: Yeah, ctrl-c/ctrl-v typo

Issue #229, opened 2020-10-16
More robust testing for the REST API/ more testcases


Eshaan7:
1. 1. [ ] Instead of manually defining the analyzer names like: [ref 1](https://github.com/intelowlproject/IntelOwl/blob/b8adb567171f6c59d5cf0dc8ca2c445af25191f4/tests/test_api.py#L145), [ref 2](https://github.com/intelowlproject/IntelOwl/blob/b8adb567171f6c59d5cf0dc8ca2c445af25191f4/tests/test_api.py#L78) - we can just pass `run_all_available_analyzers: true` in POST request body. (_not sure, might change this_)
2. 2. [ ] Currently, the `script_analyzers.general.start_analyzers` function is not called during tests. This really decreases the overall code coverage and also misses out on spotting many errors especially those that have to do with some typo or mistake in the analyzer_config entries. So, we should make changes in a way that detects such errors too.
3. 3. [x] Add `codecov` ([ref](https://docs.codecov.io/docs/python)) support to the GitHub actions and for [pull requests](https://docs.codecov.io/docs/pull-request-comments) too. (PR: #305)
4. 4. [x] Need to add tests for the `/api/jobs` , `/api/jobs/{id}` (GET) and `/api/tags`, `/api/tags/{id}` (GET, POST, PUT, DELETE) endpoints inside [`tests/test_api.py` ](https://github.com/intelowlproject/IntelOwl/blob/master/tests/test_api.py): `ApiTests` class. Refer [here](https://www.django-rest-framework.org/api-guide/testing/). (PR: #317)
5. 5. [ ] It would also be really good if we are able to add tests for verifying permissions/groups. For example, enabling/disabling certain permissions to a group/user and then making request to a protected endpoint and verify that the behavior changed.
6. 6. [x] Add tests for `DELETE /api/jobs/{id}`, `PATCH /api/jobs/{id}/kill`, `GET /api/get_analyzer_configs` and `GET /api/download_sample` endpoints as well. (PR: #317)

mlodic:
We could take this chance to add a code coverage linter for CI to evaluate how much our tests are good

m0mosenpai:
I would like to work on this. Things to keep in mind/ starting suggestions would be great if possible.

Eshaan7:
@m0mosenpai,
[...]

PR #305, opened 2020-12-30
Added codecov.io coverage support

m0mosenpai:
# Description

Added necessary files and settings for codecov.io support for coverage reports as well as pull requests.

## Fixes
Point 3 in #229

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

# Real World Example
image

Screenshot of a report generated from a sample PR I made on my own fork of this repo.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,20 @@
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "reach,diff,flags,files,footer"
+  behavior: default
+  require_changes: true
--- a/requirements.txt
+++ b/requirements.txt
@@ -72,3 +72,4 @@ dnstwist==20201022
 ppdeep>=20200505
 google>=3.0.0
 eventlet==0.29.1
+django-nose==1.4.7
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,13 @@
+[run]
+branch = True
+source = IntelOwl
+
+[report]
+exclude_lines =
+    if self.debug:
+    pragma: no cover
+    raise NotImplementedError
+    if __name__ == .__main__.:
+ignore_errors = True
+omit =
+    tests/*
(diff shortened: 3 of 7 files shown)


docker/default.yml
```diff
@@ -23,6 +23,8 @@ services:
       - ./docker//docker_entrypoint.sh
     expose:
       - "8001"
+    environment:
+      - DJANGO_SETTINGS_MODULE=intel_owl.settings
```
mlodic, line 27: If this variable is required only for CI, it should be set in the specific docker-compose file (`ci.override.yml`)...or, even better, you can just insert this variable in the specific env_file: `env_file_app_ci`

requirements.txt
```diff
@@ -71,4 +71,6 @@ django-rest-durin==0.1.0
 dnstwist==20201022
 ppdeep>=20200505
 google>=3.0.0
-eventlet==0.29.1
\ No newline at end of file
+eventlet==0.29.1
+django-nose==1.4.7
+coverage==5.3.1
```
mlodic, line 76: are these new dependencies required for production? I guess not. Those should be removed from here

.codecov.yml
```diff
@@ -0,0 +1,20 @@
...
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "reach,diff,flags,files,footer"
+  behavior: default
+  require_changes: no
```
mlodic, line 20: please set this to `true`, otherwise we risk to be flooded by coverage comments in some PR. [Ref](https://docs.codecov.io/docs/pull-request-comments#requiring-changes)

mlodic (changes requested):
thank you for your work and for this PR! The coverage works and this is good.

However there are some things I would like to review.

mlodic:
unluckily 66.8% is not good enough, we should get at least at 70%. We could focus on improving the score in another PR.

requirements.txt
Eshaan7: well, the `django-nose` package needs to be under `INSTALLED_APPS` in `settings,py`. @m0mosenpai but we don't need the `coverage` package in the production requirements file so it can be removed.

docker/default.yml
Eshaan7: agree with @mlodic

Eshaan7:
> 
> 
> unluckily 66.8% is not good enough, we should get at least at 70%. We could focus on improving the score in another PR.

Yes. We should atleast have full coverage for all API endpoints and analyzers. We can also write tests for the models/signals, but that is not a priority. An issue #229 is already open for this.

m0mosenpai:
Yeah I'll work on all the points in #229 in the priority order. Point 4 is next so this coverage percentage should hopefully improve and things should be better by the time I'm done with all the points.

mlodic:
Merged!

PR #317, opened 2021-01-12
Fixes #229: add tests for job, tag viewsets and other standalone API views | also improve some API view funcs

### LINKED ISSUE #229: More robust testing for the REST API/ more testcases

m0mosenpai:
# Description

Added CRUD tests for Tags and tests for list jobs/ get jobs by id for the API. 

## Related issues
#229 Point 4 and 6.

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

# Current thoughts/ issues

1. So far, I can't even create a tag. Am getting a `Bad Request` so I think I'm sending the data incorrectly in some way. 
2. What's the difference between creating a Tag object via `models.Tag` and sending an API request to create a tag. They might just be 2 completely different things I'm confusing.
3. As discussed on Slack, if the data doesn't persist between tests, I might have to create a tag each time before testing for listing them or deleting them etc. Because as of now, most of the tests give a `404` (creating a tag gives `400` and deleting is `Forbidden`).

Also, in regards to https://github.com/intelowlproject/IntelOwl/issues/229#issuecomment-756870257, I can work on addressing those concerns in this PR as well. Let me know your thoughts.


CITIZENDOT:
In `pyintelowl`, `durin` API_KEY needs to be sent as a part of header. I don't see it here. Does this `self.client` work in a different way?

tests/test_api.py
```diff
@@ -187,3 +188,48 @@ def test_ask_analysis_result(self):
...
+    def setUp(self):
+        self.client = APIClient()
+        self.user = User.objects.create_user(
+            username="test", email="test@intelowl.com", password="test"
+        )
+        self.client.force_authenticate(user=self.user)
+        self.tag = models.Tag.objects.create(label="Test", color="#FF5733")
+
+    def test_create_new_tag(self):
+        data = {"label": self.tag.label, "color": self.tag.color}
```
Eshaan7, line 211: the `label` and `color` here should be different. See my message on slack group.

Eshaan7:
```python3
self.assertEqual(models.Tag.objects.count(), 1)
data= {"label": "newTestLabel", "color": "#765A54"}
response = self.client.post("/api/tags", data)
self.assertEqual(response.status_code, 200)
self.assertEqual(models.Tag.objects.count(), 2)
```

m0mosenpai: Listing and creation tests for both jobs and tags work but the rest still give a `404`. Also, still not sure how to go about getting sufficient permission for `delete`.

Eshaan7: there is no PUT/DELETE for jobs, only for tags

m0mosenpai: Yes, sorry I meant listing works for jobs (GET) but get-by-id doesn't. For tags, create and list tags (GET) work but the rest don't, with delete giving a 403.

Eshaan7: This is because the test job you create in `setUp` is via admin user and the test user doesn't have permission to view/delete it, 

So we should do 2 test cases:
1. For the `job` created in `setUp`, test user doesn't have permission so the API gives 403.
2. In the `test_job_get_by_id` and `test_job_delete_by_id` functions, we can use the guardian's `assign_perm` method  [(ref.)](https://django-guardian.readthedocs.io/en/stable/api/guardian.shortcuts.html#assign-perm) to give user permission to the test `job` and then we should receive 200 status code.

tests/test_api.py
```diff
@@ -208,28 +209,30 @@ def setUp(self):
...
+        self.assertEqual(response.status_code, 201)
+        self.assertEqual(models.Tag.objects.count(), 2)
 
     def test_list_all_tags(self):
         response = self.client.get("/api/tags")
         self.assertEqual(response.status_code, 200)
 
     def test_get_tag_by_id(self):
-        tag_id = os.environ.get("TEST_TAG_ID", 1)
+        tag_id = os.environ.get("TEST_TAG_ID", "1")
```
Eshaan7, line 223: better to move this line to `setUp` function like `self.tag_id = os.environ.get("TEST_TAG_ID", "1")`

Eshaan7: so to remove duplication from L228 and L234 too.

tests/test_api.py
m0mosenpai: Yeah I was also thinking permissions were the issue since I tried sending a request manually via `curl` using `durin-auth- token` from my `django-admin-interface` and then it seemed to work.

So, I should split  `test_get<tag/job>_by_id`, `test_update_tag` and `test_delete_tag` tests each into 2 tests where in one i'm confirming that the user doesn't have enough permission to make those changes and in the other i'm giving the permission and then confirming that now they can? Did I understood that correctly?

Eshaan7: Yes. And you are right, get_by_id, put and delete.

tests/test_api.py
```diff
@@ -187,3 +189,81 @@ def test_ask_analysis_result(self):
         data = {"job_id": job_id}
         response = self.client.get("/api/ask_analysis_result", data)
         self.assertEqual(response.status_code, 200)
+
+    def test_list_all_jobs(self):
+        response = self.client.get("/api/jobs")
+        self.assertEqual(response.status_code, 200)
+
+    def test_get_job_by_id_no_admin(self):
+        job_id = os.environ.get("TEST_JOB_ID", "1")
```
Eshaan7, line 198: move this to `setUp`.

tests/test_api.py
```diff
@@ -187,3 +189,81 @@ def test_ask_analysis_result(self):
...
+        response = self.client.get(f"/api/jobs/{job_id}")
+        self.assertEqual(response.status_code, 404)
+
+    def test_get_job_by_id_admin(self):
+        assign_perm("api_app.view_job", self.user)
```
Eshaan7, line 203: i think this should be similar to Line 238.

tests/test_api.py
```diff
@@ -187,3 +189,81 @@ def test_ask_analysis_result(self):
...
+        self.assertEqual(self.user.has_perm("api_app.view_job"), True)
+        job_id = os.environ.get("TEST_JOB_ID", "1")
+        response = self.client.get(f"/api/jobs/{job_id}")
+        self.assertEqual(response.status_code, 200)
+
+
+class ApiTagTests(ApiTests):
```
Eshaan7, line 210: this should be 

<code>
class ApiTagTests(<b>TestCase</b>)
</code>

tests/test_api.py
```diff
@@ -18,7 +20,7 @@
 class ApiTests(TestCase):
```
Eshaan7, line 20: ideally we should rename this to `ApiJobTests` so the schema matches with the other `ApiTagTests`

tests/test_api.py
m0mosenpai: I'm confused about this. Previously also I thought about this, but we are not creating any job in the `setUp` of `ApiJobTests` class so why is the test `test_ask_analysis_result` successful when it sends a job id in the `data` field - asking the result for a job that has never been created? Assuming `setUp` runs everytime before each test and creates a new test database.

m0mosenpai:
There is also a `create_superuser()` available that we can maybe use instead of using `create_user()` in `setUp` which automatically grants all the permissions to the user. We won't have to use `assign_perm` explicitly in every test.

However in either case, I'm still getting a `404` in `GET` and `PUT` and `403` in `DELETE` despite the assertion of `has_perm()` being successful, before and after giving permission.

tests/test_api.py
Eshaan7: nice catch! turns out it's unintended behavour since we have hardcoded 200 status code in the `ask_analysis_resutlt` function. (https://github.com/intelowlproject/IntelOwl/blob/7463881edc8448fd6f9c78a8ca192412c1a49578/api_app/api.py#L346)

Eshaan7:
> There is also a create_superuser() available that we can maybe use instead of using create_user() in setUp which automatically grants all the permissions to the user.

That would seem more clean, so good idea. If it works as intended, we can use that. 

> However in either case, I'm still getting a 404 in GET and PUT and 403 in DELETE despite the assertion of has_perm() being successful, before and after giving permission.

I am not able to figure this out either. I'll have to test this on my own.

For now, can you replace the `assign_perm` with `create_superuser` method, update your PR and squash all commits into one ? I might need to push some changes too when I test it.

m0mosenpai:
Okay i'll do that. Shall I remove the `...no_admin` tests then (since `create_superuser()` should give the permissions to the user from the very beginning itself) ?

tests/test_api.py
m0mosenpai: Oh that makes sense. Looking at the `api.py` file, this seems to be a common theme followed in other endpoints as well so we could `assert` the response data in this test instead of the reponse code?

> i think this should be similar to Line 238.
Yes but I wasn't sure what kind of a test job I should create since the model has a lot of parameters - what kind of a sample job I should create, even thought it might make much of a difference for testing purposes.

Eshaan7:
> Okay i'll do that. Shall I remove the ...no_admin tests then (since create_superuser() should give the permissions to the user from the very beginning itself) ?

Do these 2 things,

- Create 2 users in the `setUp` function; one normal user with `create_user` and a superuser via `create_superuser`.
- add both `_admin` and `_no_admin` test cases. Use `self.client.force_authenticate` function passing the particular user in each testcase to deal with auth.

Then, I'll code further.

m0mosenpai:
Alright, I'll do this today itself. But doesn't the `force_authenticate` run before each test anyways as part of `setUp`. Are we expecting different behavior by having it run in each test case separately?

Eshaan7:
I updated the tests in a way that it doesn't do checking for the permissions for now. We are only creating a superuser and checking the cases of 200, 204, 404, 400 (where applicable) and not the 403 cases.

That is, now if the tests run successfully we can mark point 4 and 6 of the issue closed and concentrate on point 5 in a different PR.

Eshaan7:
>  Expected — Waiting for status to be reported 

it's been so many hours, something is broken on their side ?

PR #321, opened 2021-01-18
Changed requires_configuration to required_secrets

### LINKED ISSUE #306: better management of unconfigured analyzers
Regarding this closed issue (#298), we have a problem with those analyzers that require the use of an API key but either the user does not have the key or has not configured it.

Problems faced by users:

* they execute the "Run all available analyzers" option and get a lot of errors. This could be avoided just by executing the correctly configured analyzers.

* they see those analyzers as "runnable" but, actually, they are not and, when executed, they just instantly crash. I think we should not allow the user to execute them without the configuration set but, at the same time, let the user understand from the GUI that those analyzers exists and can be appropriately configured (the `requires_configuration` flag could be enough for this).

This seems easier than it is. ATM the API keys are checked in the code of each analyzer, while this should be done `before` the analysis actually runs. I am open to suggestions on how to solve this problem

CITIZENDOT:
# Description

It's not yet completed

**Changes Made**:  Changed `requires_configuration` to `required_secrets` 
**Changes willing to make**:  I have the following approach. Correct me, If I'm wrong somewhere.
* Add a new exception `UnconfiguredAnalyzer` in [exceptions.py](https://github.com/intelowlproject/IntelOwl/blob/9715a01c6032f78a3bef6c5d6e8ad922ab79491e/api_app/exceptions.py)
* In [general.py](https://github.com/intelowlproject/IntelOwl/blob/6ab6b1cc8503c1b341b6c35cd49b9201c0828bab/api_app/script_analyzers/general.py), Add a condition somewhere around [#L36](https://github.com/intelowlproject/IntelOwl/blob/6ab6b1cc8503c1b341b6c35cd49b9201c0828bab/api_app/script_analyzers/general.py#L36), whether secrets in `required_secrets` are available and proceed further only if they are available.

**Doubt**: How to ensure that when `UnconfiguredAnalyzer` exception is raised, It doesn't get counted as error in reports?

## Related issues
Fixes #306 

## Type of change

# Checklist

- [ ] I added new secrets in the files [env_file_app_template](https://github.com/intelowlproject/IntelOwl/blob/master/env_file_app_template), [env_file_app_ci](https://github.com/certego/IntelOwl/blob/master/env_file_app_travis) and in the docs: [Installation](./Installation.md)
- [ ] `Black` gave 0 errors.
- [ ] `Flake` gave 0 errors.

### DIFF
--- a/configuration/analyzer_config.json
+++ b/configuration/analyzer_config.json
@@ -16,7 +16,7 @@
     "observable_supported": ["ip"],
     "external_service": true,
     "description": "check if an ip was reported on AbuseIPDB",
-    "requires_configuration": true,
+    "required_secrets": ["ABUSEIPDB_KEY"],
     "python_module": "abuseipdb.AbuseIPDB",
     "soft_time_limit": 30
   },
@@ -27,7 +27,7 @@
     ],
     "external_service": true,
     "description": "scan an IP against the Auth0 API",
-    "requires_configuration": true,
+    "required_secrets": ["AUTH0_KEY"],
     "python_module": "auth0.Auth0",
     "soft_time_limit": 30
   },
@@ -54,7 +54,7 @@
     "observable_supported": ["domain","url"],
     "external_service": true,
     "description": "scan an observable against the CIRCL Passive DNS DB",
-    "requires_configuration": true,
+    "required_secrets": ["CIRCL_CREDENTIALS"],
     "python_module": "circl_pdns.CIRCL_PDNS",
     "soft_time_limit": 30
   },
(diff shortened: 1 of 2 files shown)


mlodic:
Even if I talked about `general.py` in the issue I think that the best position to filter those analyzer should be in the `filter_analyzers` function

configuration/analyzer_config.json
```diff
@@ -209,7 +213,10 @@
     "decription": "static generic document analysis (Oletools)",
     "python_module": "doc_info.DocInfo",
     "soft_time_limit": 30,
-    "queue": "local"
+    "queue": "local",
+    "additional_config_params": {
```
Eshaan7, line 217: I don't think this is required here. The `additional_passwords_to_check` is a feature of the `Doc_Info_Experimental` only.

PR #332, opened 2021-01-27
adding swagger documentation

spiderxm:
Added Swagger and Redoc documentation for rest apis.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 [**Upgrade Guide**](https://intelowl.readthedocs.io/en/latest/Installation.html#update-to-the-most-recent-version)
 
+## [v2.1.1](https://github.com/intelowlproject/IntelOwl/releases/tag/v2.1.1)
+**FIXES/IMPROVEMENTS/Dependency upgrades**
+- now `start.py` works with the most recent 1.28.2 version of docker-compose
+- updated Django, Yara and Speakeasy to most recent versions
+
+
 ## [v2.1.0](https://github.com/intelowlproject/IntelOwl/releases/tag/v2.1.0)
 **IMPORTANT FIX**
 We changed `docker-compose` file names for optional analyzers. In the `v.2.0.0` this broke Docker Hub builds, causing them to fail. Please upgrade to this version to be able to use the optional analyzers again.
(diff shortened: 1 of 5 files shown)


mlodic (changes requested):
Thank you for your PR!

However it is really important that you read and follow the contributions guidelines.

Also please do not delete the PR template. You should compile it. Otherwise your contributions won't be considered. Everything you need to know is written there.

Eshaan7:
We need to make this autogenerate for every commit with rest of the docs and host them publicly like: https://intelowl.readthedocs.io/api-docs. 

How can we achieve that?

spiderxm:
We can use the command
> python manage.py python manage.py generateschema 

This will generate open api definitions in yaml.
That can be used to make swagger documentation and serve them via different application.

Eshaan7:
> This will generate open api definitions in yaml.
> That can be used to make swagger documentation and serve them via different application.

Okay so what we need to do is create a GH action (or readthedocs webhook) which builds the `.yml` schema and puts it into `/docs/source/` which then gets rendered using [`sphinx openapi extension`](https://github.com/sphinx-contrib/openapi) ? The catch here is that this action should take place before readthedocs build gets triggered for which you should find something in readthedocs documentation.

api_app/urls.py
```diff
@@ -33,3 +56,4 @@
     # Viewsets
     path(r"", include(router.urls)),
 ]
+
```
Eshaan7, line 59: please fix codefactor issue

api_app/urls.py
```diff
@@ -13,14 +17,33 @@
...
 from .auth import LoginView, LogoutView
 
-
 # Routers provide an easy way of automatically determining the URL conf.
 router = routers.DefaultRouter(trailing_slash=False)
 router.register(r"tags", TagViewSet)
 router.register(r"jobs", JobViewSet)
 
+# Documentation Settings
+schema_view = get_schema_view(
```
Eshaan7, line 26: instead of adding this here, we should add a new file called `schemas.py` and add all this docs related code there.

spiderxm:
Will look into it

Eshaan7:
https://github.com/sphinx-contrib/redoc/issues/15

PR #335, opened 2021-02-13
fix #324: add DELETE /api/jobs/{id}

### LINKED ISSUE #324: [API] Delete job endpoint +  button
### Issue

At the moment there is not a way to delete old jobs manually.

The only way to remove jobs is:

1. wait for the cron that removes the old jobs. This cron is automatically launched once a day at night. It does remove jobs older than 3 days. It can be customized with the variable `OLD_JOBS_RETENTION_DAYS`.
2. Drop the database
3. The administrator can delete any job from the Django Admin Interface.

### Proposed Solution

**Web App:**

We could provide an API + a button in the GUI to allow the user to perform this operation directly. This should be added in:

1. The single job view
2. The dashboard

**CLI:**

Should also add a `IntelOwl.delete_job_by_id(job_id: int)` function in pyintelowl and create a corresponding command.

> Note: We should allow the users to only delete jobs created by themselves, while the administrator can already do it in the Django Admin Interface

Eshaan7:
## Related issues
Issue: #324

## Type of change

# Checklist

(The test for this will be added as part of #317)
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

### DIFF
--- a/api_app/signal_handlers.py
+++ b/api_app/signal_handlers.py
@@ -17,9 +17,10 @@ def save_profile(sender, instance, created, **kwargs):
     if created:
         grp, grp_created = Group.objects.get_or_create(name="DefaultGlobal")
         if grp_created:
-            # view/add permissions for Job model
+            # view/add/delete permissions for Job model
             assign_perm("api_app.view_job", grp)
             assign_perm("api_app.add_job", grp)
+            assign_perm("api_app.delete_job", grp)
             # view/add/change permissions for Tag model
             assign_perm("api_app.view_tag", grp)
             assign_perm("api_app.add_tag", grp)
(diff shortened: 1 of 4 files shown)


api_app/signal_handlers.py
```diff
@@ -17,9 +17,10 @@ def save_profile(sender, instance, created, **kwargs):
     if created:
         grp, grp_created = Group.objects.get_or_create(name="DefaultGlobal")
         if grp_created:
-            # view/add permissions for Job model
+            # view/add/delete permissions for Job model
             assign_perm("api_app.view_job", grp)
             assign_perm("api_app.add_job", grp)
+            assign_perm("api_app.delete_job", grp)
```
Eshaan7, line 23: TODO: should add a migration for this

Eshaan7: Done! `api_app/migrations/0005_delete_job_perm.py`

Eshaan7:
Note: adding permissions to group during migration is not working and we can't replace `if created` in the `signal_handlers.py`. Probably would have to provide a helper script to the users with the new release.

PR #337, opened 2021-02-14
Add new api to kill ongoing job

### LINKED ISSUE #225: [API] endpoint to stop/kill a running job
In this case it would be nice to have a button - kill job or stop job. 

_Originally posted by @WhiteHatMR in https://github.com/intelowlproject/IntelOwl/issue_comments/707807962_

We could add a `kill` button to the job (analysis) result page. This button when invoked should stop all ongoing celery tasks (i.e. all running analyzers) for the particular job. The status of such job should then be marked `failed` or `killed`. 

This sounds easy on paper, but would require much changes:

- To stop an ongoing celery task, we can send a control signal (`TERM`) ([ref.](https://stackoverflow.com/a/25561389/10534470))  but it requires the ID for the particular task. So need a way to keep track of all celery task IDs for a particular job.
- An endpoint `/api/kill_analysis/{job_id}` with appropriate permission enforcing i.e. If X created the job then only the admin and X should be allowed to kill it.

**Checklist:**
- [x] API endpoint
- [ ] Button in frontend (IntelOwl-ng)
- [ ] member function and CLI command in pyintelowl.

sp35:
Signed-off-by: Shubham Pandey <shubhampcpandey13@gmail.com>
Fixes #225 by @Eshaan7 

# Description
This PR adds a new API `kill_running_job` which would receive job_id as a parameter and revoke the running celery tasks associated with it. Once this is done, the job's status is set `killed`.

Note: The task ids are stored in cache once the task is set, in `start_analyzers` function.

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.
  
# Real World Example
image
Here the job was killed before all the tasks could run.

### DIFF
--- a/api_app/api.py
+++ b/api_app/api.py
@@ -3,11 +3,13 @@ import logging
 from api_app import models, serializers, helpers
 from api_app.permissions import ExtendedObjectPermissions
 from .script_analyzers import general
+from intel_owl.celery import app as celery_app
 
 from wsgiref.util import FileWrapper
 
 from django.http import HttpResponse
 from django.db.models import Q
+from django.core.cache import cache
 from rest_framework.response import Response
 from rest_framework import status, viewsets, mixins
 from rest_framework.decorators import api_view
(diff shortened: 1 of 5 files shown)


api_app/api.py
```diff
@@ -421,6 +423,58 @@ def download_sample(request):
...
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        # close celery tasks
+        task_ids = cache.get(data_received["job_id"])
+        if isinstance(task_ids, list):
+            celery_app.control.revoke(task_ids)
+            cache.delete((data_received["job_id"]))
+        # set job status
+        job.status = "killed"
+        job.save()
```
Eshaan7, line 467: I would prefer if this is made as an utility function and kept in general.py as `kill_running_analysis`.

api_app/api.py
```diff
@@ -421,6 +423,58 @@ def download_sample(request):
         )
 
 
+@api_view(["PATCH"])
+def kill_running_job(request):
```
Eshaan7, line 427: Rename to kill_analysis

api_app/urls.py
```diff
@@ -30,6 +31,7 @@
     path("ask_analysis_result", ask_analysis_result),
     path("get_analyzer_configs", get_analyzer_configs),
     path("download_sample", download_sample),
+    path("job/kill/", kill_running_job),
```
Eshaan7, line 34: Change the URL format to `/api/kill_analysis?job_id`

Eshaan7 (changes requested):

api_app/api.py
```diff
@@ -421,6 +423,58 @@ def download_sample(request):
...
+            return Response({"error": "821"}, status=status.HTTP_400_BAD_REQUEST)
+        # get job object
+        try:
+            job = models.Job.objects.get(id=data_received["job_id"])
+        except models.Job.DoesNotExist:
+            return Response(
+                {"detail": "job not found"}, status=status.HTTP_404_NOT_FOUND
+            )
+        # check permission
+        if not request.user.has_perm("api_app.change_job", job):
```
Eshaan7, line 449: Regarding this permission, would also need to add an entry for `assign_perm` in the post_migrate signal handler under signal_handler.py

Eshaan7 (changes requested):

api_app/api.py
sp35: I'm waiting for the pr on post migrate signal. I'll change this once it's done. For now, I added the entry in the save one.

Eshaan7: Imma merge your PR first and make changes accordingly since the tests are failing for the other PR.

api_app/script_analyzers/general.py
```diff
@@ -104,14 +108,26 @@ def start_analyzers(
                 ]
             # run analyzer with a celery task asynchronously
             stl = ac.get("soft_time_limit", 300)
-            send_task(
-                "run_analyzer",
-                args=args,
-                queue=queue,
-                soft_time_limit=stl,
+            t_id = uuid()
+            tasks.run_analyzer.apply_async(
```
Eshaan7, line 112: Is there a particular reason why you changed this? The reason for using `send_task` was to avoid cyclic import of `tasks` module. Is it not possible to send `task_id` when using `send_task`?

sp35: `apply_async` provides an argument to set task id, all examples I saw were using it. 
But it actually calls `send_task` under the hood. So yes, it should be possible to continue using it.
https://github.com/celery/celery/blob/a5357cab8aa80ff701a1970f55dc1e1083a161f5/celery/app/task.py#L561-L566

Also, I'll change it to import from the `app` instead of `execute` module, it works on my machine.
Ref: https://docs.celeryproject.org/en/v4.4.7/internals/deprecation.html#modules-to-remove

api_app/api.py
sp35: This can be resolved.

sp35: This can be resolved too.

api_app/urls.py
sp35: This can be resolved as well.

Eshaan7 (approved):

mlodic:
hi @sp35! Thank you for your contribution!

PR #339, opened 2021-02-16
chore: change url format for kill job to jobs/{job_id}/kill

sp35:
# Description
This modifies the `kill api view` to use `@action` decorator under `JobViewSet`.
The URL would change to ` /api/jobs/{job_id}/kill`  from `/api/kill_running_analysis?job_id`.
Asked by @Eshaan7 

## Related issues
#225

# Checklist

- [x] The pull request is for the branch develop.
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

### DIFF
--- a/api_app/api.py
+++ b/api_app/api.py
@@ -10,7 +10,7 @@ from django.http import HttpResponse
 from django.db.models import Q
 from rest_framework.response import Response
 from rest_framework import status, viewsets, mixins
-from rest_framework.decorators import api_view
+from rest_framework.decorators import api_view, action
 from rest_framework.permissions import DjangoObjectPermissions
 from guardian.decorators import permission_required_or_403
 from rest_framework_guardian.filters import ObjectPermissionsFilter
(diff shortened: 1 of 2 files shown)


api_app/api.py
```diff
@@ -512,6 +463,54 @@ def get_serializer_class(self, *args, **kwargs):
         except (KeyError, AttributeError):
             return super(JobViewSet, self).get_serializer_class()
 
+    @action(detail=True, methods=["patch"])
```
Eshaan7, line 466: Shouldn't it be `detail=False`? We do not want this action to be available to collections /lists.

Eshaan7 (changes requested):

api_app/api.py
sp35: It should be `True.

Eshaan7 (approved):

Eshaan7 (approved):

PR #343, opened 2021-02-24
Added Script to change analyzer_config.json to newer format

### LINKED ISSUE #306: better management of unconfigured analyzers

CITIZENDOT:
# Description
Starting work for Refactoring analyzer_config.json. Might fix #306 and good maintainability for future purposes.

## Related issues
Please add related issues.

## Type of change

# Checklist

- [ ] The pull request is for the branch develop
- [ ] I added new secrets in the files [env_file_app_template](https://github.com/intelowlproject/IntelOwl/blob/master/env_file_app_template), [env_file_app_ci](https://github.com/certego/IntelOwl/blob/master/env_file_app_travis) and in the docs: [Installation](./Installation.md)
- [X] `Black` gave 0 errors.
- [ ] `Flake` gave 0 errors.
  
### please follow these rules

# Real World Example

### DIFF
--- a/format_analyzer_config.py
+++ b/format_analyzer_config.py
@@ -1,0 +1,85 @@
+import json
+
+
+def main():
+    # file_path = input("File Path: ")
+    file_path = "configuration/analyzer_config.json"
+    with open(file_path, "r") as f:
+        config = json.load(f)
+    new_config = {}
+    for analyzer in config:
+        new_config[analyzer] = {}
+        new_config[analyzer]["type"] = config[analyzer]["type"]
+        new_config[analyzer]["python_module"] = config[analyzer]["python_module"]
+
+        if "disabled" in config[analyzer]:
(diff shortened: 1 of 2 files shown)


Eshaan7:
Thankyou!

PR #346, opened 2021-03-04
[add] MISP init connecton SSL check

raw-data:
# Description

Introduced the possibility to enable/disable SSL check while connecting to a MISP platform. 
e.g. useful when testing IntelOwl with dummy MISP instance with self signed certificate.

## Type of change
- [x] New analyzer configuration option


mlodic:
hey @raw-data thank you for your interest in the project!

Even if your code was working, I have just changed it to be more accurate for the actual IntelOwl structure.

In particular, there is no need to use an environment variable to change that behavior. In this way, you should restart IntelOwl each time you want to change that specific configuration. On the contrary, you can extract the configuration directly from the `analyzer_config.json` file at runtime. This means that you can change the `ssl_check` value there when you want, without restarting the entire application.

So, I guess that  in your case you have a local MISP instance without SSL enabled. You can just change the `ssl_check` value to `true` directly in the file **when you want** and change it back when you will enable SSL.

I have also added this in the documentation.

PR #349, opened 2021-03-10
Updated Installation.md

m0mosenpai:
# Description

Should be `intel_owl_uwsgi` instead of `intelowl_uwsgi` according to the name of the containers that start up.

## Related issues
NA

## Type of change

- [x] Updating documentation

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.
 
# Real World Example

`intel_owl_uwsgi` container is created when starting the application.
image

Following the steps of the current documentation, a superuser is created with the command:
```
docker exec -ti intelowl_uwsgi python3 manage.py createsuperuser
```
which yields the error
image

Performing the suggested change fixes it
image

### DIFF
--- a/docs/source/Installation.md
+++ b/docs/source/Installation.md
@@ -21,7 +21,7 @@ cd ..
 python start.py prod up
 
 # create a super user 
-docker exec -ti intelowl_uwsgi python3 manage.py createsuperuser
+docker exec -ti intel_owl_uwsgi python3 manage.py createsuperuser
 
 # now the app is running on http://localhost:80
 ```


mlodic:
You should rebase your branch to the latest `develop` version. That parameter was changed.

m0mosenpai:
> You should rebase your branch to the latest `develop` version. That parameter was changed.

Oh I see. I see that it's still the same here https://github.com/intelowlproject/IntelOwl/blob/develop/docs/source/Installation.md
Sorry am I missing something? I tried rebasing as well as referring to the documentation website but found the same result.

mlodic:
I was talking about the docker-compose file, not the documentation that is right

m0mosenpai:
Got it, thanks. My bad I missed that.

PR #368, opened 2021-03-16
Add FireHol IPList Observable Analyzer

### LINKED ISSUE #53: FireHOL blocklists analyzer
http://iplists.firehol.org/

We could integrate an analyzer for the FireHOL blocklists. In this way, we could check if a specific IP was reported as malicious by one of those lists.

We should manage the download of these lists internally in the project. Then we should update these lists regularly to allow the IntelOwl users to get fresh results.

subhashissuara:
# Description

This PR adds an observable analyzer `FireHol_IPList`.

## Related issues
Closes #53 by @mlodic

## Type of change

# Checklist

- [x] If I added a new analyzer, I updated the file [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md). If the analyzer provides additional optional configuration, I added the available options here: [Advanced-Usage](./Advanced-Usage.md)
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

# Real World Example

Users add list names (to search from) in the custom config as shown below:
image

The analyzer then gives the following output:
image

Database is updated every day. Here is the log of 2 consecutive scans on the same list.
image

### DIFF
--- a/api_app/script_analyzers/observable_analyzers/firehol_iplist.py
+++ b/api_app/script_analyzers/observable_analyzers/firehol_iplist.py
@@ -1,0 +1,98 @@
+import os
+import logging
+import traceback
+import requests
+import ipaddress
+
+from api_app.exceptions import AnalyzerConfigurationException, AnalyzerRunException
+from api_app.script_analyzers import classes
+from intel_owl import settings
+from datetime import datetime
+
+logger = logging.getLogger(__name__)
+
+db_path = f"{settings.MEDIA_ROOT}"
+
+
+class FireHol_IPList(classes.ObservableAnalyzer):
+    def set_config(self, additional_config_params):
+        self.list_names = additional_config_params.get(
+            "list_names", ["firehol_level1.netset"]
+        )
+
+    def run(self):
+        ip = self.observable_name
+        result = {}
+
+        if not self.list_names:
+            raise AnalyzerConfigurationException(
+                "list_names is empty in custom analyzer config, add an iplist"
+            )
+
+        for list_name in self.list_names:
+            result[list_name] = False
+
+            self.check_iplist_status(list_name)
+
+            with open(f"{db_path}/{list_name}", "r") as f:
+                db = f.read()
+
+            db_list = db.split("\n")
+
+            for ip_or_subnet in db_list:
+                if ip_or_subnet != "":
+                    if ipaddress.ip_address(ip) in ipaddress.ip_network(ip_or_subnet):
+                        result[list_name] = True
+                        break
+
+        return result
+
+    def download_iplist(self, list_name):
+        if ".ipset" not in list_name and ".netset" not in list_name:
+            raise AnalyzerConfigurationException(
+                f"extension missing from {list_name} (add .ipset or .netset to name)"
+            )
+
+        try:
+            iplist_location = f"{db_path}/{list_name}"
+            data_cleaned = ""
+
+            logger.info(f"starting download of {list_name} from firehol iplist")
+            url = f"https://iplists.firehol.org/files/{list_name}"
+            r = requests.get(url)
+            r.raise_for_status()
+
+            data_extracted = r.content.decode()
+
+            for line in data_extracted.splitlines():
+                if not line.startswith("#"):
+                    data_cleaned += f"{line}\n"
+
+            with open(iplist_location, "w") as f:
+                f.write(data_cleaned)
+
+            if not os.path.exists(iplist_location):
+                raise AnalyzerRunException(f"failed extraction of {list_name} iplist")
+
+            logger.info(f"ended download of {list_name} from firehol iplist")
+
+        except Exception as e:
+            traceback.print_exc()
+            logger.exception(e)
+
+    def check_iplist_status(self, list_name):
+        iplist_location = f"{db_path}/{list_name}"
+
+        if not os.path.exists(iplist_location):
+            self.download_iplist(list_name)
+
+        now = datetime.now()
+        timestamp = os.path.getctime(iplist_location)
+        dt_object = datetime.fromtimestamp(timestamp)
+        time_diff = now - dt_object
+
+        if time_diff.days < 1:
+            logger.info("iplist is up to date")
+        else:
+            os.remove(iplist_location)
+            self.download_iplist(list_name)
--- a/configuration/analyzer_config.json
+++ b/configuration/analyzer_config.json
@@ -255,6 +255,17 @@
     "soft_time_limit": 30,
     "queue": "local"
   },
+  "FireHol_IPList": {
+    "type": "observable",
+    "observable_supported": ["ip"],
+    "external_service": true,
+    "description": "check if an IP is in FireHol's IPList",
+    "python_module": "firehol_iplist.FireHol_IPList",
+    "additional_config_params": {
+      "list_names": ["firehol_level1.netset"]
+    },
+    "soft_time_limit": 180
+  },
   "Floss": {
     "type": "file",
     "supported_filetypes": ["application/x-dosexec"],
--- a/docs/source/Usage.md
+++ b/docs/source/Usage.md
@@ -147,6 +147,7 @@ The following is the list of the available analyzers you can run out-of-the-box:
 * `XForceExchange`: scan an observable on [IBM X-Force Exchange](https://exchange.xforce.ibmcloud.com/)
 * `Renderton`: get screenshot of a web page using rendertron (puppeteer) [renderton repo](https://github.com/GoogleChrome/rendertron)
 *  `SSAPINet`: get a screenshot of a web page using [screenshotapi.net](https://screenshotapi.net/) (external source); additional config options can be added to `extra_api_params` [in the config](https://github.com/intelowlproject/IntelOwl/blob/master/configuration/analyzer_config.json).
+* `FireHol_IPList`: check if an IP is in [FireHol's IPList](https://iplists.firehol.org/)
 
 #### Generic analyzers (email, phone number, ...)
 Some Analyzers require details other than just IP, URL, Domain etc... We classified them as Generic Analyzers. Since the type of field is not known, there is a format for strings to be followed.
--- a/tests/test_observables.py
+++ b/tests/test_observables.py
@@ -44,6 +44,7 @@ from api_app.script_analyzers.observable_analyzers import (
     crxcavator,
     rendertron,
     ss_api_net,
+    firehol_iplist,
 )
 from api_app.models import Job
 from .mock_utils import (
@@ -95,6 +96,17 @@ def mocked_triage_get(*args, **kwargs):
 
 def mocked_triage_post(*args, **kwargs):
     return MockResponse({"id": "sample_id", "status": "pending"}, 200)
+
+
+def mocked_firehol_iplist(*args, **kwargs):
+    return MockResponse(
+        json_data={},
+        status_code=200,
+        text="""0.0.0.0/8\n
+            1.10.16.0/20\n
+            1.19.0.0/16\n
+            3.90.198.217\n""",
+    )
 
 
 @mock_connections(patch("requests.get", side_effect=mocked_requests))
@@ -265,6 +277,17 @@ class IPAnalyzersTests(
             self.observable_name,
             self.observable_classification,
             {},
+        ).start()
+        self.assertEqual(report.get("success", False), True)
+
+    @mock_connections(patch("requests.get", side_effect=mocked_firehol_iplist))
+    def test_firehol_iplist(self, *args):
+        report = firehol_iplist.FireHol_IPList(
+            "FireHol_IPList",
+            self.job_id,
+            self.observable_name,
+            self.observable_classification,
+            {"list_names": ["firehol_level1.netset"]},
         ).start()
         self.assertEqual(report.get("success", False), True)
 
--- a/docs/source/Advanced-Usage.md
+++ b/docs/source/Advanced-Usage.md
@@ -159,7 +159,8 @@ List of some of the analyzers with optional configuration:
      * `destroy_screenshot` (default `false`) - if `true` the screenshot is not stored on their servers. Please make sure to use `output` parameter with value `image`, so you don't lose the screenshot, as the image link provided in the `json` result would work only once.
      
      Refer to the [docs](https://screenshotapi.net/documentation) for a reference to what other parameters are and their default values.
- 
+* `FireHol_IPList`:
+  * `list_names`: Add [FireHol's IPList](https://iplists.firehol.org/) names as comma separated values in `list_names` array.
 
 There are two ways to do this:
 


docs/source/Usage.md
```diff
@@ -145,6 +145,7 @@ The following is the list of the available analyzers you can run out-of-the-box:
 * `InQuest_REPdb`: Search in [InQuest Lab's](https://labs.inquest.net/repdb) Reputation Database
 * `InQuest_DFI`: Deep File Inspection by [InQuest Labs](https://labs.inquest.net/dfi)
 * `XForceExchange`: scan an observable on [IBM X-Force Exchange](https://exchange.xforce.ibmcloud.com/)
+* `FireHol_IPList`: check if an IP is in [FireHol's IPList](https://iplists.firehol.org/)
```
mlodic, line 148: not correct, you just integrated their highest priority list.

intel_owl/celery.py
```diff
@@ -69,4 +69,10 @@
         "schedule": crontab(minute=0, hour=0),
         "options": {"queue": "default"},
     },
+    # execute 1 time a day
```
mlodic, line 72: We should not leverage celery crons to do this. A lot of IntelOwl users could never use FireHol lists. This could be an incredible amount of useless computation and API abuse.

Please check how [stratosphere analyzer](https://github.com/intelowlproject/IntelOwl/blob/master/api_app/script_analyzers/observable_analyzers/stratosphere.py) was implemented and change the analyzer accordingly

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
```diff
@@ -0,0 +1,81 @@
...
+        result = {"found": False}
+
+        if not os.path.isfile(database_location):
+            self.updater()
+
+        with open(database_location, "r") as f:
+            db = f.read()
+
+        if self.observable_name in db:
+            result["found"] = True
```
mlodic, line 29: we cannot use a boolean for this. This is not maintanable considering that your PR does not solve the issue entirely.

This PR can be accepted only in the case that you already prepare this analyzer to have the same output schema that it would have once the issue is solved entirely. This means that you should provide the name of the lists + the category of the lists where the search IP address matched. In this case it would at maximum show the one that you integrated

mlodic (changes requested):
Thank you for the PR!

docs/source/Usage.md
Eshaan7: I believe we can write something like: 
> check if an IP is in `firehol_level1` list by [FireHol's IP List](https://iplists.firehol.org/).

intel_owl/celery.py
Eshaan7: Well actually it was my idea to use celery cron but yes I did not think about the drawback you mention. 

@subhashissuara We can dynamically download the IP list only when the user requests the analysis. So your `updater` method would just be a `download_ipset` method which would get called at the beginning of `run`.

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
Eshaan7: For simplicity, we can make it: `result[list_name] = True/False`.

Add `list_name` parameter to the `additional_config_parameters` inside the analyzer config.

Eshaan7:
@mlodic what do u think about adding `list_names` parameter to the `additional_config_parameters` inside the analyzer config ?

Whatever list names user provides would get downloaded dynamically when the analyzer is run and the IP would be checked against them one by one making final result in the format `result[list_name] = True/False`.

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
```diff
@@ -0,0 +1,81 @@
...
+            )
+
+            with open(database_location, "w") as f:
+                for ip in ipaddresses:
+                    if ip:
+                        f.write(f"{ip}")
+
+                for subnet in subnets:
+                    if subnet:
+                        f.write(f"{subnet}\n")
```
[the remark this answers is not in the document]
subhashissuara: Yes, there was an extra `\n` from IP regex, so to avoid blank lines I used it only for subnets. I will edit accordingly. Thank you!

docs/source/Usage.md
subhashissuara: Gotcha!

subhashissuara:
> @mlodic what do u think about adding `list_names` parameter to the `additional_config_parameters` inside the analyzer config ?
> 
> Whatever list names user provides would get downloaded dynamically when the analyzer is run and the IP would be checked against them one by one making final result in the format `result[list_name] = True/False`.
> 
> I believe this implementation would then solve the issue completely and not partially.
> 
> EDIT: The download links for the lists are in the same format like `https://iplists.firehol.org/files/{list_name}.ipset`. (For example, `https://iplists.firehol.org/files/cleantalk_1d.ipset`)

Thank you for the suggestion! Regarding the extension of `list_name`, we have both `.ipset` and `.netset` so, do we include the extension as a part of the `list_name`?

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
subhashissuara: Yes, that makes sense!

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
```diff
@@ -0,0 +1,70 @@
...
+            logger.info(f"starting download of {list_name} from firehol iplist")
+            url = f"https://iplists.firehol.org/files/{list_name}"
+            r = requests.get(url)
+            r.raise_for_status()
+
+            data_extracted = r.content.decode()
+
+            for line in data_extracted.splitlines():
+                if not line.startswith("#"):
+                    data_cleaned += f"{line}\n"
```
mlodic, line 62: what's the point to use this as a string instead of a list?

configuration/analyzer_config.json
```diff
@@ -255,6 +255,17 @@
...
     "queue": "local"
   },
+  "FireHol_IPList": {
+    "type": "observable",
+    "observable_supported": ["ip"],
+    "external_service": true,
+    "description": "check if an IP is in FireHol's IPList",
+    "python_module": "firehol_iplist.FireHol_IPList",
+    "additional_config_params": {
+      "list_names": ["firehol_level1.netset", "cleantalk_1d.ipset"]
```
mlodic, line 265: was cleantalk added randomly or there is a specific reason? I guess you used it for development purpose but why keep this here. Maybe we should think about which kind of lists we should provide by default.

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
```diff
@@ -0,0 +1,70 @@
...
+class FireHol_IPList(classes.ObservableAnalyzer):
+    def set_config(self, additional_config_params):
+        self.list_names = additional_config_params.get("list_names", [])
+
+    def run(self):
+        result = {}
+
+        for list_name in self.list_names:
+            result[list_name] = False
+            db = self.download_iplist(list_name)
```
mlodic, line 22: I guess you did not check the stratosphere analyzer as I mentioned....we cannot scrape the site each time an analysis is done...

mlodic (changes requested):
thank you for the updates but there is still some work to do

docs/source/Usage.md
subhashissuara: I did not change the sentence since we are fulfilling it with the new code changes. I added about the `list_names` parameter in Advanced Usage docs. Please suggest if the sentence needs to be changed.

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
subhashissuara: For the regex search we are doing in `run()`. Am I missing something?

configuration/analyzer_config.json
subhashissuara: Thank you for pointing out! I was just writing a comment asking about default set of list names. I should have done it beforehand, my bad. So, what kind of lists should we provide by default?

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
mlodic: the regex is not required, you can just put each element in a list then use ipaddress library to check what they are

subhashissuara: My point of view was that, if we don't find an IP address directly in the list then we filter out the subnets by matching the first block of given IP and search for the IP under the filtered subnets only. I thought this could reduce the searching we do. So, will there be any performance difference between both the ways? Thank you!

mlodic: which list are you talking about? `if self.observable_name in db:` this is not a check in a list but in a string

You used `ipaddress` library. Please make the effort to understand how to match the IP address by using only this library. There is no need to create huge strings neither to make any kind of regexes.

The only fact that you check whether a string is in a huge strings, instead of iterating in a list and performing a check for each string you saved in a list, is bad for performance. Plus regexes are expensive too. Also we can't  use a custom regex for things (like IP addresses) that follow a very well known schema: we need to leverage the standard library

configuration/analyzer_config.json
Eshaan7: By default, only Firehol_level1 should be enough. It's composed of many different iplists.

mlodic:
if someone would like to pick up this and finish it, it would be appreciated. Most of the work is already done

subhashissuara:
@mlodic I apologize for my inactivity, I have been busy due to university. I am looking forward to fixing the PR within tomorrow.

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
subhashissuara: My bad, I will change accordingly.

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
subhashissuara: Thank you for explaining it to me. I will change the data into a list and leverage the `ipaddress` library for checking.

`if we don't find an IP address directly in the list`

The list I referred here was fireHol's iplist.

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
```diff
@@ -0,0 +1,91 @@
...
+            self.check_iplist_status(list_name)
+
+            with open(f"{db_path}/{list_name}", "r") as f:
+                db = f.read()
+
+            db_list = db.split("\n")
+
+            for ip_or_subnet in db_list:
+                if ip_or_subnet != "":
+                    if ipaddress.ip_address(ip) in ipaddress.ip_network(ip_or_subnet):
```
sp35, line 37: Since `ipaddress.ip_address(ip)` would be a constant value, it makes sense to initialize only once, outside the loop.

sp35 (changes requested):

api_app/script_analyzers/observable_analyzers/firehol_iplist.py
subhashissuara: Right! I overlooked it. Changing it. Thanks!

tests/test_observables.py
```diff
@@ -268,6 +269,16 @@ def test_tor(self, mock_get=None, mock_post=None):
         ).start()
         self.assertEqual(report.get("success", False), True)
 
+    def test_firehol_iplist(self, mock_get=None, mock_post=None):
+        report = firehol_iplist.FireHol_IPList(
+            "FireHol_IPList",
+            self.job_id,
+            self.observable_name,
+            self.observable_classification,
+            {},
```
mlodic, line 278: there is a hidden problem here. Considering that `list_names` is initialized as empty, this would mean that most of the code is not reached by tests here. Could you add the default list `firehol_level1.netset` here or, maybe better, directly in the `set_config` method? https://github.com/intelowlproject/IntelOwl/pull/368/files#diff-3b52a5391525c883aaa2861e5aef8d6fea9e50cb88f8555e80a137ce40437938R19

mlodic (changes requested):
Thank you for coming back as soon as I have requested to finish this! I think that we are about to end this PR 👍🏻

tests/test_observables.py
subhashissuara: That's a beautiful catch! Adding it asap.

mlodic: If the key `list_names` exists in the `analyzer_config.py` and it is empty, it would override the default configuration.

Yes, your point is good, we could add an `AnalyzerConfigurationException` in the case the `list_names` is empty to manage also that case. Could you do it? Then we can merge

subhashissuara: Okay! I added `AnalyzerConfigurationException` for the empty list case. While testing I noticed a `522 Server Error` and the final error that showed up in the application was:
image
Since, it didn't download the db, how should we handle this error?

mlodic: the better things should be to mock the request to FireHol so we do not really contact the server in the tests. You can check how `test_triage_search` is implemented and try to add something similar.

subhashissuara: I couldn't explain it properly, my bad. I ran a scan in the web application to check if the analyzer was working as expected and got that error, so, I was asking if we can/should handle this? This is the whole traceback:

```
intelowl_celery_worker_default | [2021-04-06 10:30:04,459: INFO/MainProcess] starting download of firehol_level1.netset from firehol iplist
intelowl_celery_worker_default | Traceback (most recent call last):
intelowl_celery_worker_default |   File "/opt/deploy/intel_owl/api_app/script_analyzers/observable_analyzers/firehol_iplist.py", line 60, in download_iplist
intelowl_celery_worker_default |     r.raise_for_status()
intelowl_celery_worker_default |   File "/usr/local/lib/python3.8/site-packages/requests/models.py", line 943, in raise_for_status
intelowl_celery_worker_default |     raise HTTPError(http_error_msg, response=self)
intelowl_celery_worker_default | requests.exceptions.HTTPError: 522 Server Error:  for url: https://iplists.firehol.org/files/firehol_level1.netset
intelowl_celery_worker_default | [2021-04-06 10:30:37,763: ERROR/MainProcess] 522 Server Error:  for url: https://iplists.firehol.org/files/firehol_level1.netset
intelowl_celery_worker_default | Traceback (most recent call last):
intelowl_celery_worker_default |   File "/opt/deploy/intel_owl/api_app/script_analyzers/observable_analyzers/firehol_iplist.py", line 60, in download_iplist
intelowl_celery_worker_default |     r.raise_for_status()
intelowl_celery_worker_default |   File "/usr/local/lib/python3.8/site-packages/requests/models.py", line 943, in raise_for_status
intelowl_celery_worker_default |     raise HTTPError(http_error_msg, response=self)
intelowl_celery_worker_default | requests.exceptions.HTTPError: 522 Server Error:  for url: https://iplists.firehol.org/files/firehol_level1.netset
intelowl_celery_worker_default | Traceback (most recent call last):
intelowl_celery_worker_default |   File "/opt/deploy/intel_owl/api_app/script_analyzers/classes.py", line 99, in start
intelowl_celery_worker_default |     result = self.run()
intelowl_celery_worker_default |   File "/opt/deploy/intel_owl/api_app/script_analyzers/observable_analyzers/firehol_iplist.py", line 32, in run
intelowl_celery_worker_default |     self.check_iplist_status(list_name)
intelowl_celery_worker_default |   File "/opt/deploy/intel_owl/api_app/script_analyzers/observable_analyzers/firehol_iplist.py", line 87, in check_iplist_status
intelowl_celery_worker_default |     timestamp = os.path.getctime(iplist_location)
intelowl_celery_worker_default |   File "/usr/local/lib/python3.8/genericpath.py", line 65, in getctime
intelowl_celery_worker_default |     return os.stat(filename).st_ctime
intelowl_celery_worker_default | FileNotFoundError: [Errno 2] No such file or directory: '/opt/deploy/files_required/firehol_level1.netset'
intelowl_celery_worker_default | [2021-04-06 10:30:37,764: ERROR/MainProcess] job_id:3, analyzer:'FireHol_IPList'. Unexpected error: '[Errno 2] No such file or directory: '/opt/deploy/files_required/firehol_level1.netset''
intelowl_celery_worker_default | Traceback (most recent call last):
intelowl_celery_worker_default |   File "/opt/deploy/intel_owl/api_app/script_analyzers/classes.py", line 99, in start
intelowl_celery_worker_default |     result = self.run()
intelowl_celery_worker_default |   File "/opt/deploy/intel_owl/api_app/script_analyzers/observable_analyzers/firehol_iplist.py", line 32, in run
intelowl_celery_worker_default |     self.check_iplist_status(list_name)
intelowl_celery_worker_default |   File "/opt/deploy/intel_owl/api_app/script_analyzers/observable_analyzers/firehol_iplist.py", line 87, in check_iplist_status
intelowl_celery_worker_default |     timestamp = os.path.getctime(iplist_location)
intelowl_celery_worker_default |   File "/usr/local/lib/python3.8/genericpath.py", line 65, in getctime
intelowl_celery_worker_default |     return os.stat(filename).st_ctime
intelowl_celery_worker_default | FileNotFoundError: [Errno 2] No such file or directory: '/opt/deploy/files_required/firehol_level1.netset'
```

Yes, you are right we should mock the request for testing, I will try to implement it!

subhashissuara:
@mlodic For tests, Talos is requesting from server which had an internal error and caused the tests to fail ([reference](https://github.com/intelowlproject/IntelOwl/pull/368/checks?check_run_id=2278465096))

mlodic (approved):

mlodic:
Great! Merged!

subhashissuara:
Thank you @mlodic, @Eshaan7 and @sp35 for helping me through this PR. It was a great experience and was fun learning more about the project!

PR #370, opened 2021-03-19
Fixes #9: Add rendertron analyzer [docker based]

### LINKED ISSUE #9: [analyzer] screenshot to detect phishing campaigns
We could add a Manet analyzer to get a screenshot from a specified URL that we would like to analyze.

https://github.com/vbauer/manet

We should add the docker image as optional in the docker-compose configuration and use the Manet APIs to run the analysis.

sp35:
Signed-off-by: Shubham Pandey <shubhampcpandey13@gmail.com>

# Description
This adds a new docker analyzer that uses `rendertron` to capture a screenshot of a web page.
https://github.com/intelowlproject/IntelOwl/issues/9#issuecomment-801867465

~~work in progress, initials reviews appreciated!~~
1. ~~Issue with `renderton` using `--no-sandbox` flag: I did set up a sandbox either by using the chrome itself or using the sandbox with chromium that `puppeteer` installs, right now I'm looking for a solution to - `Failed to move to new namespace: PID namespaces supported, Network namespace supported, but failed: errno = Operation not permitted`  [Relevant Documentation](https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#setting-up-chrome-linux-sandbox)~~

2. Refactor code and dockerfile

## Related issues
Related to #9 by @mlodic 

## Type of change

# Checklist

- [x] If I added a new analyzer, I updated the file [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md). If the analyzer provides additional optional configuration, I added the available options here: [Advanced-Usage](./Advanced-Usage.md)
- [x] I added new secrets in the files [env_file_app_template](https://github.com/intelowlproject/IntelOwl/blob/master/env_file_app_template), [env_file_app_ci](https://github.com/certego/IntelOwl/blob/master/env_file_app_travis) and in the docs: [Installation](./Installation.md)
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.
  
# Real World Example
image


sp35:
>Failed to move to new namespace: PID namespaces supported, Network namespace supported, but failed: errno = Operation not permitted
>

@mlodic @Eshaan7 

setting a custom `seccomp` seems an option to me than allowing unprivileged user namespaces or `--no-sandbox` flag.
https://stackoverflow.com/questions/50662388/running-headless-chrome-puppeteer-with-no-sandbox/53975412#53975412

mlodic:
> > Failed to move to new namespace: PID namespaces supported, Network namespace supported, but failed: errno = Operation not permitted
> 
> @mlodic @Eshaan7
> 
> setting a custom `seccomp` seems an option to me than allowing unprivileged user namespaces or `--no-sandbox` flag.
> https://stackoverflow.com/questions/50662388/running-headless-chrome-puppeteer-with-no-sandbox/53975412#53975412

Yep, you can try it with the default `seccomp` config and let us know

sp35:
> Yep, you can try it with the default `seccomp` config and please let us know if it works
>
Do you mean [this one](https://github.com/moby/moby/blob/master/profiles/seccomp/default.json) ? It's not working.

api_app/script_analyzers/classes.py
```diff
@@ -294,6 +295,19 @@ def __poll_for_result(self, req_key):
             raise AnalyzerRunException("max polls tried without getting any result.")
         return json_data
 
+    def _handle_test_run(self):
+        if hasattr(self, "is_test") and getattr(self, "is_test"):
+            # only happens in case of testing
+            self.report["success"] = True
+            return {}
```
mlodic, line 302: could this be the problem. In this case the `return` would just exit from the new function instead of the `_docker_run` function

mlodic: I got anything else about the error for Thug.

sp35: my bad,  thank you

configuration/analyzer_config.json
```diff
@@ -697,6 +697,18 @@
...
+  "Rendertron": {
+    "type": "observable",
+    "observable_supported": [
+      "url"
+    ],
+    "description": "get screenshot of a web page using rendertron (puppeteer)",
+    "python_module": "rendertron.Rendertron",
+    "requires_configuration": true,
+    "additional_config_params": {},
+    "soft_time_limit": 600,
```
mlodic, line 709: isn't this too much maybe?

mlodic (commented):
it seems good work to me. Please just reduce the soft time limit.

I guess we should just add an issue for the frontend to add the visualization of the encoded image there. Otherwise this would be useful only if used programmatically

configuration/analyzer_config.json
sp35: Yep, you are right. This was for testing purpose as the proxy I was using was too slow.
I'll change it to 120s.

Eshaan7:
> This was for testing purpose as the proxy I was using was too slow.
I'll change it to 120s.

That is a thing for consideration. Given that many users may be using a proxy (which are generally slow) to use this analyzer, it may make sense to keep the default time limit a little high? Maybe around 300 seconds.

integrations/rendertron/Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM node:12-slim
```
Eshaan7, line 1: Do we really need our own Dockerfile for rendertron? Do they not provide their own docker image on docker hub that we could directly use?

integrations/rendertron/compose-tests.yml
```diff
@@ -0,0 +1,21 @@
...
+    build:
+      context: ../integrations/rendertron
+      dockerfile: Dockerfile
+    container_name: intelowl_rendertron
+    restart: unless-stopped
+    expose:
+      - "4006"
+    env_file:
+      - env_file_integrations
+    volumes:
```
Eshaan7, line 18: Probably not required here since renderton does not the log the way other containers do.

integrations/rendertron/Dockerfile
sp35: No, they don't provide now. Ref https://github.com/GoogleChrome/rendertron#deploying-using-docker

Eshaan7:
Thank you once again!

mlodic:
Yep, good work indeed

PR #456, opened 2021-05-21
Refactor nginx, compose files | hot reload django when DEBUG=True

Eshaan7:
## Type of change

# Checklist

### DIFF
--- a/configuration/nginx_errors.conf
+++ b/configuration/nginx_errors.conf
@@ -1,0 +1,54 @@
+error_page 500 /500.json;
+location /500.json {
+    return 500 '{"code":500,"message":"Internal Server Error"}';
+}
+
+error_page 502 /502.json;
+location /502.json {
+    return 502 '{"code":502,"message":"Bad Gateway"}';
+}
+
+error_page 503 /503.json;
+location /503.json {
+    return 503 '{"code":503,"message":"Service Temporarily Unavailable"}';
+}
+
+error_page 504 /504.json;
+location /504.json {
+    return 504 '{"code":504,"message":"Gateway Timeout"}';
+}
+
+error_page 400 /400.json;
+location /400.json {
+    return 400 '{"code":400,"message":"Bad Request"}';
+}
+
+error_page 401 /401.json;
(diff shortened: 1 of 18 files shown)


configuration/nginx_errors.conf
```diff
@@ -0,0 +1,54 @@
+error_page 500 /500.json;
```
mlodic, line 1: out of curiosity, why you added this?

configuration/nginx_http.conf
```diff
@@ -3,33 +3,39 @@ upstream django {
...
-        alias /var/www/static/;
-   }
+    location = /favicon.ico { 
+        access_log off;
+        log_not_found off; 
+    }
 
-   location ~^/(admin|api) {
+    # Finally, send all non-media requests to the Django server.
+    location ~^/(admin|api|silk) {
```
mlodic, line 18: again out of curiosity, `silk` is for?

mlodic (approved):

configuration/nginx_errors.conf
Eshaan7: When nginx encounters error, it responds back with HTML. These `error_page` are to overwrite this behavior to respond back with JSON and not HTML. If you might have noticed, sometimes in the IntelOwl-ng and pyintelowl we are not able to parse the error response correctly since it's HTML but they expected JSON.

configuration/nginx_http.conf
Eshaan7: The `silk/` endpoint is where the interface lives for the [django-silk](https://github.com/jazzband/django-silk) app. I intend to add this in IntelOwl as well so we can analyse performance and optimize. (#457).

configuration/nginx_errors.conf
mlodic: thank you for the explanation!

configuration/nginx_http.conf
mlodic: ah yes, this is the project you talked me about. 👍🏻

Issue #397, opened 2021-04-05
[refactoring] sub-apps: `analyzers_manager` and `connectors_manager`


Eshaan7:
- It would be better if the `api_app/script_analyzers` were made into it's own django sub-app since we are now also going to be managing models, serializers and viewset for it now. (ref. #392)
- For the connectors (ref. #249), we will have another `connectors_manager` django sub-app which follows the same pattern as we have in the analyzers (base class, config file, dynamic imports, and corresponding models/serializers/viewsets).

Here is the directory structure I have in mind:

```
IntelOwl/                                           project root
|- api_app/                                         django app for main "API"
|- admin.py
|- models.py                                        includes: `Job`, `Tag`
|- serializers.py
|- views.py
|- signals.py                                       includes: `on_job_analysis_finished`
|  |- analyzers_manager/                            sub-app for "analyzers"
|    |- observable_analyzers/                       same as it is right now
|    |- file_analyzers/                             same as it is right now
|    |- classes.py                                  includes: `BaseAnalyzer` class, etc.
|    |- admin.py         
|    |- models.py                                   includes: `Analyzer`, `AnalyzerResult`    
|    |- serializers.py
|    |- views.py
|  | -connectors_manager/                          sub-app for "connectors"
|    |- connectors/                                similar to the analyzers, this directory will hold the individual python modules.
|    |- admin.py         
|    |- models.py                                  includes: `Connector`, `ConnectorCallSummary`
|    |- serializers.py
|    |- views.py
```

Important Points:
- The `observable_analyzers`, `file_analyzers` and `connectors` directories would be marked as docker volumes to allow users to add or edit these python modules without the need to rebuild the docker image.
- We should define SQL relationships between the models `Job`, `Analyzer` and `Connector` to store as much useful information as possible. (For example, `Job.connectors_called` as a many to many field with `ConnectorCallSummary` and `Job.analyzers_called` as a many to many field with `AnalyzerResult`).
- The `AnalyzerResult` model would be modelled after the [`basic_report_template`](https://github.com/intelowlproject/IntelOwl/blob/caf4bc9ea60ce6e60a4a2c203ab9a4a8b2e81264/api_app/script_analyzers/utils.py#L65).

(Ofcourse, this is tonnes of refactoring and rewriting but what better time than GSoC to do it.)

Issue #392, opened 2021-04-05
Create `Analyzer' model, serializer, viewset to allow storing/editing.


Eshaan7:
- The model should be based on the new format of the `analyzer_config.json` that is part of issue #306. 
- We should also think about combining the `BaseAnalyzer` class with this new model.

PR #473, opened 2021-06-07
refactored script_analyzers into analyzers_manager Django sub-app

m0mosenpai:
# Description

Refactored `script_analyzers` to it's own Django sub-app: `analyzers_manager`, following the directory structure mentioned in #397. Added basic code for the new `Analyzer`, `Secrets`, `AnalyzerReport` models.

## Related issues
#397 , #392 

## Type of change

# To-Do

- [x] Creating `analyzers_manager` sub-app and copying `script_analyzers` into it.
- [x] Replacing `script_analyzers` instances in analyzer files, tests and docs
- [x] Creating new Analyzer, Secret and AnalyzerReport models
- [x] Incorporating `Analyzer` model in filtering, executing jobs.
- [x] Changes in how reports are generated (utils.py, classes.py)
- [x] Registering models in Admin interface
- [ ] Database Migrations for analyzer_config.json
- [ ] Refactoring everything (cleaner code, adding type information, docstrings etc.)

# Checklist
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

# Real World Example

TBD


m0mosenpai:
I have made a preliminary PR for refactoring and adding mostly the code @Eshaan7 had written before along with some more additions. I haven't removed existing `script_analyzers` yet. Was thinking if the directory structure is correct, I will continue to build on it and remove the old folder. (this way i won't break functionality immediately).

mlodic:
are all of these new files? It is important to keep the history while moving the files.

The python files should be moved, so we cannot keep both versions.
This is an example of what we should see from Github:
image

It is also very difficult to understand what was changed in this way because Github understand that everything is new in this way.

This is a developing branch so do not be afraid of breaking functionality. We will merge this on `develop` only when all the work will be completed

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,87 @@
...
+from cache_memoize import cache_memoize
+
+from intel_owl.secrets import get_secret
+from api_app.models import Job
+
+
+class Analyzer(models.Model):
+    analyzer_type = models.CharField(
+        max_length=50,
+        choices=(
```
Eshaan7, line 12: can you extract out the `choices` into a constant variable inside the `Analyzer` model ? thanks

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,87 @@
...
+    started_time = models.TimeField()
+    started_time_str = models.TextField()
+
+
+class Secret(models.Model):
+    name = models.CharField(max_length=50, blank=False, null=False, unique=True)
+    env_variable_key = models.CharField(max_length=50)
+    datatype = models.CharField(
+        max_length=8,
+        choices=(("str", "str"), ("int", "int"), ("bool", "bool"), ("float", "float")),
```
Eshaan7, line 79: extract out into a constant variable `CHOICES` inside Secret class

api_app/analyzers_manager/views.py
```diff
@@ -0,0 +1,9 @@
+# from django.shortcuts import render
+# from rest_framework import serializers as rfs
+from rest_framework import viewsets
+
+from api_app.analyzers_manager.serializers import AnalyzerSerializer
+
+
+class AnalyzerViewSet(viewsets.ModelViewSet):
```
Eshaan7, line 8: use `viewsets.ReadOnlyModelViewSet` because we do not wish to allow update/create at the moment.

intel_owl/settings.py
```diff
@@ -50,6 +50,7 @@
     "durin",
     "guardian",
     "api_app.apps.ApiAppConfig",
+    "api_app.analyzers_manager.apps.AnalyzersManagerConfig",
```
Eshaan7, line 53: Since the `api_app` is a django app and the `analyzers_manager` only a sub-app, we shouldn't add it here.

This also means no need for the `analyzers_manager/apps.py` file.

Same is true for `connectors_manager`, so ping @sp35.

api_app/analyzers_manager/admin.py
```diff
@@ -0,0 +1,3 @@
+# from django.contrib import admin
+
+# Register your models here.
```
Eshaan7, line 3: I guess it is already in your to-do list, but we would need admin interfaces against all the new models here.

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,87 @@
...
+            "missing_secrets": [],
+        }
+        for secret in self.secrets.all():
+            var = get_secret(secret.env_variable_key)
+            if var:
+                self._cached_secrets[secret.name] = var
+            else:
+                verification["missing_secrets"].append(secret.name)
+
+        if len(verification["missing_secrets"]):
```
Eshaan7, line 44: isn't this condition wrong ? if there any `missing_secrets`, that means `configured` should be False.

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,87 @@
...
+    python_module = models.CharField(max_length=128, blank=False, null=False)
+    config = postgres_fields.JSONField(default=dict, blank=False, null=False)
+
+    @property
+    def _cached_secrets(self) -> dict:
+        return self.__cached_secrets
+
+    @_cached_secrets.setter
+    def _cached_secrets(self, key, val):
+        self.__cached_secrets[key] = val
```
Eshaan7, line 28: remove this. it's broken anyway.

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,87 @@
...
+    def verify_secrets(self) -> dict:
+        verification = {
+            "configured:": False,
+            "error_message": None,
+            "missing_secrets": [],
+        }
+        for secret in self.secrets.all():
+            var = get_secret(secret.env_variable_key)
+            if var:
+                self._cached_secrets[secret.name] = var
```
Eshaan7, line 40: replace with `self.__cached_secrets[secret.name]` note the extra underscore.

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,87 @@
...
+            else:
+                verification["missing_secrets"].append(secret.name)
+
+        if len(verification["missing_secrets"]):
+            verification["configured"] = True
+        else:
+            total_secrets = len(self.secrets.all())
```
Eshaan7, line 47: using `self.secrets.all().count()` is more performant and django-recommended way.

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,87 @@
...
+
+class AnalyzerReport(models.Model):
+    analyzer = models.ForeignKey(
+        Analyzer, related_name="reports", on_delete=models.CASCADE
+    )
+    job = models.ForeignKey(
+        Job, related_name="analyzer_reports", on_delete=models.CASCADE
+    )
+
+    success = models.BooleanField()
```
Eshaan7, line 64: add `default=False`.

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,87 @@
...
+    report = postgres_fields.JSONField(default=dict, blank=False, null=False)
+    errors = postgres_fields.ArrayField(
+        models.CharField(max_length=512, blank=True, default=list)
+    )
+    process_time = models.FloatField()
+    started_time = models.TimeField()
+    started_time_str = models.TextField()
```
Eshaan7, line 71: remove the started_time_str field, it's pretty redundant.

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,87 @@
...
+        Job, related_name="analyzer_reports", on_delete=models.CASCADE
+    )
+
+    success = models.BooleanField()
+    report = postgres_fields.JSONField(default=dict, blank=False, null=False)
+    errors = postgres_fields.ArrayField(
+        models.CharField(max_length=512, blank=True, default=list)
+    )
+    process_time = models.FloatField()
+    started_time = models.TimeField()
```
Eshaan7, line 70: should be a `models.DateTimeField`

Eshaan7: add a `end_time` field too

intel_owl/settings.py
sp35: How would django recognize this sub app then (e.g. for making migrations)?

m0mosenpai: Even I made this change after finding that a sub-app is independent from the main app and would require it to be added in the settings.py separately. I'll test it out without this line as well once

api_app/analyzers_manager/models.py
m0mosenpai: yeah, my bad

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,85 @@
...
+        max_length=8,
+        choices=CHOICES,
+    )
+    required = models.BooleanField(blank=False, default=True)
+    default = models.CharField(max_length=50, null=True)
+    description = models.TextField()
+
+    analyzer = models.ForeignKey(
+        Analyzer, related_name="secrets", on_delete=models.CASCADE
+    )
```
sp35, line 85: This may be done using a through model `AnalyzerSecret` with one to one fields of `Analyzer` and `Secret` and correspondingly `ConnectorSecret`. This way I can use the `Secret` model for `Connector` also.

api_app/analyzers_manager/utils.py
```diff
@@ -65,16 +67,17 @@ def set_report_and_cleanup(analyzer_name, job_id, report):
...
-        "started_time_str": timezone.now().strftime("%Y-%m-%d %H:%M:%S"),
-    }
+def get_report_model_object():
+    report_obj = AnalyzerReport(
+        success=False,
+        report={},
+        errors=[],
+        start_time=datetime.now(),
+        end_time=datetime.now(),
+    )
```
m0mosenpai, line 77: I assume, I will also need to pass in `Analyzer` and `Job` model objects (since `AnalyzerModel` has a foreign key relationship with them? If so, for `Job`, i'll need to do some refactoring but i was confused how to incorporate the `Analyzer` model here.

intel_owl/settings.py
Eshaan7: AFAIK and I have used it like this before, just adding the "api_app" to the list should be sufficient. Django recursively traverses all sub-directories.

api_app/analyzers_manager/models.py
Eshaan7: ^ yep, definetly correct.

api_app/analyzers_manager/utils.py
Eshaan7: This function shouldn't be inside the `utils.py`. Remember that we are refactoring the codebase to make it better so we don't necessarily need to adopt previous code. 

A better approach inline with new changes would be to have a member method `init_report` inside `Analyzer` model. The signature being something like: `def init_report(self) -> AnalyzerReport`

Eshaan7:
Could you include a todo type list in this PR ? like create a checkbox list of everything that you aim to do in this PR and keep crossing them off as you progress ? that way it will also be easier for us to track progress. Thanks.

api_app/analyzers_manager/utils.py
Eshaan7: Another way could be to have this method inside `Job` model. `def init_report(self, analyzer: Analyzer) -> AnalyzerReport`. I'll let you decide which one is better as you are coding it.

api_app/analyzers_manager/classes.py
```diff
@@ -1,6 +1,7 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
+from datetime import datetime
```
Eshaan7, line 4: `from django.utils import timezone` and then use `timezone.now()`. Otherwise django throws naive datetime warnings.

api_app/analyzers_manager/models.py
Eshaan7: To enable re-usability of `Secret` model, it should be placed inside a different `models.py` to avoid circular import issues.

intel_owl/settings.py
Eshaan7: @m0mosenpai when you ran `makemigrations`, did it create migration files inside `api_app/migrations` or `api_app/analyzers_manager/migrations/` ?

m0mosenpai: Yeah, i got a `0001_initial.py` in `analyzers_manager/migrations` and a `0005_auto_xxxx.py` in `api_app/migrations`

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,85 @@
...
+
+class AnalyzerReport(models.Model):
+    analyzer = models.ForeignKey(
+        Analyzer, related_name="reports", on_delete=models.CASCADE
+    )
+    job = models.ForeignKey(
+        Job, related_name="analyzer_reports", on_delete=models.CASCADE
+    )
+
+    success = models.BooleanField(default=False)
```
Eshaan7, line 61: Instead of a boolean field like this, i think better would be to have a `status = models.CharField(choices=...)`, with choices like: `pending`, `running`, `failed`, `success`.

intel_owl/settings.py
Eshaan7: okay then that's good so we should keep the line including the sub-app in `INSTALLED_APPS`.

api_app/analyzers_manager/utils.py
m0mosenpai: I put it in Job cause the report model as i understood, represents a Job's report (which may contain individual reports from different analyzers) so it made more sense to me that way

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,128 @@
...
+        ("observable_analyzer", "observable_analyzer"),
+    )
+    QUEUE_CHOICES = (
+        ("default", "default"),
+        ("long", "long"),
+        ("local", "local"),
+    )
+    HASH_CHOICES = (("md5", "md5"), ("sha256", "sha256"))
+
+    name = models.CharField(max_length=128, blank=False, null=False, unique=True)
```
sp35, line 20: can be skipped: `blank` and `null` are `False` by default

sp35: Please omit this everywhere!

sp35 (changes requested):

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,128 @@
...
+        default="md5",
+    )
+    observable_supported = postgres_fields.ArrayField(
+        models.CharField(default=list, max_length=50, blank=True, null=True)
+    )
+    leaks_info = models.BooleanField(default=False, blank=True, null=True)
+    external_service = models.BooleanField(default=False, blank=True, null=True)
+    queue = models.CharField(max_length=50, choices=QUEUE_CHOICES, default="default")
+    soft_time_limit = models.IntegerField(default=300)
+    additional_config_params = models.JSONField(default=dict, blank=True, null=True)
```
sp35, line 49: I feel there is no use in keeping `blank` and `null` when you have set a default value.

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,128 @@
...
+        choices=HASH_CHOICES,
+        default="md5",
+    )
+    observable_supported = postgres_fields.ArrayField(
+        models.CharField(default=list, max_length=50, blank=True, null=True)
+    )
+    leaks_info = models.BooleanField(default=False, blank=True, null=True)
+    external_service = models.BooleanField(default=False, blank=True, null=True)
+    queue = models.CharField(max_length=50, choices=QUEUE_CHOICES, default="default")
+    soft_time_limit = models.IntegerField(default=300)
```
sp35, line 48: Maybe use `PositiveIntegerField`?

api_app/analyzers_manager/models.py
m0mosenpai: I have changed where I have used `blank` and `null` based on some documentation I read. There were places it was not good of me to use them, so removed from there. But otherwise, i would prefer to have `blank` at other places for the sake of being a bit explicit.

api_app/analyzers_manager/models.py
```diff
@@ -0,0 +1,129 @@
...
+        max_length=50,
+        choices=STATUS_CHOICES,
+        blank=False,
+    )
+    report = models.JSONField(default=dict, blank=False)
+    errors = postgres_fields.ArrayField(
+        models.CharField(max_length=512, blank=True, default=list)
+    )
+    start_time = models.DateTimeField()
+    end_time = models.DateTimeField()
```
Eshaan7, line 108: use these same fields in the `ConnectorReport`.

PR #499, opened 2021-06-22
Initialize Connectors Manager sub app

sp35:
# Description
- [x] Add `connectors_manager` sub app
- [x] Add config file and verification
- [x] Cache the verified json
- [x] Validate keys of `secrets` in serializer
- [x] Add connector report serializer 

## Related issues
#397

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.


api_app/connectors_manager/serializers.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    serializer for connectors from connector_config.json.
+    """
+
+    disabled = serializers.BooleanField()
+    description = serializers.CharField(max_length=512)
+    python_module = serializers.CharField(max_length=128)
+    config = serializers.JSONField()
+    secrets = serializers.JSONField()
+    verification = serializers.SerializerMethodField()
```
m0mosenpai, line 26: What is this field for? I didn't really understand `SerializerMethodFields` and what they do...

api_app/connectors_manager/api.py
```diff
@@ -0,0 +1,20 @@
...
+# See the file 'LICENSE' for copying permission.
+
+from rest_framework import generics
+from rest_framework.response import Response
+
+from api_app.connectors_manager.serializers import ConnectorConfigSerializer
+from api_app.connectors_manager import helpers
+
+
+class ConnectorListAPI(generics.ListAPIView):
```
m0mosenpai, line 11: Can we do this using `Viewsets`? If so, any particular reason of choosing one over the other for our case?

api_app/connectors_manager/serializers.py
Eshaan7: https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield

api_app/connectors_manager/api.py
Eshaan7: We use viewsets when we have corresponding models using which we can define a `queryset`. But since we don't have model for Connector, we use a generic view.

- In a viewset, you define methods like `create`, `list`, `destroy` which corresponds to the HTTP methods `POST`, `GET`, `DELETE` and performs operations on the defined queryset and model. (mostly you don't have to define these methods yourself because DRF takes care of it)
- In a view class, you can define methods like `get`, `post` etc. which correspond to the HTTP methods. So this is useful when you wish to define a custom view that does not necessarily depend on a model/queryset.

For further clarification, please refer to the DRF docs, they are pretty well written.

api_app/connectors_manager/serializers.py
Eshaan7: Basically `verification` will be populated by calling the `get_verification(self, obj)` method. So it's a  dynamic field that gets it value from a function call.

api_app/connectors_manager/serializers.py
```diff
@@ -25,6 +68,11 @@ class ConnectorConfigSerializer(serializers.Serializer):
     secrets = serializers.JSONField()
     verification = serializers.SerializerMethodField()
 
+    def validate_secrets(self, secrets):
+        serializer = SecretSerializer(data=list(secrets.items()), many=True)
+        if serializer.is_valid(raise_exception=True):
```
Eshaan7, line 73: the `if` is not necessary,

api_app/connectors_manager/helpers.py
```diff
@@ -2,15 +2,45 @@
...
+from api_app.connectors_manager.serializers import ConnectorConfigSerializer
 
-def get_connector_config():
+
+logger = logging.getLogger(__name__)
+
+
+def get_verified_connector_config():
+    # get cached json else cache and return
+    config = cache.get("verified_connectors_config", None)
```
Eshaan7, line 19: instead of `cache.get` or `cache.set` use the `@cache_memoize` decorator from https://github.com/peterbe/django-cache-memoize. The package is already included in IntelOwl.

api_app/connectors_manager/helpers.py
```diff
@@ -2,15 +2,45 @@
 # See the file 'LICENSE' for copying permission.
 
 from django.conf import settings
+from django.core.cache import cache
 
 import os
 import json
+import logging
 
+from api_app.connectors_manager.serializers import ConnectorConfigSerializer
```
Eshaan7, line 11: I prefer relative paths over absolute paths: `from .serializers import ConnectorConfigSerializer`.

api_app/connectors_manager/serializers.py
Eshaan7: what if you pass the data to the serializer like: `data=["{key_name": key, **secret_dict} for key, secret_dict in secrets.items()]`. This way you are passing to it a `dict` only and not a `tuple`.

api_app/connectors_manager/serializers.py
```diff
@@ -13,6 +13,49 @@
...
+        if default is not None and type(default) is not type(secret_type):
+            validation_error = {
+                self.key_name: {
+                    "default": f"should be of type {secret_type}, got {type(default)}"
+                }
+            }
+            raise serializers.ValidationError(validation_error)
+        return data
+
+    def to_internal_value(self, data):
```
Eshaan7, line 54: if done this way: https://github.com/intelowlproject/IntelOwl/pull/499/commits/d9b1872e344d58f76c454106af7fbffe1498d2dc#r656200927

then no need for handling tuple.

api_app/connectors_manager/serializers.py
```diff
@@ -25,6 +68,11 @@ class ConnectorConfigSerializer(serializers.Serializer):
...
     verification = serializers.SerializerMethodField()
 
+    def validate_secrets(self, secrets):
+        serializer = SecretSerializer(data=list(secrets.items()), many=True)
+        if serializer.is_valid(raise_exception=True):
+            return secrets
+
     def check_secrets(self, secrets):
         exceptions = {}
         for key_name, secret_dict in secrets.items():
```
Eshaan7, line 78: on line 79, the code is:

```python3
if "required" in secret_dict and secret_dict["required"]:
```

a better way to achieve this would be:

```python3
if secret_dict,get("required", False):
```

api_app/connectors_manager/helpers.py
```diff
@@ -2,15 +2,45 @@
...
+    if config is not None:
+        return config
+    else:
+        success, config = verify_connector_config()
+        if success:
+            cache.set("verified_connectors_config", config)
+        return config
+
+
+def verify_connector_config():
```
Eshaan7, line 29: I'd prefer this method to be under the `ConnectorConfigSerializer` as a `@classmethod` with name `read_and_verify_config`. Then we can do something like: `ConnectorConfigSerializer.read_and_verify_config()`; 

I mean, it's just better to keep related code together in one class.

api_app/connectors_manager/helpers.py
```diff
@@ -1,46 +1,15 @@
...
-            else:
-                serializer_errors[key] = serializer.errors
-
-        if bool(serializer_errors):  # returns False if empty
-            logger.error(f"connector config serializer failed: {serializer_errors}")
-            return False, {}
-        return True, connector_config
+    success, config = ConnectorConfigSerializer.read_and_verify_config()
+    if success:
+        cache.set("verified_connectors_config", config)
```
Eshaan7: This is buggy. You cannot mix `cache_memoize` and `cache.set` like this. What u should do is not return a tuple of (success, config) instead raise `ValidationError` inside `read_and_verify_config` if verification failed. `cache_memoize` won't cache exceptions so the result will be cached only if were successful.

api_app/connectors_manager/serializers.py
```diff
@@ -68,15 +70,40 @@ class ConnectorConfigSerializer(serializers.Serializer):
...
+                if serializer.is_valid():
+                    connector_config[
+                        key
+                    ] = serializer.data  # mutate with processed config
+                else:
+                    serializer_errors[key] = serializer.errors
+
+            if bool(serializer_errors):  # returns False if empty
+                logger.error(f"connector config serializer failed: {serializer_errors}")
+                return False, {}
```
Eshaan7, line 92: raise `rfs.ValidationError(serializer_errors)`.

api_app/connectors_manager/helpers.py
sp35: I removed that in the last commit.

That seems right.

sp35: This will raise 400 now, earlier it was returning empty.
TODO: Handle on UI with something like "There are some errors in the configuration file"

api_app/connectors_manager/serializers.py
```diff
@@ -133,3 +133,33 @@ def get_verification(self, raw_instance):
...
+
+    class Meta:
+        model = ConnectorReport
+        fields = (
+            "name",
+            "success",
+            "report",
+            "errors",
+            "process_time",
+            "started_time_str",
```
sp35, line 155: To be changed to the new format when `AnalyzerReport` is up (since UI breaks now)

api_app/migrations/0003_auto_20201020_1406.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("api_app", "0002_added_job_field"),
+        # added dependency to enable using models from app2 in move_m1
+        ("durin", "0001_initial"),
+    ]
+
+    operations = [migrations.RunPython(create_default_clients)]
```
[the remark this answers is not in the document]
sp35: @Eshaan7 thoughts?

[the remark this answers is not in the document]
Eshaan7: I think it can be safely ignored.

Eshaan7 (approved):

PR #528, opened 2021-07-12
Fixes #12: Add MISP Connector

### LINKED ISSUE #12: MISP connector
IntelOwl was designed to be integrated with other security tools very easily.
In particular, it is a perfect fit for Threat Intelligence Platforms because it can enrich the data available in those platform and can be customized on its own needs.

So, in order to help the community to achieve this result, we should add new connectors for the most used open source TIPs, for example: [MISP](https://github.com/MISP/MISP)

We should study their project and add a relative connector to integrate IntelOwl with MISP.
An idea could be to create a new expansion module (https://github.com/MISP/misp-modules) to allow MISP users to enrich an indicator via IntelOwl and integrate the results on the relative event.
We should leverage the official [IntelOwl client](https://github.com/mlodic/pyintelowl) for this purpose.

sp35:
# Description
This connector creates an event and attribute on provided MISP instance after a job analysis finished.

- [ ] Docs work to be handled in #562 

## Related issues
Fixes #12 

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

# Real World Example
```json
{"Event": {"id": "60", "Org": {"id": "1", "name": "ORGNAME", "uuid": "9d522381-861e-46a2-838c-62023b538793", "local": true}, "Tag": [{"id": "9", "name": "source:intelowl", "local": 0, "colour": "#6627b1", "user_id": "0", "hide_tag": false, "is_galaxy": false, "exportable": true, "numerical_value": null, "is_custom_galaxy": false}, {"id": "1", "name": "tlp:white", "local": 0, "colour": "#e80866", "user_id": "0", "hide_tag": false, "is_galaxy": false, "exportable": true, "numerical_value": null, "is_custom_galaxy": false}], "Orgc": {"id": "1", "name": "ORGNAME", "uuid": "9d522381-861e-46a2-838c-62023b538793", "local": true}, "date": "2021-07-14", "info": "Intelowl Job-60: file.html|12c0f7cc05cbae1475bf6c0caaa1ac70", "uuid": "371cd335-8cfd-47fb-a1b2-baca0a1c0c49", "Galaxy": [], "Object": [], "locked": false, "org_id": "1", "orgc_id": "1", "analysis": "2", "Attribute": [{"id": "86", "type": "filename|md5", "uuid": "37df934f-d2f4-4cad-9361-00291925a04c", "value": "file.html|12c0f7cc05cbae1475bf6c0caaa1ac70", "Galaxy": [], "to_ids": true, "comment": "Analyzers Executed: ['File_Info']", "deleted": false, "category": "Payload delivery", "event_id": "60", "last_seen": null, "object_id": "0", "timestamp": "1626268307", "first_seen": null, "distribution": "5", "ShadowAttribute": [], "object_relation": null, "sharing_group_id": "0", "disable_correlation": false}, {"id": "87", "type": "mime-type", "uuid": "b47ee9f1-7e65-4e06-a288-d4b4a7b06c8d", "value": "text/html", "Galaxy": [], "to_ids": false, "comment": "", "deleted": false, "category": "Artifacts dropped", "event_id": "60", "last_seen": null, "object_id": "0", "timestamp": "1626268307", "first_seen": null, "distribution": "5", "ShadowAttribute": [], "object_relation": null, "sharing_group_id": "0", "disable_correlation": false}, {"id": "88", "type": "link", "uuid": "3f78e608-565e-4c4e-b4f2-4992e4e2f35a", "value": "http://localhost/pages/scan/result/60", "Galaxy": [], "to_ids": false, "comment": "View Analysis on IntelOwl", "deleted": false, "category": "External analysis", "event_id": "60", "last_seen": null, "object_id": "0", "timestamp": "1626268307", "first_seen": null, "distribution": "5", "ShadowAttribute": [], "object_relation": null, "sharing_group_id": "0", "disable_correlation": true}], "published": false, "timestamp": "1626268307", "EventReport": [], "RelatedEvent": [{"Event": {"id": "59", "Org": {"id": "1", "name": "ORGNAME", "uuid": "9d522381-861e-46a2-838c-62023b538793"}, "Orgc": {"id": "1", "name": "ORGNAME", "uuid": "9d522381-861e-46a2-838c-62023b538793"}, "date": "2021-07-14", "info": "Intelowl Job-59: file.html|12c0f7cc05cbae1475bf6c0caaa1ac70", "uuid": "9cd59957-bcc3-4468-9eb1-d0cef5d78226", "org_id": "1", "orgc_id": "1", "analysis": "2", "published": false, "timestamp": "1626267670", "distribution": "0", "threat_level_id": "4"}}], "distribution": "0", "extends_uuid": "", "ShadowAttribute": [], "attribute_count": "3", "threat_level_id": "4", "sharing_group_id": "0", "publish_timestamp": "0", "disable_correlation": false, "event_creator_email": "admin@admin.test", "proposal_email_lock": false}}
```
<details>
      <summary>Click to expand</summary>

</details>


api_app/connectors_manager/connectors/misp.py
```diff
@@ -1,10 +1,88 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
+import pymisp
 
 from ..classes import Connector
```
Eshaan7, line 6: use absolute import. `from api_app.blah blah import Connector`

api_app/connectors_manager/connectors/misp.py
```diff
@@ -1,10 +1,94 @@
...
+
+    @property
+    def _link_attr_obj(self):
+        """
+        Returns attribute linking analysis on IntelOwl instance
+        """
+        obj = pymisp.MISPAttribute()
+        obj.type = "link"
+        # TODO: get instance url
+        obj.value = f"http://localhost/pages/scan/result/{self.job_id}"
```
sp35, line 70: TODO

api_app/connectors_manager/connectors/misp.py
```diff
@@ -1,10 +1,94 @@
...
+        for tag in self._job.tags.all():
+            obj.add_tag(f"{tag.label}:{tag.color}")
+
+        return obj
+
+    @property
+    def _base_attr_obj(self):
+        if self._job.is_sample:
+            type = "file"
+            value = f"{self._job.file_name} ({self._job.file_mimetype})"
```
mlodic, line 46: for the files, maybe it is better to use the attribute `filename|md5`. Otherwise we could add an additional attribute for the `md5` too.

Also, for the mimetype, we could add a separated attribute `mime-type`

See: https://www.circl.lu/doc/misp/categories-and-types/

mlodic (changes requested):
Great work! Some suggestions:

We could add a chance to customize the connector by selecting which kind of Traffic Light Protocol (TLP) value they would like to use for sharing the events generated by the integration.
TLP is very important and is the most used one taxonomy used because it regulates how the information should be shared among other entities.
We could add that value as a tag to the event, using the already existing taxonomy
(See: https://www.misp-project.org/taxonomies.html#_tlp_2)
By default, we could use the `tlp:white`.


Also, maybe `analyzers_to_execute` value could be added to the MISP report somewhere?
It can happen that the same observable is analyzed multiple times with different analyzers. Without this value, it is impossibile to understand from the MISP which kind of analysis was done

api_app/connectors_manager/connectors/misp.py
sp35: Done

api_app/connectors_manager/connectors/misp.py
sp35: I can't do much about this, does it make sense to ask for it in the config?

```json
...
config: {
    "intelowl_instance_url": "http://myintelowl.com"
}


```

mlodic: Considering that this is more a "project" config rathen than a "connector" config, could it make sense to add that value as a variable in `settings.py` and extract the configuration from the `env_file_app` if set? 
Then we could set a default like "localhost" if it is not set

sp35: Yep, that's better, should have thought of that.

api_app/helpers.py
```diff
@@ -64,3 +65,21 @@ def get_binary(job_id, job_object=None):
...
+        "md5": re.compile(r"^[a-f\d]{32}$", re.IGNORECASE | re.ASCII),
+        "sha1": re.compile(r"^[a-f\d]{40}$", re.IGNORECASE | re.ASCII),
+        "sha256": re.compile(r"^[a-f\d]{64}$", re.IGNORECASE | re.ASCII),
+        "sha512": re.compile(r"^[a-f\d]{128}$", re.IGNORECASE | re.ASCII),
+    }
+
+    for hash_type, re_hash in RE_HASH_MAP.items():
+        if re.match(re_hash, hash_value):
+            return hash_type
+    return None  # no matches
```
mlodic, line 85: In this function, having 2 different `return` code is not a particular problem because the function is short and we all get what is happening.

Anyway, it would be cleaner to just use `return` only a single time in the code. In this way it is easier to read by other programmers because there is no hidden behavior that could led to return the function before expected.

So, in this case, you can just use a variable, change it when you find a match and then return that variable only a single time at the end of the code.

api_app/connectors_manager/connectors/misp.py
```diff
@@ -1,10 +1,128 @@
...
+        # add attributes to event on MISP Instance
+        for attr in attributes:
+            misp_instance.add_attribute(misp_event.id, attr)
+
+        return misp_instance.get_event(misp_event.id)
+
+    @classmethod
+    def _monkeypatch(cls):
+        # completely skip because doesnt work without connection.
+        patches = [if_mock_connections(patch.object(cls, "run", return_value={}))]
```
mlodic, line 127: to avoid to have 0 coverage for this code, is it possible to try to make a more advanced patching?

you can see as an example how we patched the MWdb library [here](https://github.com/intelowlproject/IntelOwl/blob/develop-2/api_app/analyzers_manager/file_analyzers/mwdb_scan.py#L18)

Is it enough to just set the attributes and methods required by our code. In this way all the logic unrelated to the library pymisp would be correctly tested.

mlodic (changes requested):
gj! we are almost done

api_app/connectors_manager/connectors/misp.py
sp35: done, I've added a mock class only for the methods which connect with the instance.

mlodic (approved):

PR #531, opened 2021-07-14
adding missing migration

mlodic:
added makemigrations for durin package only to allow customization of its parameters.

@Eshaan7 do you think there is a cleanest way to do this? maybe directly in the durin package?

In this way the durin migration would be applied each time the IntelOwl project is brought up. Not of a big problem but still unecessary


Eshaan7:
Unfortunately, the way the `durin` settings are configured makes migrations a bit painful. At the moment, I have no elegant solution other than what you implemented.

PR #532, opened 2021-07-14
Fixes #540: tests analyzer scripts, use monkeypatch in `Plugin`, update GH action, refactoring

### LINKED ISSUE #540: Dynamic testing of analyzers, instead of having to define individual test cases

Eshaan7:
**TODO:**
- [x] define `_monkeypatch` methods in all analyzer subclasses. (except docker based analyzers)
- [x] refactor testing suite (if any bugs/issues found)
- [x] update workflow file


api_app/analyzers_manager/file_analyzers/cuckoo_scan.py
```diff
@@ -10,9 +10,27 @@
...
+        ),
+        patch(
+            "requests.Session.post",
+            side_effect=mocked_requests,
+        ),
+    ]
+)
 class CuckooAnalysis(FileAnalyzer):
     def set_params(self, params):
         # cuckoo installation can be with or without the api_token
```
sp35, line 36: `self. __api_key` on line 40

tests/analyzers_manager/__init__.py
```diff
@@ -39,64 +44,140 @@ def get_params(cls):
...
+            "start_analyzers",
+            args=[
+                self.test_job.pk,
+                self.test_job.analyzers_to_execute,
+                self.runtime_configuration,
+            ],
+            queue="default",
+        )
+
+        for i in range(0, 1000):
```
mlodic, line 80: maybe too much 5000 seconds?

tests/analyzers_manager/__init__.py
```diff
@@ -39,64 +44,140 @@ def get_params(cls):
...
+        return super().setUp()
+
+    def test_start_analyzers_all(self, *args, **kwargs):
+        print("\n[START] ---test_start_analyzers_all---")
+        print(
+            f"[REPORT] Job:{self.test_job.pk}, status:'{self.test_job.status}', analyzers:{self.test_job.analyzers_to_execute}"
+        )
+
+        # execute analyzers
+        celery_app.send_task(
```
mlodic, line 70: I am not sure what happens with `celery_app.send_task(` but surely celery is not executed. We could just execute the function directly without involving celery

tests/analyzers_manager/__init__.py
Eshaan7: yeah, I didnt think about this. I  have to implement a base check + timeout yet.

tests/analyzers_manager/__init__.py
Eshaan7: the `start_analyzers` function inside fires a celery task for each analyzer, so the problem will still persist.

PR #544, opened 2021-07-15
Fixes #539: Update Plugin and subclasses to use dataclasses

### LINKED ISSUE #539: use dataclasses and not serializer in `Plugin`
Now that we have the dataclasses, we should use them only in the `Plugin` (and it's subclasses).

Changes required:

- [x]  `Plugin.__init__` will take an instance of `config: AbstractConfig` and not `config_dict: Dict`. (and saved as `self._config`) 
     * changes will be required in `Plugin`, `ObservableAnalyzer`, `FileAnalyzer`, `Connector` classes and `run_analyzer` and `run_connector` functions.
      
- [x] `_serializer` cached_property: not needed anymore.
- [x] `_params` cached_property:  make it a normal `@property` and `return self._config.config`
- [x]  The `_read_secrets` method on `AbstractConfigSerializer` should be move to `AbstractConfig` dataclass instead.
- [x] `_secrets` cached_property: `return self._config._read_secrets()`
- [x] other bug fixes that might arise because of this change

sp35:
Fixes #539 

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

### DIFF
--- a/api_app/analyzers_manager/classes.py
+++ b/api_app/analyzers_manager/classes.py
@@ -15,6 +15,8 @@ from api_app.core.classes import Plugin
 from api_app.helpers import generate_sha256
 
 from .models import AnalyzerReport
+from .dataclasses import AnalyzerConfig
+from .constants import HashChoices, ObservableTypes, TypeChoices
 from .serializers import AnalyzerConfigSerializer
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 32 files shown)


Eshaan7 (approved):
LGTM!

PR #545, opened 2021-07-16
Flower + Celery > 5.0.0 

0ssigeno:
# Description

Upgrade celery to version 5+ .
Fixed Click conflict with Quark Engine
Added Flower to manage celery worker at runtime 

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.
  
### please follow these rules


docker/flower.override.yml
```diff
@@ -0,0 +1,26 @@
+version: "3"
+
+services:
+  rabbitmq:
+    image: library/rabbitmq:management
```
mlodic, line 5: it is better to set also the version to avoid unexpected problems

configuration/nginx_https.conf
```diff
@@ -9,6 +9,31 @@ server {
     return 301 https://intel_owl.com$request_uri;
 }
 
+# flower
```
mlodic, line 12: I think that also `nginx_http.conf` needs the respective configuration

docker/flower.override.yml
```diff
@@ -0,0 +1,26 @@
...
+  flower:
+    image: intelowlproject/intelowl:${INTELOWL_TAG_VERSION}
+    container_name: intelowl_flower
+    restart: unless-stopped
+    stop_grace_period: 3m
+    volumes:
+      - generic_logs:/var/log/intel_owl
+    expose:
+      - "5555"
+    ports:
```
mlodic, line 19: shouldn't the ports be set on the `nginx` service?

docker/Dockerfile
```diff
@@ -15,13 +15,16 @@ RUN mkdir -p ${LOG_PATH} \
 RUN apt-get update \
     && apt-get install -y --no-install-recommends apt-utils libsasl2-dev libssl-dev \
     vim libldap2-dev python-dev libfuzzy-dev net-tools python-psycopg2 git osslsigncode \
+    apache2-utils \
```
mlodic, line 18: what is the purpose of this package?

mlodic (commented):
flower ❤️

docker/flower.override.yml
0ssigeno: You are absolutely correct

docker/Dockerfile
0ssigeno: Creation of the .httpasswd

docker/flower.override.yml
```diff
@@ -0,0 +1,33 @@
+version: "3"
+
+services:
+  rabbitmq:
+    image: library/rabbitmq:3.8.19-management
```
mlodic, line 5: sorry If I am being annoying about this, could you please just change it to only the minor version? In this way it is the same as in the non-flower configuration

docker/flower.override.yml
```diff
@@ -0,0 +1,33 @@
...
+    depends_on:
+      - rabbitmq
+    entrypoint:
+      - ./docker/entrypoint_flower.sh
+    env_file:
+      - env_file_app
+
+  nginx:
+    volumes:
+      - ./.htpasswd:/.htpasswd
```
mlodic, line 29: I think it is cleaner to put it in the nginx config folder: `/etc/nginx/.htpasswd`.

This would also need a change in the previous nginx config

mlodic (approved):

mlodic:
still errors with the dependencies 😢

0ssigeno:
Technically it is not an error: the codeql is trying to resolve the pip dependencies that are declared in the package, that right now are different from the dependency _actually_ requested from the git package.
In [this issue](https://github.com/quark-engine/quark-engine/issues/202) I have tried to explain the problem that we actually have in Quark.

requirements.txt
```diff
@@ -33,13 +33,13 @@ pypssl==2.2
 pysafebrowsing==0.1.1
 PySocks==1.7.1
 python-magic==0.4.22
-quark-engine==21.6.3
+git+https://github.com/quark-engine/quark-engine.git@v21.7.1 #pip version has conflict with celery
```
Eshaan7, line 36: https://github.com/intelowlproject/IntelOwl/pull/549

Eshaan7: @0ssigeno

0ssigeno: Issue should be still present in that version

0ssigeno: We can wait for a version of Quark that solves this issue. I don't think we are in a hurry to merge this pr

PR #547, opened 2021-07-16
Fixes #390: Split `send_analysis_request` into `analyze_file` and `analyze_observable`

### LINKED ISSUE #390: Divide `/send_analysis_request` endpoint into `/analyse_observable` and `/analyse_file`
**Note: Would also require changes in the pyintelowl and IntelOwl-ng**

On the client side, we already provide two different methods to build the request body differently based on whether user chose `observable` or `file`. So it just makes sense to divide the endpoint into two for easier management and less confusion.

m0mosenpai:
# Description

- [X] New views for endpoints
- [X] New Serializers
- [ ] Testing and Refactoring

## Related issues
Fixes #390 

## Type of change

# Checklist

- [X] `Black` gave 0 errors.
- [X] `Flake` gave 0 errors.


m0mosenpai:
The `file_mimetype` `SerializerMethodField()` doesn't get automatically populated for some reason. This results in `KeyError` in the `filter_analyzers()` when it checks that key in the `serialized_data`. What am I missing here?

Eshaan7:
> The `file_mimetype` `SerializerMethodField()` doesn't get automatically populated for some reason. This results in `KeyError` in the `filter_analyzers()` when it checks that key in the `serialized_data`. What am I missing here?

Ah, that's an issue I didn't think about. That's because the `SerializerMethodField` is by default a `read_only` field. So it doesn't get populated because creation is a `write` operation. I guess you could try making it a `HiddenField(default=calculcate_mimetype)` field.

Eshaan7:
Please merge your branch with latest changes of `develop-2` once fixing the merge conflicts.

PR #551, opened 2021-07-19
Fixes #390: Split `send_analysis_request` into `analyze_file` and `analyze_observable`

### LINKED ISSUE #390: Divide `/send_analysis_request` endpoint into `/analyse_observable` and `/analyse_file`

m0mosenpai:
# Description

- [X] New views for endpoints
- [X] New Serializers
- [x] Testing and Refactoring

## Related issues
Fixes #390 

## Type of change

# Checklist

- [X] `Black` gave 0 errors.
- [X] `Flake` gave 0 errors.

### DIFF
--- a/api_app/api.py
+++ b/api_app/api.py
@@ -53,21 +53,19 @@ def send_for_analysis(prep_data, test):
     response_dict = {
         "status": "accepted",
         "job_id": prep_data["job_id"],
-        "warnings": prep_data["warnings"],
         "analyzers_running": prep_data["analyzers_to_execute"],
     }
 
     logger.debug(response_dict)
 
-    return Response(response_dict, status=status.HTTP_200_OK)
+    return response_dict
 
 
-def prepare_for_analysis(serializer, params):
+def prepare_for_analysis(serializer, params, warnings):
     """
     Prepare file/observable for analysis
     """
 
-    warnings = []
     serialized_data = serializer.validated_data
 
     cleaned_analyzer_list = analyzers_controller.filter_analyzers(
@@ -101,7 +99,6 @@ def prepare_for_analysis(serializer, params):
         "job_id": job_id,
         "runtime_config": runtime_configuration,
         "analyzers_to_execute": cleaned_analyzer_list,
-        "warnings": warnings,
     }
 
     return prep_data
@@ -349,6 +346,7 @@ def ask_analysis_availability(request):
 @permission_required_or_403("api_app.add_job")
 def analyze_file(request):
     source = str(request.user)
+    warnings = []
     try:
         data_received = request.data
         logger.info(
@@ -364,7 +362,7 @@ def analyze_file(request):
         )
 
         if serializer.is_valid():
-            prep_data = prepare_for_analysis(serializer, params)
+            prep_data = prepare_for_analysis(serializer, params, warnings)
         else:
             error_message = f"serializer validation failed: {serializer.errors}"
             logger.error(error_message)
@@ -372,7 +370,12 @@ def analyze_file(request):
                 {"error": error_message}, status=status.HTTP_400_BAD_REQUEST
             )
 
-        return send_for_analysis(prep_data, test)
+        response_dict = {
+            **send_for_analysis(prep_data, test),
+            "warnings": warnings,
+        }
+
+        return Response(response_dict, status=status.HTTP_200_OK)
 
     except Exception as e:
         logger.exception(f"receive_analysis_request requester:{source} error:{e}.")
--- a/api_app/api.py
+++ b/api_app/api.py
@@ -6,6 +6,7 @@ import logging
 from intel_owl.celery import app as celery_app
 from api_app import models, serializers
 from api_app.permissions import ExtendedObjectPermissions
+from .exceptions import AnalyzerPreparationException
 from .analyzers_manager import controller as analyzers_controller
 
 from wsgiref.util import FileWrapper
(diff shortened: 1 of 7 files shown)


m0mosenpai:
https://github.com/intelowlproject/IntelOwl/pull/547#issuecomment-882345688

@Eshaan7 
Regarding this, It requires passing the function the `file` and `file_name` arguments. I'm not sure how i can access those values here. ```file_mimetype = serializers.HiddenField(default=calculate_mimetype(file_buffer, file_name))```

tests/test_api.py
```diff
@@ -109,9 +107,7 @@ def test_send_analysis_request_sample(self):
             "file": uploaded_file,
             "test": True,
         }
-        response = self.client.post(
-            "/api/send_analysis_request", data, format="multipart"
-        )
+        response = self.client.post("/api/analyze_file", data, format="multipart")
         self.assertEqual(response.status_code, 200)
 
     def test_send_analysis_request_domain(self):
```
Eshaan7, line 113: also rename these function like `test_analyse_observable_ip` or `test_analyse_file` etc.

Eshaan7:
I am not sure but I think it should be possible to do like:

```python3
file_mimetype = serializers.HiddenField(default=lambda obj: calculate_mimetype(obj.file, obj.file_name))
```

api_app/api.py
```diff
@@ -32,6 +33,77 @@
...
+        "job_id": prep_data["job_id"],
+        "analyzers_running": prep_data["analyzers_to_execute"],
+    }
+
+    logger.debug(response_dict)
+
+    return response_dict
+
+
+def prepare_for_analysis(serializer, params, warnings):
```
Eshaan7, line 64: While this approach with 2 common functions `send_for_analysis` and `prepare_for_anlaysis` is okay, there's a better way I just thought of:

Combine these 2 functions into an individual function 

```python3
def _analysis_request(request, serializer_class):
```

And then the `analyze_file` and `analyze_observable` would look like:

```python3
@api_view()
def analyze_file(request):
    return _analysis_request(request, FileAnalysisSerializer)

@api_view()
def analyze_observable(request):
    return _analysis_request(request, ObservableAnalysisSerializer)
```

m0mosenpai:
> I am not sure but I think it should be possible to do like:
> 
> ```python
> file_mimetype = serializers.HiddenField(default=lambda attrs: calculate_mimetype(attrs["file"], attrs["file_name"]))
> ```

I don't think this works. It says it's missing a positional requirement `attrs`. I tried looking it up online as well but didn't find examples where they had called a function with positional arguments in `HiddenField`

Eshaan7:
> I don't think this works. It says it's missing a positional argument `attrs`. I tried looking it up online as well but didn't find examples where they had called a function with positional arguments in `HiddenField`

Okay, we need to have a hacky implementation then:
- keep the `SerializerMethodField` just the way it is right now.
- overwrite `validate(self, attrs)` function of `FileAnalysisSerializer` like:

```python3
def validate(self, attrs):
  super().validate(attrs)
  attrs["file_mimetype"] = self.get_file_mimetype(attrs)
```

api_app/serializers.py
```diff
@@ -81,16 +83,10 @@ def get_no_of_connectors_executed(self, obj: Job):
 class JobSerializer(ObjectPermissionsAssignmentMixin, serializers.ModelSerializer):
```
Eshaan7, line 83: `ObjectPermissionsAssignmentMixin` and `get_permissions_map` should be moved to the 2 new serializers as well.

Eshaan7: Create a base serializer like `class JobCreateSerializerMixin(ObjectPermissionsAssignmentMixin)` to keep the common code like `validate`, `create`, `get_permissions_map` methods.

Eshaan7:
Looks good to me. Can you run the relevant tests on your local to test and post the screenshot here ? (inconvenience caused since the testing suite is broken for develop-2 branch)

m0mosenpai:
API Tests:
image

GUI:
Observable:
image

FIle:
image

api_app/api.py
```diff
@@ -199,6 +287,95 @@ def ask_analysis_availability(request):
...
+            description=r"""
+            Default {}, contains additional parameters for particular analyzers""",
+        ),
+        OpenApiParameter(
+            name="test",
+            type=OpenApiTypes.BOOL,
+            description="disable analysis for API testing",
+        ),
+    ],
+    responses={
```
Eshaan7, line 337: We do not need to specify the parameters now, set this like: `request=serializers.FileAnalysisSerializer` (refer to: vhttps://drf-spectacular.readthedocs.io/en/latest/customization.html#step-2-extend-schema)

Eshaan7: same for the `analyze_observable`.

api_app/serializers.py
```diff
@@ -120,12 +116,57 @@ def get_permissions_map(self, created):
...
+                raise serializers.ValidationError(
+                    {
+                        """
+                        analyzers_requested and run_all_available_analyzers
+                        cannot be used together
+                        """
+                    }
+                )
+
+            data["analyzers_requested"] = analyzers_controller.ALL_ANALYZERS
```
Eshaan7, line 168: Remove this line. And then in the `_analysis_request` function check the `if` condition again for this. (as was done before)

api_app/serializers.py
```diff
@@ -135,3 +181,81 @@ def create(self, validated_data):
...
+    Job model's serializer for Observable Analysis.
+    Used for create()
+    """
+
+    tags_id = serializers.PrimaryKeyRelatedField(
+        many=True, write_only=True, queryset=Tag.objects.all()
+    )
+    runtime_configuration = serializers.JSONField(
+        required=False, default={}, write_only=True
+    )
```
Eshaan7, line 240: Is it not possible to move these common fields to the `_JobCreateSerializerMixin` ?

PR #552, opened 2021-07-19
boto3_storage implementation + removed Django static file config

0ssigeno:
# Description

Please include a summary of the change.
Added support for boto3 storage 

## Type of change

# Checklist

- [] I added new secrets in the files [env_file_app_template](https://github.com/intelowlproject/IntelOwl/blob/master/docker/env_file_app_template), [env_file_app_ci](https://github.com/certego/IntelOwl/blob/master/docker/env_file_app_ci) and in the docs: [Installation](./Installation.md)
- [x] `Black` gave 0 errors.
- [ ] `Flake` gave 0 errors.


0ssigeno:
Before writing docs, I would like to discuss the implementation.
The main issue is that, we have a local copy FOR analyzer, because some of them require a filepath, because they want to do stuff on their own, and S3 does not support that.

_Why have a sample for analyzer? Couldn't we have a single shared file?_

Yes, but no. Even with the current docker composition, we need a sample for container, or a shared directory for each container (that we were doing before).
Unfortunately, this does not scale with something like docker swarm, that we are going to add to IntelOwl in the near future.


_Couldn't we save the file for each container/server?_


Nope.
If a single file for container/server was made, we would have one of these two issues:
- We have a race condition when multiple analyzer on the same container access the file
- We have to send a signal to all the containers to delete the file once the analysis is over, making it "infrastructure dependent"

Any idea/ feedback is welcome

mlodic:
I guess we can make some more brainstorm and solve this without hurry

api_app/script_analyzers/classes.py
```diff
@@ -195,15 +198,31 @@ class FileAnalyzer(BaseAnalyzerMixin):
...
+    def __init__(self, analyzer_name, job_id, md5, additional_config_params):
+        from api_app.models import Job
+
+        self.job_object = Job.object_by_job_id(job_id)
         self.md5 = md5
-        self.filepath = fpath
-        self.filename = fname
+        self.filename = self.job_object.file_name
+        self._binary = None
+        self._filepath = None
```
Eshaan7, line 225: We have made similar types of changes in `develop-2` branch. (see [here](https://github.com/intelowlproject/IntelOwl/blob/82a390fde91f9725494efc1e21c829460b1732f6/api_app/analyzers_manager/classes.py#L163-L173)).

I see that you also changed `get_binary` calls in analyzer scripts to use `self.binary` which is great but it might get lost when we merge `develop-2` into `develop`. :crying_cat_face: 

So, IMO, it will be better to have this boto3 storage implementation directly in `develop-2`.

0ssigeno: If the idea makes sense and it is approved, i can change the target branch and fix the conflicts. DW about that

api_app/helpers.py
```diff
@@ -128,17 +127,8 @@ def filter_analyzers(
...
-    logger.info(f"got job_file {job_file} for job_id {job_id}")
-
-    binary = job_file.read()
-    return binary
-
-
 def generate_sha256(job_id):
-    binary = get_binary(job_id)
-    return hashlib.sha256(binary).hexdigest()
+    from api_app.models import Job
```
Eshaan7, line 131: I think it could make sense to have this function directly be part of `Job` model. Like:

```def get_binary_sha256(self) -> str:
       return hashlib.sha256(self.file.read()).hexdigest() 
```

Eshaan7: or,

```python3
@property
def sha256(self) -> str:
```

api_app/script_analyzers/classes.py
```diff
@@ -213,6 +231,14 @@ def before_run(self):
         )
 
     def after_run(self):
+        # We delete the file only if we have single copy for analyzer
```
Eshaan7, line 234: It'd be better to have this logic be part of the `clean_old_jobs` celery task. We shouldn't immediately delete this file because of it might be needed in connectors.

0ssigeno: Mmmh no. If we do not delete the sample copies, we will instantly fill the disk space.
No race condition should be present, since each analyzer will have its own copy.

Moreover, connectors will use the main sample, the aws one, we do not actually need them anymore after the analysis is done

api_app/helpers.py
0ssigeno: Agree, i hate the `helpers.py` file. It should be better to add these functions in the correct models/file

mlodic (commented):
very good, would you be so lovely to add the documentation too?

PR #571, opened 2021-07-24
Fixes #560: use test mode instead of passing test

### LINKED ISSUE #560: [tests] Remove `test` parameter from `send_analysis_request` and docker analyzers
Instead we can use the `settings.TEST_MODE`.

sp35:
# Description
Uses global `settings.TEST_MODE` instead of passing `is_test` param.

## Related issues
Fixes #560 

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.
- [x] I squashed the commits into a single one. (optional, they will be squashed anyway by the maintainer)

### DIFF
--- a/api_app/api.py
+++ b/api_app/api.py
@@ -14,6 +14,7 @@ from wsgiref.util import FileWrapper
 from django.utils.decorators import method_decorator
 from django.http import HttpResponse
 from django.db.models import Q
+from django.conf import settings
 from rest_framework.response import Response
 from rest_framework import serializers as BaseSerializer
 from rest_framework import status, viewsets, mixins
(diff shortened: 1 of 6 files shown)


api_app/api.py
```diff
@@ -74,7 +74,7 @@ def _analysis_request(
     logger.info(f"New Job added to queue <- {repr(job)}.")
 
     # Check if task is test or not
-    if not test:
+    if settings.TEST_MODE:
```
Eshaan7, line 77: should be `if not settings.TEST_MODE`.

Eshaan7: I've no idea how the tests passed successfully....

Issue #534, opened 2021-07-14
[v3] Plugin actions: Kill, retry, healthcheck


Eshaan7:
- The `kill` can be made on top of [`celery.app.control`](https://docs.celeryproject.org/en/stable/reference/celery.app.control.html) by using corresponding task IDs which we store already.
- For `retry`, we have to construct the same arguments again and call `run_plugin` again. Another way could be to have someway to store the `Plugin` object instance and just call the `.start()` method again but I don't think this is doable.
- `healthcheck`: it would be good to have a some sort of basic check to see if that instance/URL is up or not.

**Checklist**:
- [x] 1. Kill (#572)
- [x] 2. Retry (#582)
- [x] 3. Healthchecks (for: connectors + docker based analyzers) (#627 )

sp35:
We know if an analyzer, say docker plugin is configured or not but is it running? (health check)

sp35:
The point behind this was that all my connector calls would fail if the associated instance is not running. For this we are adding a call re-run option. But it would be good to have a some sort of basic check to see if that instance is up or not.

mlodic:
Yep a healthcheck against the associated instances (MISP, etc) can be useful to add

Eshaan7:
> But it would be good to have a some sort of basic check to see if that instance is up or not.

I did not think about it like that. You are definitely right, this could indeed be a great addition. Like a ping-pong that just tells whether the instance/URL is reachable or not.

Note: This could be useful even for the docker analyzers.

Eshaan7:
Tracking in clients as:
- PyIntelOwl (https://github.com/intelowlproject/pyintelowl/issues/111)
- IntelOwl-ng (https://github.com/intelowlproject/IntelOwl-ng/issues/123)

PR #582, opened 2021-07-28
Plugin `retry` functionality (w/ views, tests)

sp35:
# Description
This adds an action `retry` for retrying a failed/killed plugin call.  Arguments are reconstructed and the plugin call starts again in a new celery task. Post this, `PluginReport` is updated with a new report. Job status is updated in the case of analyzers.

## Related issues
#534 (2. Plugin Retry)

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.
 

# Real World Example
image
image


api_app/analyzers_manager/classes.py
```diff
@@ -47,15 +48,31 @@ def init_report_object(self) -> AnalyzerReport:
...
+        else:
+            return AnalyzerReport.objects.create(
+                job_id=self.job_id,
+                analyzer_name=self.analyzer_name,
+                report={},
+                errors=[],
+                status=AnalyzerReport.Statuses.PENDING.name,
+                runtime_configuration=self.kwargs.get("runtime_conf", {}),
+                task_id=self.kwargs["task_id"],
+            )
```
Eshaan7, line 75: Can't we use the `get_or_create` or `get_or_update` method instead ? what do u think ?

api_app/analyzers_manager/views.py
```diff
@@ -70,3 +70,11 @@ def get_object(self, job_id, analyzer_name) -> AnalyzerReport:
     def _post_kill(self, report):
         # clean up job
         analyzers_controller.job_cleanup(report.job)
+
+    def _start_retry(self, report: AnalyzerReport):
+        analyzers_to_execute = [report.analyzer_name]
+        report.job.status = "running"  # update job status to running
```
Eshaan7, line 76: shortway would be to use the `Job.update_status()` method.

api_app/analyzers_manager/classes.py
Eshaan7: The logic duplication of `init_report_object` method in the 2 subclasses would be redundant after #583.

api_app/analyzers_manager/views.py
```diff
@@ -70,3 +70,11 @@ def get_object(self, job_id, analyzer_name) -> AnalyzerReport:
...
+        report.job.save(update_fields=["status"])
+        analyzers_controller.start_analyzers(
+            report.job.id, analyzers_to_execute, report.runtime_configuration
```
Eshaan7, line 79: The `report.runtime_configuration` won't work like this.

The `runtime_configuration` stored in `AbstractReport` stores only the key-value pairs but the `start_analyzers` method expects `runtime_configuration` in the format: `{"ANALYZER_NAME": {"key1": "value1", "key2": "value2"} }`

api_app/analyzers_manager/views.py
Eshaan7: Actually, since we are calling the `start_analyzers`, this is not required.

See here: https://github.com/intelowlproject/IntelOwl/blob/03b51e47269ad4af27e0c3c09338ea8f0ce2ea63/api_app/analyzers_manager/controller.py#L123-L125

api_app/analyzers_manager/classes.py
sp35: anyways I have to update an existing object, `get_or_create` would only be useful if I had to avoid duplicating some object

sp35: well, I haven't checked out on `update_or_create` yet, looks like it should do it.

api_app/connectors_manager/views.py
```diff
@@ -19,14 +20,21 @@ def list(self, request):
...
-    def get_object(self, job_id, connector_name) -> ConnectorReport:
+    def get_object(self, job_id, name) -> ConnectorReport:
         try:
             return self.queryset.get(
                 job_id=job_id,
-                connector_name=connector_name,
+                name=name,
             )
         except ConnectorReport.DoesNotExist:
             raise NotFound()
```
Eshaan7, line 30: The implementation of `get_object` is the same in both subclasses, so it can be moved to `PluginActionViewSet`.

api_app/core/classes.py
```diff
@@ -70,13 +70,33 @@ def after_run(self):
...
 
     @abstractmethod
-    def init_report_object(self) -> AbstractReport:
+    def get_report_class(self):
         """
-        Returns: instance of a subclass of ``AbstractReport``.
-        Called inside `__init__`.
+        Returns Model to be used for *init_report_object*
         """
         raise NotImplementedError()
```
Eshaan7, line 77: I'd prefer to change this to:

```python3
    @property
    @abstractmethod
    def report_model(self):
          raise NotImplementedError()
```

tests/connectors_manager/test_views.py
```diff
@@ -29,16 +28,5 @@ def setUp(self):
...
-                "job_id": _job.id,
-                "connector_name": "MISP",
-                "status": ConnectorReport.Statuses.PENDING.name
-                if status is None
-                else status,
-                "task_id": "4b77bdd6-d05b-442b-92e8-d53de5d7c1a9",
-            }
-        )
-        return _report
+    def get_report_class(self):
```
Eshaan7, line 31: https://github.com/intelowlproject/IntelOwl/pull/582#discussion_r678218117

api_app/core/views.py
```diff
@@ -54,9 +61,41 @@ def kill(self, request, job_id, name):
         # kill celery task
         celery_app.control.revoke(report.task_id, terminate=True)
         # update report
-        report.status = AbstractReport.Statuses.KILLED.name
-        report.save(update_fields=["status"])
+        report.update_status(AbstractReport.Statuses.KILLED.name)
         # execute callback post kill
         self._post_kill(report)
 
         return Response(status=status.HTTP_200_OK)
```
Eshaan7, line 79: rreturn 204 status code.

api_app/core/views.py
```diff
@@ -54,9 +61,41 @@ def kill(self, request, job_id, name):
...
+            AbstractReport.Statuses.KILLED.name,
+        ]:
+            raise ValidationError(
+                {"detail": "Plugin call status should be failed or killed"}
+            )
+
+        # retry with the same arguments
+        self._start_retry(report)
+
+        return Response(status=status.HTTP_200_OK)
```
Eshaan7, line 112: return 204 status code.

api_app/connectors_manager/views.py
```diff
@@ -19,14 +20,21 @@ def list(self, request):
...
 
     def _post_kill(self, report):
         pass
+
+    def _start_retry(self, report: ConnectorReport):
+        connectors_to_execute = [report.connector_name]
+        connectors_controller.start_connectors(
```
Eshaan7, line 31: missing the `runtime_configuration = {report.analyzer_name: report.runtime_configuration}` as is there in the `_start_retry` of `AnalyzerActionViewSet`.

api_app/core/views.py
```diff
@@ -54,9 +61,41 @@ def kill(self, request, job_id, name):
         # kill celery task
         celery_app.control.revoke(report.task_id, terminate=True)
         # update report
-        report.status = AbstractReport.Statuses.KILLED.name
-        report.save(update_fields=["status"])
+        report.update_status(AbstractReport.Statuses.KILLED.name)
         # execute callback post kill
         self._post_kill(report)
```
Eshaan7, line 77: i'd say that this part of the code should be a method named `perform_kill()`   (if you see the DRF code, that's the naming convention they follow). Then we do not require `_post_kill` since we can simply override the method in the subclasses and call `super()`.

Eshaan7: Similar change could be made for `perform_retry` where we can simply call `super()` and then subsequent individual logic.

Eshaan7:
Please mark the PR as "ready for review" when you are done.

Eshaan7:
Thank you for solving the all the small nitpick details. Appreciate it. :D

Issue #11, opened 2020-02-24
OpenCTI connector


mlodic:
IntelOwl was designed to be integrated with other security tools very easily.
In particular, it is a perfect fit for Threat Intelligence Platforms because it can enrich the data available in those platform and can be customized on its own needs.

So, in order to help the community to achieve this result, we should add new connectors for the most used open source TIPs, for example: [OpenCTI](https://github.com/OpenCTI-Platform)

We should study their project and add a relative connector to integrate IntelOwl with OpenCTI. A related issue was already opened: https://github.com/OpenCTI-Platform/connectors/issues/43.
We should leverage the official [IntelOwl client](https://github.com/mlodic/pyintelowl) for this purpose

Shailja1806:
I would like to work on this. plz ,guide me.

mlodic:
This work should be done only after having solved a larger architectural problem in IntelOwl. See #249

PR #602, opened 2021-07-31
Fixes: #11 Add OpenCTI connector

sp35:
# Description
This adds a connector for OpenCTI. It creates a report and an observable from the job data on a connected OpenCTI instance.

- [ ] Docs work to be handled in #562 

Notes
- `requests` was downgraded to `2.25.1` for `pycti==4.5.5`
- IP validation on GUI doesn't allow `IPv6` addresses. 

## Related issues
#11 

## Type of change

# Checklist

- [x] I have added tests 
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

# Real World Example
<details>
      <summary>Click to expand</summary>

</details>

### DIFF
--- a/api_app/connectors_manager/connectors/opencti.py
+++ b/api_app/connectors_manager/connectors/opencti.py
@@ -32,7 +32,7 @@ class OpenCTI(Connector):
         self.__url_name = self._secrets["url_key_name"]
         self.__api_key = self._secrets["api_key_name"]
 
-    def get_observable_type(self):
+    def get_observable_type(self) -> str:
         if self._job.is_sample:
             type = INTELOWL_OPENCTI_TYPE_MAP["file"]
         elif self._job.observable_classification == "hash":
@@ -56,7 +56,7 @@ class OpenCTI(Connector):
 
         return type
 
-    def create_observable(self):
+    def generate_observable_data(self) -> dict:
         observable_data = {"type": self.get_observable_type()}
         if self._job.is_sample:
             observable_data["name"] = self._job.file_name
@@ -71,16 +71,11 @@ class OpenCTI(Connector):
         else:
             observable_data["value"] = self._job.observable_name
 
-        observable = self.opencti_api_client.stix_cyber_observable.create(
-            observableData=observable_data,
-            createdBy=self.organization["id"],
-            objectMarking=self.marking_definition["id"],
-        )
-        return observable
+        return observable_data
 
     def run(self):
         # set up client
-        self.opencti_api_client = OpenCTIApiClient(
+        opencti_api_client = OpenCTIApiClient(
             url=self.__url_name,
             token=self.__api_key,
             ssl_verify=self.ssl_verify,
@@ -88,7 +83,7 @@ class OpenCTI(Connector):
         )
 
         # Create author (if it doesn't exist)
-        self.organization = self.opencti_api_client.identity.create(
+        organization = opencti_api_client.identity.create(
             type="Organization",
             name="IntelOwl",
             description=(
@@ -99,7 +94,7 @@ class OpenCTI(Connector):
             ),
         )
         # Create the marking definition
-        self.marking_definition = self.opencti_api_client.marking_definition.create(
+        marking_definition = opencti_api_client.marking_definition.create(
             definition_type="TLP",
             definition="TLP:%s" % self.tlp["type"].upper(),
             x_opencti_color=self.tlp["color"].upper(),
@@ -107,10 +102,15 @@ class OpenCTI(Connector):
         )
 
         # Create the observable
-        observable = self.create_observable()
+        observable_data = self.generate_observable_data()
+        observable = opencti_api_client.stix_cyber_observable.create(
+            observableData=observable_data,
+            createdBy=organization["id"],
+            objectMarking=marking_definition["id"],
+        )
 
         # Create the report
-        report = self.opencti_api_client.report.create(
+        report = opencti_api_client.report.create(
             name=f"IntelOwl Job-{self.job_id}",
             description=(
                 f"This is IntelOwl's analysis report for Job: {self.job_id}."
@@ -118,29 +118,29 @@ class OpenCTI(Connector):
             ),
             published=self._job.received_request_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
             report_types=["internal-report"],
-            createdBy=self.organization["id"],
-            objectMarking=self.marking_definition["id"],
+            createdBy=organization["id"],
+            objectMarking=marking_definition["id"],
             x_opencti_report_status=2,  # Analyzed
         )
         # Create the external reference
-        external_reference = self.opencti_api_client.external_reference.create(
+        external_reference = opencti_api_client.external_reference.create(
             source_name="IntelOwl Analysis",
             description="View analysis report on the IntelOwl instance",
             url=f"{settings.WEB_CLIENT_URL}/pages/scan/result/{self.job_id}",
         )
         # Add the external reference to the report
-        self.opencti_api_client.stix_domain_object.add_external_reference(
+        opencti_api_client.stix_domain_object.add_external_reference(
             id=report["id"], external_reference_id=external_reference["id"]
         )
 
         # Link Observable and Report
-        self.opencti_api_client.report.add_stix_object_or_stix_relationship(
+        opencti_api_client.report.add_stix_object_or_stix_relationship(
             id=report["id"], stixObjectOrStixRelationshipId=observable["id"]
         )
 
         return {
-            "observable": self.opencti_api_client.stix_cyber_observable.read(
+            "observable": opencti_api_client.stix_cyber_observable.read(
                 id=observable["id"]
             ),
-            "report": self.opencti_api_client.report.read(id=report["id"]),
+            "report": opencti_api_client.report.read(id=report["id"]),
         }
--- a/api_app/connectors_manager/connectors/opencti.py
+++ b/api_app/connectors_manager/connectors/opencti.py
@@ -1,0 +1,146 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+from django.conf import settings
+
+from pycti import OpenCTIApiClient
+
+from api_app import helpers
+from ..classes import Connector
+
+
+INTELOWL_OPENCTI_TYPE_MAP = {
+    "ip": {
+        "v4": "ipv4-addr",
+        "v6": "ipv6-addr",
+    },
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -34,11 +34,12 @@ pysafebrowsing==0.1.1
 PySocks==1.7.1
 python-magic==0.4.22
 quark-engine==21.7.2
-requests==2.26.0
+requests==2.25.1
```
Eshaan7, line 37: any particular reason for this change ?

api_app/connectors_manager/connectors/opencti.py
```diff
@@ -3,8 +3,10 @@
 
 from django.conf import settings
 
-from pycti import OpenCTIApiClient
+from pycti.api.opencti_api_client import File
+import pycti
 
+from tests.mock_utils import patch, if_mock_connections
 from api_app import helpers
 from ..classes import Connector
```
Eshaan7, line 11: use Absolute import.

Eshaan7:
Please rebase this with latest changes from develop-2.

mlodic:
I have opened this in the meanwhile: https://github.com/OpenCTI-Platform/client-python/issues/200

It is bad for us to leverage clients that do not have requirements set properly. This would impact the project badly considering all the integrations that we have

api_app/connectors_manager/connectors/opencti.py
```diff
@@ -0,0 +1,171 @@
...
+            createdBy=organization["id"],
+            objectMarking=marking_definition["id"],
+        )
+
+        # Create the report
+        report = pycti.Report(opencti_instance).create(
+            name=f"IntelOwl Job-{self.job_id}",
+            description=(
+                f"This is IntelOwl's analysis report for Job: {self.job_id}."
+                f" Analyzers Executed: {self._job.analyzers_to_execute}"
```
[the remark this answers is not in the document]
mlodic: If you can do the same to the MIPS connector too, it would be great for the readers

mlodic (changes requested):
thank you for your patience and for your work @sp35

api_app/connectors_manager/connectors/opencti.py
```diff
@@ -0,0 +1,171 @@
...
+    def run(self):
+        # set up client
+        opencti_instance = pycti.OpenCTIApiClient(
+            url=self.__url_name,
+            token=self.__api_key,
+            ssl_verify=self.ssl_verify,
+            proxies=self.proxies,
+        )
+
+        # Create author (if it doesn't exist)
```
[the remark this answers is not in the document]
sp35: I need the uuid of an author for further steps.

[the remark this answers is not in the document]
mlodic: if we save the uuid in a new table of the DB at the time of the first creation? and we try to retrieve it at the time of the analysis? If we do not find it, we can just follow the normal flow.

api_app/connectors_manager/connectors/opencti.py
```diff
@@ -0,0 +1,171 @@
...
+        marking_definition = pycti.MarkingDefinition(opencti_instance).create(
+            definition_type="TLP",
+            definition="TLP:%s" % self.tlp["type"].upper(),
+            x_opencti_color=self.tlp["color"].upper(),
+            x_opencti_order=self.tlp["x_opencti_order"],
+        )
+
+        # Create the observable
+        observable_data = self.generate_observable_data()
+        observable = pycti.StixCyberObservable(opencti_instance, File).create(
```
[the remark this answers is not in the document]
sp35: As long as the type and values are same, it just updates the existing one.

api_app/connectors_manager/connectors/opencti.py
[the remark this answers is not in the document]
Eshaan7: Something like `cached_property` could work. Similar to how we use a session in PyIntelOwl ([reference](https://github.com/intelowlproject/pyintelowl/blob/681096d37690a63c26e133956e7036f04295ec43/pyintelowl/pyintelowl.py#L33-L50)).

api_app/connectors_manager/connectors/opencti.py
```diff
@@ -0,0 +1,171 @@
...
+            name=f"IntelOwl Job-{self.job_id}",
+            description=(
+                f"This is IntelOwl's analysis report for Job: {self.job_id}."
+                f" Analyzers Executed: {self._job.analyzers_to_execute}"
+            ),
+            published=self._job.received_request_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
+            report_types=["internal-report"],
+            createdBy=organization["id"],
+            objectMarking=marking_definition["id"],
+            x_opencti_report_status=2,  # Analyzed
```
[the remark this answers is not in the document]
sp35: Yep

api_app/connectors_manager/connectors/opencti.py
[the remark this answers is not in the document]
Eshaan7:
```
", ".join(self._job.analyzers_to_execute)
```

would be one way.

api_app/connectors_manager/connectors/opencti.py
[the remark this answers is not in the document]
Eshaan7: > As long as the type and values are same, it just updates the existing one.

This should be code comment before line 107.

mlodic:
I have also opened this little PR https://github.com/OpenCTI-Platform/client-python/pull/201

api_app/connectors_manager/connectors/opencti.py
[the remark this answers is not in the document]
sp35: ~updates~ returns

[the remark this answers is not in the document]
sp35: screencapture-143-244-140-149-8080-dashboard-observations-observables-a9439404-82d0-4220-a720-ab5fd7d44470-2021-08-03-21_44_00

api_app/connectors_manager/connectors/opencti.py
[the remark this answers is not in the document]
sp35: `cached_property` won't be useful as it would stay only with the instance and these are already called once per instance.
We need something more persistent.

mlodic (approved):

api_app/connectors_manager/connectors/opencti.py
```diff
@@ -0,0 +1,215 @@
...
+            token=self.__api_key,
+            ssl_verify=self.ssl_verify,
+            proxies=self.proxies,
+        )
+
+        self.opencti_db = {}
+        # get presaved values stored in MEDIA_ROOT
+        if os.path.exists(OPENCTI_DB):
+            with open(OPENCTI_DB, "r") as file:
+                self.opencti_db = json.load(file)
```
Eshaan7, line 127: I do not like this. If we want to persist, we can use django's cache framework. The [django-cache-memoize](https://github.com/peterbe/django-cache-memoize) library is already installed,

Eshaan7: or simply use the `django.core.cache.cache.set(cache_key, function_out, expiration)`

sp35: Thanks, I'll use `cache_memoize` here.

api_app/connectors_manager/connectors/opencti.py
```diff
@@ -0,0 +1,199 @@
...
+    @cache_memoize(timeout=sys.maxsize)
+    def marking_definition_id(self) -> str:  # called only once (cached indefinitely)
+        # Create the marking definition (if not exists)
+        md = pycti.MarkingDefinition(self.opencti_instance).create(
+            definition_type="TLP",
+            definition=f"TLP:{self.tlp['type'].upper()}",
+            x_opencti_color=self.tlp["color"].upper(),
+            x_opencti_order=self.tlp["x_opencti_order"],
+        )
+        return md["id"]
```
Eshaan7, line 107: Caching indefinitely can lead to various bugs that will be super hard to debug. The `pycti.Identity` call depends on the `self.opencti_instance` value so if user were to change the URL, things will end up breaking. I'll go through the pycti docs and refer a better solution.

api_app/connectors_manager/connectors/opencti.py
```diff
@@ -0,0 +1,199 @@
...
+        else:
+            observable_data["value"] = self._job.observable_name
+
+        return observable_data
+
+    @property
+    @cache_memoize(timeout=sys.maxsize)
+    def organization_id(self) -> str:  # called only once (cached indefinitely)
+        # Create author (if not exists)
+        org = pycti.Identity(self.opencti_instance).create(
```
Eshaan7, line 85: If you see the PyCTI docs,

> The Threat-Actor entity will only be created if it doesn’t exists By setting update to True it acts like an upsert and updates fields of an existing Threat-Actor entity.

The same applies to the `pycti.Identity.create` method (they do not document this but I looked at the [code](https://github.com/OpenCTI-Platform/client-python/blob/f75785e4cd6c249eb1db3dc53126cc0ca98370bb/pycti/entities/opencti_identity.py#L8)). So we don't need to have our own caching, we can just provide `update=True` keyword argument.

api_app/connectors_manager/connectors/opencti.py
Eshaan7: Things will also break if they manually change something on the OpenCTI dashboard.

api_app/connectors_manager/connectors/opencti.py
Eshaan7: A way to test if this does work or not would be to run 2 connector calls and see that only one Identity was created.

api_app/connectors_manager/connectors/opencti.py
Eshaan7: Earlier, I thought we were caching only for a particular instance of a connector call else I'd have pointed this out at that moment only.

api_app/connectors_manager/connectors/opencti.py
sp35: If you were concerned that it would create two identities, it wouldn't. I thought we were avoiding making that call after first one.

api_app/connectors_manager/connectors/opencti.py
sp35: Thanks for pointing this out, it slipped out of my mind.

sp35: I'm reverting back to what I did earlier.

api_app/connectors_manager/connectors/opencti.py
sp35: I agree that we do not need our own caching and handling, let's leave it to the API.

PR #616, opened 2021-08-05
Add `drf-flex-fields` pkg and use in `JobSerializer`

Eshaan7:


Eshaan7:
@sp35,

Please use this branch to solve the polling issue in IntelOwl-ng and if any changes are needed in the backend, push them to this branch only.

PR #617, opened 2021-08-05
Fixes #59: `VirusTotalv3` , `behavioral` and `sigma` options, `VirusTotalv3AnalyzerMixin` class

### LINKED ISSUE #59: VirusTotal Sandbox analyzer
We could add another analyzer for VirusTotal that extracts the sandbox reports available for a specific file.

https://developers.virustotal.com/v3.0/reference#file-behaviour-summary

m0mosenpai:
# Description

Updated VirusTotalv3 Analyzer to leverage `/files/{id}/behaviour_summary` and `sigma_analyses/{id}`  to get a summary of behavioral analysis reports and sigma analysis reports in addition to default reports.

## Related issues
#59 

## Type of change

# Checklist

- [X] If I added a new analyzer, I updated the file [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md). If the analyzer provides additional optional configuration, I added the available options here: [Advanced-Usage](./Advanced-Usage.md)
- [X] `Black` gave 0 errors.
- [X] `Flake` gave 0 errors.
 

# Real World Example

image
image


configuration/analyzer_config.json
```diff
@@ -2486,6 +2486,28 @@
...
     }
   },
+  "VirusTotal_v3_Get_Behavioral_Summary": {
+    "type": "file",
+    "python_module": "vt3_sandbox.VirusTotalv3Sandbox",
+    "description": "get a file's behavioral summary from VT's integrated Sandboxes",
+    "disabled": false,
+    "external_service": true,
+    "leaks_info": true,
+    "run_hash": true,
```
mlodic, line 2496: I don't think `run_hash` is necessary, am I right?

api_app/analyzers_manager/observable_analyzers/vt3_sandbox.py
```diff
@@ -0,0 +1,90 @@
...
+            response = requests.get(vt_base + uri, headers=headers)
+            if not response.status_code == 404:
+                response.raise_for_status()
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
+
+        result = response.json()
+
+        # If file hash not found in VirusTotal Database
+        if response.status_code == 404:
```
mlodic, line 79: even if this does make sense if seen with only this perspective, in practice we cannot do that because there is already another analyzer that would perform the "active" scan if the sample is not available in the DB (`VirusTotal_v3_Get_File_And_Scan`). 

So, a user that launches "all analyzers" would ask to VT to perform an active scan 2 times. This is why other similar endpoints (like `VirusTotal_v3_Get_File`) are disabled by default. This is extremely critical when managing the available quota of an expensive service like VT.

I would not force the active scanner option.
I would add different options that can be configured:
* force the active scan if the file was not found (useful if this analyzer is run alone), default=False
* force the polling if we know that this analyzer is executed at the same time of another VT scan analyzer (case of `all analyzers` execution, the most common one), default=True. (without this flag, only a single request is done)

Thoughts?

api_app/analyzers_manager/observable_analyzers/vt3_sandbox.py
```diff
@@ -0,0 +1,90 @@
...
+                patch(
+                    "requests.get",
+                    return_value=MockResponse({}, 200),
+                ),
+            )
+        ]
+        return super()._monkeypatch(patches=patches)
+
+
+def vt_get_behavior_summary(
```
mlodic, line 52: why is this not a method of the previous class?

mlodic (changes requested):

configuration/analyzer_config.json
Eshaan7: I think the confusion arised because In the `files/{id}/behaviour_summary` endpoint as stated in VT's [docs](https://developers.virustotal.com/v3.0/reference#file-all-behaviours-summary), the `id` is the hash to identify a file.

Eshaan7: But this analyzer should be a file analyzer without `run_hash` because we do need the actual _file_ to be able to upload it to VT.

api_app/analyzers_manager/observable_analyzers/vt3_sandbox.py
Eshaan7: > force the active scan if the file was not found (useful if this analyzer is run alone), default=False

You mean similar to the `force_active_scan` config variable in the `VirusTotal_v3_Get_File_And_Scan` analyzer, right ?

Eshaan7: > force the polling if we know that this analyzer is executed at the same time of another VT scan analyzer (case of all analyzers execution, the most common one), default=True. (without this flag, only a single request is done)

Although this can be done easily by checking is the `self._job.run_all_analyzers` is True or not, I was thinking that maybe in that case we just don't conduct the analysis and write like "Please see the VirusTotal_v3_Get_File_And_Scan report instead" because  doesn't the other VT analyzer contain the data of this "behaviour summary" too ?

Eshaan7: A problem with both of these approacesh is that if a user has the other VT analyzer disabled, the analyzer will break. Maybe we should not make an analyzer dependent on another.

mlodic: the other VT analyzer does not contain this data unfortunately

configuration/analyzer_config.json
mlodic: If we remove the chance to upload the file, we could just move this analyzer to an `observable` analyzer and add the related `file` analyzer with `run_hash` option. This would be more clear because otherwise we do not have support for hash analysis for this analyzer.

About this, I am also making some changes to Intezer and MWDB analyzers because they have the same problem. They can be used only as file analyzer and not as observable (hash only) analyzers

m0mosenpai: @Eshaan7 I thought the `id` was the hash of the file and that this endpoint just `GET`s the summary from an already existing file in the database. Else, it's a `404`, in which case we'll have to upload it first.

api_app/analyzers_manager/observable_analyzers/vt3_sandbox.py
m0mosenpai: I think having them not depend on each other would be the best thing as well. If each VT analyzer can have it's own check for sending a file for analysis if required, otherwise just retrieving the result they are designed to, it would be cleaner allbeit a bit redundant.

m0mosenpai: Another thing could be, what if there's a single analyzer that deals with sending and retrieving a report from VT and we can simple retrieve from different endpoints based on the config options. LIke default could be the default report for analysis but a `report_type = behavioral` could retrieve a behavioral report instead. (Similar to how it's done in Triage Scan currently)

api_app/analyzers_manager/observable_analyzers/vt3_sandbox.py
m0mosenpai: There was no particular reason (I'm not sure which case would be better). I thought I would be consistent with the formatting of other `VirusTotal` analyzers.

api_app/analyzers_manager/observable_analyzers/vt3_sandbox.py
mlodic: Yes, your last idea is the better one I think. We could add the behavioral report as "optional" in the already existing VT 3 analyzer

api_app/analyzers_manager/observable_analyzers/vt3_sandbox.py
mlodic: yeah they are out the class there because they are called externally but this is not the case

api_app/analyzers_manager/observable_analyzers/vt3_sandbox.py
Eshaan7: I love that idea! just a boolean config var `include_behavioral_summary`.

m0mosenpai: Behavioral analysis works but I'm having some trouble finding files that match to sigma rules. All the samples I tried didn't have that analysis available to them, so the default report is retrieved in place. It should work otherwise.

Eshaan7: Instead of a string config var like `report_type`, I think it'd be better to have a parameter like: `include_behavioral_summary`, so we will always fetch the relationship/default detail but also include the behavioral data if this param is set to True.

The problem I see with `report_type` is that user cannot request multiple types of information in a single analysis only.

@mlodic, thoughts ?

Eshaan7:
Can you squash all commits into one ? single the yesterday changes are not relevant at all anymore.

api_app/analyzers_manager/observable_analyzers/vt3_sandbox.py
mlodic: I agree with Eshaan.

To test some samples already in VT with sigma rules try 478331245c3ad724ce0880ebd911fd54 or 2e74c9eb3caf855ea2d050b4ece174c8 and let me know if it does find anything

Eshaan7: Cool. Similar to behaviour one would be the `include_sigma_analyses` for sigma.

m0mosenpai: That would be better, yeah. Although the way the analyzer is written rn, it's a bit messy. Maybe after we get the report, just before returning it, we can call on these endpoints if the config variable is set to True and append the data to the rest of the dictionary.

Eshaan7: yes, that's how it should be. I think we can make the default relationship values as a parameter too.

Eshaan7: So basically, 3 parameters:
- `include_relationships`
- `include_behaviour_summary`
- `include_sigma_analyses`

Eshaan7: Actually, the hashes @mlodic provided are md5 but the `sigma_analyses` API works only with sha256 hash. ([docs reference](https://developers.virustotal.com/v3.0/reference#get-sigma-analysis-id))

m0mosenpai:
image
image

@mlodic I used the first hash you shared on the website to find the file's `sha256` (Sigma endpoint only supports `sha256`, hence the `run_hash_type="sha256"`). Hardcoded the hash in that function's argument for the above example and It worked with that so I guess it should be fine.

Eshaan7 (approved):

PR #618, opened 2021-08-05
`Intezer_Get`, `MWDB_Get` analyzers | refactor `Intezer_Scan` | bug fix in quark | adjusts to docker conf

### LINKED ISSUE #586: updates for Intezer integration
We need to do some updates for the Intezer integration:

- [x] They expanded the support for different type of files:

Supported formats:
* Windows Executable Files (PE) –
exe, .dll, .sys – native x86, native x64 and .NET.

* Linux Executable Files (ELF) –
native x86, native x64, ARM32, ARM64

* Document and script files (Office, PDF, Powershell, VBS)

* Android applications (APK)

* Installers

- [x] Also, we do not have support for analysis by hash only. See https://github.com/intezer/analyze-scripts/blob/master/analyze_by_hash.py

- [x] the access token is valid for 10 hours only so we should update how we manage it

- [x] add detailed analysis -> the URLs in the doc do not work

Reference: https://analyze.intezer.com/api/docs/documentation

mlodic:
# Description

closes #586 

## Related issues
#586

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

### DIFF
--- a/api_app/analyzers_manager/file_analyzers/intezer_scan.py
+++ b/api_app/analyzers_manager/file_analyzers/intezer_scan.py
@@ -1,102 +1,72 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
-from django.conf import settings
-import os
-import time
-import requests
-import logging
+from datetime import timedelta
+
+from intezer_sdk import api as intezer_api, errors as intezer_errors
+from intezer_sdk.analysis import Analysis
 
 from api_app.exceptions import AnalyzerRunException
 from api_app.analyzers_manager.classes import FileAnalyzer
-from api_app.helpers import get_now_date_only, get_binary
 
-from tests.mock_utils import (
-    patch,
-    if_mock_connections,
-    MagicMock,
-    MockResponse,
-)
-
-logger = logging.getLogger(__name__)
-
-
-def _get_access_token(api_key):
-    """
-    this should be done just once in a day
-    """
-    base_url = "https://analyze.intezer.com/api/v2-0"
-    response = requests.post(
-        base_url + "/get-access-token", json={"api_key": api_key}
-    )  # lgtm [py/uninitialized-local-variable]
-    response.raise_for_status()
-    response_json = response.json()
-    token = response_json.get("result", "")
-    os.environ["INTEZER_TOKEN"] = token
-    os.environ["INTEZER_TOKEN_DATE"] = get_now_date_only()
-    return token
+from tests.mock_utils import patch, if_mock_connections
 
 
 class IntezerScan(FileAnalyzer):
-    base_url: str = "https://analyze.intezer.com/api/v2-0"
-
     def set_params(self, params):
-        self.__api_key = self._secrets["api_key_name"]
-        # max no. of tries when polling for result
-        self.max_tries = params.get("max_tries", 200)
-        # interval b/w HTTP requests when polling
-        self.poll_distance = 3
+        # soft time limit
+        soft_time_limit = params.get("soft_time_limit", 300)
+        self.timeout = soft_time_limit - 5
+        # interval
+        self.poll_interval = 3
+        # disable_dynamic_unpacking
+        self.disable_dynamic_unpacking = params.get("disable_dynamic_unpacking", False)
+        # disable_static_unpacking
+        self.disable_static_unpacking = params.get("disable_static_unpacking", False)
+        # upload file or not
+        self.upload_file = params.get("upload_file", True)
+        intezer_api.set_global_api(api_key=self._secrets["api_key_name"])
 
     def run(self):
-        intezer_token = os.environ.get("INTEZER_TOKEN", "")
-        intezer_token_date = os.environ.get("INTEZER_TOKEN_DATE", None)
-        today = get_now_date_only()
-        if not intezer_token or intezer_token_date != today:
-            intezer_token = _get_access_token(self.__api_key)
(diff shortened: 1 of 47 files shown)


api_app/analyzers_manager/file_analyzers/mwdb_scan.py
```diff
@@ -58,7 +58,8 @@ def run(self):
         else:
             try:
                 file_info = mwdb.query_file(query)
-            except Exception:
+            except Exception as e:
+                logger.info(e, stack_info=True)
```
Eshaan7, line 62: `logger.error` is called in the `_handle_exception` method which will be called since we are raising `AnalyzerRunException` so this here is redundant.

api_app/analyzers_manager/observable_analyzers/intezer_get.py
```diff
@@ -0,0 +1,61 @@
...
+class IntezerGet(classes.ObservableAnalyzer):
+    base_url: str = get_intezer_base_url()
+
+    def set_params(self, params):
+        self.__api_key = self._secrets["api_key_name"]
+        # intezer access token
+        self.intezer_token = get_access_token(self.__api_key)
+
+    def run(self):
+        return self.__intezer_analysis()
```
Eshaan7, line 27: I'd argue that the `_intezer_analysis` serves no actual purpose and adds an unnecessary overhead to the call stack, the function body can directly be part of `run` method.

docker/Dockerfile_nginx_develop
```diff
@@ -0,0 +1,13 @@
+# Stage 1: Get build artifacts from intelowl-ng
```
Eshaan7, line 1: Instead of a new file, we should use a docker variable `${INTELOWL_NG_DOCKER_TAG}` which can be defined in `docker/.env` file and then in the `test.override.yml` compose file, we just re-define it to `develop`.

api_app/analyzers_manager/file_analyzers/intezer_scan.py
```diff
@@ -33,59 +35,107 @@ def _get_access_token(api_key):
...
+        or (intezer_token_datetime_object < now - timedelta(hours=9))
+    ):
+        intezer_token = _retrieve_access_token_from_intezer(api_key)
+        if not intezer_token:
+            raise AnalyzerRunException("token extraction failed")
+
+    return intezer_token
+
+
+def get_intezer_base_url():
```
Eshaan7, line 63: would be better as a top level constant, `INTEZER_BASE_URL=`.

api_app/analyzers_manager/file_analyzers/intezer_scan.py
```diff
@@ -33,59 +35,107 @@ def _get_access_token(api_key):
...
+    if (
+        not intezer_token
+        or not intezer_token_datetime_object
+        or (intezer_token_datetime_object < now - timedelta(hours=9))
+    ):
+        intezer_token = _retrieve_access_token_from_intezer(api_key)
+        if not intezer_token:
+            raise AnalyzerRunException("token extraction failed")
+
+    return intezer_token
```
Eshaan7, line 60: Why is the Intezer API so complicated ? I really didn't think about this before but the we are writing to the `os.environ` which can be super buggy when used with celery. 

Maybe instead of all this logic we should use their Python SDK (https://github.com/intezer/analyze-python-sdk) ?

It provides a very simple interface:

```python3
from intezer_sdk import consts

index = Index(file_path=<file_path>, 
              index_as=consts.IndexType.MALICIOUS, 
              family_name=<family_name>)
index.send(wait=True)
index_id = index.index_id
```

api_app/analyzers_manager/file_analyzers/intezer_scan.py
mlodic: I'll change this

api_app/analyzers_manager/file_analyzers/mwdb_scan.py
mlodic: ok so I do not like that we raise a so generic Exception. I'll dive deeper to what was the case the previous developer wanted to catch exactly and manage it without a AnalyzerRunException

api_app/analyzers_manager/file_analyzers/intezer_scan.py
mlodic: yeah, the env variables are a concern. Considering that my main goal was to add new functionalities, I added even other complexity. Even if I would always try to avoid the usage of external packages if unecessary, if that sdk works correctly it would really make this analyzer simpler and without problems. I'll test it and update this PR. Thank you for the suggestions

docker/Dockerfile_nginx_develop
mlodic: I tried that but it did not work. I had to use ARG to make it work properly in the FROM. Result is the same

api_app/analyzers_manager/file_analyzers/intezer_scan.py
mlodic: I have no time to end this before my vacation. Either you just get the fix for Quark and I'll finish this later or someone else should finish this.

mlodic: I am in contact with the Intezer team to make official our integration so I just need to end this until the end of the month. No particular hurry otherwise

Eshaan7: No worries, I'll finish the work in this PR.

Issue #612, opened 2021-08-04
Disable connectors before release


sp35:
Connectors by nature are optional.
Task: set `disabled=true` in `connector_config.json`

sp35:
Another possible issue is that users might want to use separate instances for the connectors (connectors: MISP and OpenCTI share the same keys as their analyzers (`MISP_KEY`, `OPENCTI_KEY`, etc)). 

The keys for connectors can be renamed to `CONNECTOR_*`. Then the connectors will be unconfigured at the start so they won't run and we can be kept enabled.

mlodic:
yep, it makes sense. They should not share the keys

PR #622, opened 2021-08-06
Fixes #620, #612: Add separate keys for connectors and disable by default

### LINKED ISSUE #620: Use separate keys in analyzer and connector
Another possible issue is that users might want to use separate instances for the connectors (connectors: MISP and OpenCTI share the same keys as their analyzers (`MISP_KEY`, `OPENCTI_KEY`, etc)). 

The keys for connectors can be renamed to `CONNECTOR_*`. Then the connectors will be unconfigured at the start so they won't run and can be kept enabled.

_Originally posted by @sp35 in https://github.com/intelowlproject/IntelOwl/issues/612#issuecomment-892805524_

sp35:
- [x] #620 
- [x] #612


sp35:
The coverage decreased because the connectors weren't called.

tests/analyzers_manager/__init__.py
```diff
@@ -128,7 +128,7 @@ def test_start_analyzers(self, *args, **kwargs):
                     msg="all `analyzer_reports` status must be `SUCCESS`",
                 )
                 # check connectors status
-                if connectors_stats["all"] > 0 and connectors_stats["running"] == 0:
+                if connectors_stats["all"] >= 0 and connectors_stats["running"] == 0:
```
Eshaan7, line 131: should we include 0 ? that can raise a false positive in case no connectors were executed, no ?

sp35: Hmm, this is also the case when no connectors are present

Eshaan7: but during testing, all connectors in the config should get executed regardless they are marked as disabled or not.

sp35: I don't know why I was thinking that day, if someone removes all the connectors, the tests would start to fail.
Reverting this right away.

PR #627, opened 2021-08-10
Fixes #534: Add healthcheck for docker analyzers and connectors

### LINKED ISSUE #534: [v3] Plugin actions: Kill, retry, healthcheck

sp35:
# Description

- [x] docker based analyzers
- [x] connectors (workaround)
- [x]  boolean in analyzers (`docker_based`)  list for clients
- [x] unit tests 

## Related issues
#534 

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.

### DIFF
--- a/api_app/analyzers_manager/views.py
+++ b/api_app/analyzers_manager/views.py
@@ -1,8 +1,6 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
-from django.utils.module_loading import import_string
-from api_app.analyzers_manager.classes import DockerBasedAnalyzer
 import logging
 
 from rest_framework import generics, status
@@ -89,8 +87,4 @@ class AnalyzerHealthCheckAPI(PluginHealthCheckAPI):
         return config.get_full_import_path()
 
     def perform_healthcheck(self, analyzer_name):
-        klass: DockerBasedAnalyzer = import_string(self.get_cls_path(analyzer_name))
-        # docker analyzers have a common method for health check
-        if not hasattr(klass, "health_check"):
-            raise ValidationError({"detail": "No healthcheck implemented"})
-        return klass.health_check()
+        return analyzers_controller.run_healthcheck()
--- a/api_app/analyzers_manager/views.py
+++ b/api_app/analyzers_manager/views.py
@@ -1,14 +1,16 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
+from django.utils.module_loading import import_string
+from api_app.analyzers_manager.classes import DockerBasedAnalyzer
 import logging
 
 from rest_framework import generics, status
+from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
 from rest_framework import serializers as BaseSerializer
 
-
-from api_app.core.views import PluginActionViewSet
+from api_app.core.views import PluginActionViewSet, PluginHealthCheckAPI
 from .serializers import AnalyzerConfigSerializer
 from . import controller as analyzers_controller
 from .models import AnalyzerReport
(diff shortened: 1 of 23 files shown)


api_app/analyzers_manager/views.py
```diff
@@ -76,3 +78,19 @@ def perform_retry(self, report: AnalyzerReport):
         analyzers_controller.start_analyzers(
             report.job.id, analyzers_to_execute, runtime_configuration
         )
+
+
+class AnalyzerHealthCheckAPI(PluginHealthCheckAPI):
+    def get_cls_path(self, analyzer_name) -> str:
+        analyzer_dataclasses = AnalyzerConfigSerializer.get_as_dataclasses()
```
Eshaan7, line 85: The `get_as_dataclasses` function reads the entire config and then turns each `dict` object to a dataclass. So here there is much waste of computatation.

I think we could define a function, like:

```
@classmethod
AnalyzerConfig.get(cls, analyzer_name: str) -> Union[AnalyzerConfig, None]:
```

basically this would serve as queryset alike for dataclasses, to get an analyzer directly by it's name.

Eshaan7:
Please change the target branch to `develop`.

api_app/core/dataclasses.py
```diff
@@ -94,3 +97,14 @@ def _read_secrets(self, secrets_filter=[]) -> dict:
     @abstractmethod
     def get_full_import_path(self) -> str:
         raise NotImplementedError()
+
+    @classmethod
+    def get(cls, name: str) -> typing.Optional["AbstractConfig"]:
```
Eshaan7, line 102: Although I like that we can use an abstract method like this, but because of how things are intertwined at the moment this will cause circular imports and what not.

api_app/core/dataclasses.py
```diff
@@ -35,6 +36,8 @@ class AbstractConfig:
     secrets: dict
     verification: _SecretsVerification
 
+    serializer_class = AbstractConfigSerializer
```
Eshaan7, line 39: This is causing circular import.

api_app/core/dataclasses.py
Eshaan7: I'd say directly define the `get` function in the subclasses.

Eshaan7: and in the `AbstractConfig`, wrap it w/ the `abstractmethod` decorator and call `raise NotImplemented()`.

api_app/connectors_manager/classes.py
```diff
@@ -51,3 +54,41 @@ def after_run(self):
...
+        url_loc = cls.get_healthcheck_url_loc()
+        url, health_status = None, None
+
+        if url_loc.get("url", None) is not None:
+            url = url_loc["url"]
+        elif url_loc.get("secrets", None) is not None:
+            secret_dict = cc._read_secrets(url_loc["secrets"])
+            url = secret_dict[url_loc["secrets"]]
+        elif url_loc.get("config", None) is not None:
+            url = cc.config[url_loc["config"]]
```
Eshaan7, line 83: I do not really get why this is implemented in such a complex manner.

Eshaan7: As I mentioned yesterday, I think passing the `cc: ConnectorConfig` argument is redundant. We can just hardcode the connector name inside the classes and use `ConnectorConfig.get(name)` and then `read_secrets["url_key_name"]`

Eshaan7: We can just hardcode: `url = cc._read_secrets["url_key_name"]`

PR #633, opened 2021-08-13
Fixes #5: Added ClamAV Analyzer

### LINKED ISSUE #5: ClamAV integration
We should add a new analyzer to allow users to analyze files with ClamAV.

https://github.com/Cisco-Talos/clamav-devel.

It's important to:

- [ ] if the license is permissive, add public-available rules by default  

- [ ] give the chance to customize the rules used to scan a file with ClamAV, in particular to run its own rules

m0mosenpai:
# Checklist

- [X] Barebones changes + configuration
- [ ] Set up `clamd` daemon + configuration
- [ ] Test changes
- [ ] Update docs

## Related issues
#5 

## Type of change

# Checklist

- [X] `Black` gave 0 errors.
- [X] `Flake` gave 0 errors.


mlodic:
Please pull from `develop`, I have updated the psycopg package https://github.com/intelowlproject/IntelOwl/commit/c550d7fda52912ad569104ac78b98271b82d716f

api_app/analyzers_manager/file_analyzers/clamav.py
```diff
@@ -0,0 +1,27 @@
...
+# See the file 'LICENSE' for copying permission.
+
+from api_app.analyzers_manager.classes import FileAnalyzer, DockerBasedAnalyzer
+
+
+class ClamAV(FileAnalyzer, DockerBasedAnalyzer):
+    name: str = "ClamAV"
+    url: str = "http://static_analyzers:4002/clamav"
+    # interval between http request polling
+    poll_distance: int = 10
```
Eshaan7, line 11: `poll_distance` should be 3 seconds, `max_tries` should be 5 and `timeout` should be 15.

integrations/static_analyzers/entrypoint.sh
```diff
@@ -3,9 +3,10 @@ mkdir -p ${LOG_PATH}
 touch ${LOG_PATH}/gunicorn_access.log ${LOG_PATH}/gunicorn_errors.log
 chown -R static_analyzers-user:static_analyzers-user ${LOG_PATH}
 su static_analyzers-user -s /bin/bash
+freshclam && clamd
```
Eshaan7, line 6:
```suggestion
# start clam service itself and the updater in background as daemon
freshclam -d &
clamd &

```
so they both run in the background.

m0mosenpai: I think `clamd` starts in the background only by default and `-d` option for `freshclam` makes it start as a daemon. So,
```bash
freshclam -d
clamd
```
should be enough I think

m0mosenpai:
**When file is OK**:
uninfected

**Infected File**:
infected

PR #641, opened 2021-08-20
Fixes #638: Add tlp support for analyzers and connectors

### LINKED ISSUE #638: Support for TLP, both for analyzers and connectors
> That could/should condition which analyzers can be invoked and if/how an IOC and relatives analyses can be shared with external parts. Example: tag an IP as TLP:RED => only run internal/fully controlled analyzers, do not submit the IP to external platforms, and keep the result as well as the IOC for internal users only

At the moment there are 3 options for visibility and privacy:
image

* `leaks_info` flag: disable analyzers that could impact privacy
* `external_service` flag: disable analyzers that use external services
* `private` flag: limit view permissions to my groups

I guess that, even if this does make sense, it is too much complicated and could be simplified by leveraging the Traffic Light Protocol.
Also the TLP could be used to check if we can share an analysis or not with the configured connectors. ATM the connectors can only be active or not and there is no a granular way to choose which analysis we want to share (@sp35 could you confirm this?).
We could add an option on each connector where the user can choose which is the TLP minimum value of the related analysis that can be accepted for sharing (default would be the highest, TLP:white)

Also, the corresponding relation with the flags could be something like:
* TLP: RED. all the 3 previous flag set. 
* TLP: AMBER: `leaks_info` and `private` flag set.
* TLP: GREEN: `leaks_info` flag set.
* TLP: WHITE: no flag set.
(obviously this should be explicitly declared somewhere in the GUI)

Plus, this could be also used to change the configuration at runtime of the analyzers like VT, MWDB, Intezer, etc that provide the option `upload_file` that would be changed based on the TLP set. (upload only if TLP:white is set for instance). Anyway this could be changed even without the TLP rework.

sp35:
# Description
Deprecated the use of booleans `force_privacy`, `disable_external_analyzers`, `private` with standard TLP.
Used these `TLPs` to manage connector calls.

## Related issues
Fixes #638 

## Type of change

# Checklist

- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.


Eshaan7:
I think the tests are failing because of the #643.

Eshaan7 (approved):

Eshaan7:
LGTM. Waiting for a review from Matteo.

mlodic (approved):
that's good but this should not close completely the issue because the last part of the issue was not managed in this PR:

> Plus, this could be also used to change the configuration at runtime of the analyzers like VT, MWDB, Intezer, etc that provide the option upload_file that would be changed based on the TLP set. (upload only if TLP:white is set for instance). Anyway this could be changed even without the TLP rework.

PR #655, opened 2021-08-24
fix: use django-celery-results as result backend

Eshaan7:
We can't use the `rpc` result backend because it does not work w/ chords.

Error:
> raised unexpected: NotImplementedError('The "rpc" result backend does not support chords!\n\nNote that a group chained with a task is also upgraded to be a chord,\nas this pattern requires synchronization.\n\nResult backends that supports chords: Redis, Database, Memcached, and more.')

So, I try here to use the postgreSQL database as backend using the [django-celery-results](https://github.com/celery/django-celery-results) backend library.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 boto3==1.18.7
 celery==5.1.2
+django-celery-results
 checkdmarc==4.4.1
 darksearch==2.1.2
 dataclasses==0.6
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,5 +1,6 @@
 boto3==1.18.7
 celery==5.1.2
+django-celery-results
```
mlodic, line 3: misses version. should be 2.2.0

PR #674, opened 2021-08-31
bugfix: add back "django.contrib.staticfiles", other adjusts

Eshaan7:


intel_owl/settings.py
```diff
@@ -92,6 +92,7 @@ def retrieve(self, file, analyzer):
     "django.contrib.contenttypes",
     "django.contrib.sessions",
     "django.contrib.messages",
+    "django.contrib.staticfiles",
```
mlodic, line 95: @Eshaan7 is this really necessary? I think Simone deleted it because those static files are just used in the readme

Eshaan7: The django admin breaks without this

Eshaan7: Because it needs css/JS static files to work

mlodic: ah yeah right thanks to the explanation

PR #780, opened 2021-11-25
fixed Quark Engine

mlodic:


mlodic:
merging because error is not related to Quark that, on the contrary, performed good

PR #786, opened 2021-12-06
fix to VT quotas management

mlodic:


mlodic:
merging because error is related to another analyzer

eshaan7:
> merging because error is related to another analyzer

yeah im fixing it.

Issue #798, opened 2021-12-13
add Malpedia Yara rules


mlodic:
similar to #782 

 https://github.com/malpedia/signator-rules/tree/main/rules

The problem is that this repo contains a lot of rules, so we should think about how to properly manage them.
We should solve the issue #741 before.

This problem has a high priority

CypherpunkSamurai:
Hello sir, would you mind if i work on this issue please? :)

CypherpunkSamurai:
Hello Sir, as there are issues regarding mass importing of Yara rules I'll make a draft PR for now and add the Malpedia rules.

mlodic:
done with v3.3.1

PR #815, opened 2021-12-25
Add `Yara_Scan_MALPEDIA` rules. Solve #798

CypherpunkSamurai:
> yeah, if we won't solve that resources issue now, we could merge the PR anyway with one condition: set the analyzer as `disabled` by default and writing down something on the analyzer description about the risk of running it

Sir, I'm closing this PR because this PR is unlinked.


eshaan7:
Why is this a draft ?

mlodic:
yeah, if we won't solve that resources issue now, we could merge the PR anyway with one condition: set the analyzer as `disabled` by default and writing down something on the analyzer description about the risk of running it

CypherpunkSamurai:
I'm closing this PR because this PR is unlinked. I'll update changes in a new PR. Sorry for any Inconvienience.

PR #829, opened 2022-01-02
Fixes #505: Added `Onionscan` analyzer using tor_analyzers docker container

### LINKED ISSUE #505: [New analyzer] - onionscan
GitHub: https://github.com/s-rah/onionscan

```bash
onionscan --jsonReport notarealhiddenservice.onion
```

There are 2 ways to integrate this:
a) as a docker based analyzer
b) compile and store as binary which can executed with subprocess

I'd prefer option `b`, because docker based analyzer should be made only as the last resort.

CypherpunkSamurai:
# Description

`Onionscan` module and configs were added along with `linux-amd64` binary located in `/configurations/bundled/onionscan`. Usage docs were updated and tests were patched.

**Todo:**
- [x] Update `legal_notice.md`
- [x] Provide raw json output sample.
    * [protonmail_output.json.txt](https://github.com/intelowlproject/IntelOwl/files/7921530/protonmail_output.json.txt)

- [ ] ~Update Advanced Usage docs for `verbose` and `torProxyAddress` params.~ (not required)

## Related issues
#505 

## Type of change

# Checklist

- [X] A new analyzer or connector was added, in which case:
    - [X] [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md) file was updated.
    - [X] If the analyzer/connector requires mocked testing, `_monkeypatch()` was used in it's class to apply the necessary decorators.
  
### Important Rules

# Real World Example

- [X] **TODO**

IntelOwl_onionscan1
IntelOwl_onionscan2

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/onionscan.py
+++ b/api_app/analyzers_manager/observable_analyzers/onionscan.py
@@ -6,7 +6,7 @@ from api_app.analyzers_manager.classes import DockerBasedAnalyzer, ObservableAna
 
 class Onionscan(ObservableAnalyzer, DockerBasedAnalyzer):
     name: str = "Onionscan"
-    url: str = "http://tor_analyzers:4004/onionscan"
+    url: str = "http://tor_analyzers:4001/onionscan"
     # http request polling max number of tries
     max_tries: int = 60
     # interval between http request polling (in seconds)
--- a/api_app/analyzers_manager/observable_analyzers/onionscan.py
+++ b/api_app/analyzers_manager/observable_analyzers/onionscan.py
@@ -8,7 +8,7 @@ class Onionscan(ObservableAnalyzer, DockerBasedAnalyzer):
     name: str = "Onionscan"
     url: str = "http://tor_analyzers:4004/onionscan"
     # http request polling max number of tries
-    max_tries: int = 48
+    max_tries: int = 60
     # interval between http request polling (in seconds)
     poll_distance: int = 10
     verbose = True
--- a/api_app/analyzers_manager/observable_analyzers/onionscan.py
+++ b/api_app/analyzers_manager/observable_analyzers/onionscan.py
@@ -2,48 +2,34 @@
 # See the file 'LICENSE' for copying permission.
 
 from api_app.analyzers_manager.classes import DockerBasedAnalyzer, ObservableAnalyzer
-from tests.mock_utils import patch
 
 
 class Onionscan(ObservableAnalyzer, DockerBasedAnalyzer):
     name: str = "Onionscan"
     url: str = "http://tor_analyzers:4004/onionscan"
     # http request polling max number of tries
-    max_tries: int = 1000
+    max_tries: int = 48
     # interval between http request polling (in seconds)
-    poll_distance: int = 30
+    poll_distance: int = 10
+    verbose = True
+    tor_proxy_address = None
 
     def set_params(self, params):
-        self.args = self._onionscan_args_builder(params)
-
-    @staticmethod
-    def _onionscan_args_builder(params):
-        verbose = params.get("verbose", True)
-        tor_proxy_address = params.get("torProxyAddress", None)
-        # make request arguments
-        args = []
-        if verbose:
-            args.append("-verbose")
-        if tor_proxy_address:
-            args.extend(["-torProxyAddress", tor_proxy_address])
-        return args
+        self.verbose = params.get("verbose", True)
+        self.tor_proxy_address = params.get("torProxyAddress", None)
 
     def run(self):
+        # make request params
+        args = []
+        if self.verbose:
+            args.append("-verbose")
+        if self.tor_proxy_address:
+            args.extend(["-torProxyAddress", self.tor_proxy_address])
         # make request data
-        self.args.extend(["-jsonReport", self.observable_name])
+        args.extend(["-jsonReport", self.observable_name])
 
         req_data = {
-            "args": self.args,
+            "args": args,
         }
 
         return self._docker_run(req_data=req_data, req_files=None)
-
-    @classmethod
-    def _monkeypatch(cls, *_) -> None:
-        patches = [
-            patch(
-                "DockerBasedAnalyzer._docker_run",
-                return_value="{}",
-            )
-        ]
-        return super()._monkeypatch(patches=patches)
--- a/api_app/analyzers_manager/observable_analyzers/onionscan.py
+++ b/api_app/analyzers_manager/observable_analyzers/onionscan.py
@@ -1,0 +1,49 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+from api_app.analyzers_manager.classes import DockerBasedAnalyzer, ObservableAnalyzer
+from tests.mock_utils import patch
+
+
+class Onionscan(ObservableAnalyzer, DockerBasedAnalyzer):
+    name: str = "Onionscan"
+    url: str = "http://tor_analyzers:4004/onionscan"
+    # http request polling max number of tries
+    max_tries: int = 1000
+    # interval between http request polling (in seconds)
+    poll_distance: int = 30
+
+    def set_params(self, params):
+        self.args = self._onionscan_args_builder(params)
+
+    @staticmethod
+    def _onionscan_args_builder(params):
+        verbose = params.get("verbose", True)
+        tor_proxy_address = params.get("torProxyAddress", None)
+        # make request arguments
+        args = []
+        if verbose:
+            args.append("-verbose")
+        if tor_proxy_address:
+            args.extend(["-torProxyAddress", tor_proxy_address])
+        return args
+
+    def run(self):
+        # make request data
+        self.args.extend(["-jsonReport", self.observable_name])
+
+        req_data = {
+            "args": self.args,
+        }
+
+        return self._docker_run(req_data=req_data, req_files=None)
+
+    @classmethod
+    def _monkeypatch(cls, *_) -> None:
+        patches = [
+            patch(
+                "DockerBasedAnalyzer._docker_run",
+                return_value="{}",
+            )
+        ]
+        return super()._monkeypatch(patches=patches)
--- a/integrations/tor_analyzers/Dockerfile
+++ b/integrations/tor_analyzers/Dockerfile
@@ -1,0 +1,41 @@
+FROM python:3.8-slim
+
+ENV PROJECT_PATH /opt/deploy
+ENV LOG_PATH /var/log/intel_owl/tor_analyzers
+ENV USER tor-user
+
+# Add a new low-privileged user
+RUN useradd -r -s /sbin/nologin ${USER}
+
+# update and install packages
+RUN DEBIAN_FRONTEND=noninteractive apt-get update -qq \
+    && apt-get install --no-install-recommends -y git build-essential gcc pandoc curl tor \
+    && rm -rf /var/lib/apt/lists/*
+
+# Place to bind a mount point to for scratch pad work
+WORKDIR ${PROJECT_PATH}
+RUN mkdir input/ \
+    && chown -R ${USER}:${USER} input/
+
+# Cleanup
+RUN apt-get remove --purge -y git gcc \
+    && apt-get clean \
+    && apt-get autoclean \
+    && apt-get autoremove -y \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /usr/share/doc/* /usr/share/man/* > /dev/null 2>&1
+
+# 2. Build Flask REST API
+WORKDIR ${PROJECT_PATH}/tor-flask
+COPY app.py requirements.txt entrypoint.sh ./
+
+RUN pip3 install -r requirements.txt --no-cache-dir \
+    && chown -R ${USER}:${USER} . \
+    && chmod +x entrypoint.sh
+
+# 3. Copy Bundled tools
+COPY bundled bundled
+RUN chmod 755 -R bundled
+
+# Serve Flask application using gunicorn
+EXPOSE 4004
+ENTRYPOINT ["./entrypoint.sh"]
--- a/integrations/tor_analyzers/Dockerfile
+++ b/integrations/tor_analyzers/Dockerfile
@@ -37,5 +37,5 @@ COPY bundled bundled
 RUN chmod 755 -R bundled
 
 # Serve Flask application using gunicorn
-EXPOSE 4004
+EXPOSE 4001
 ENTRYPOINT ["./entrypoint.sh"]
--- a/integrations/tor_analyzers/app.py
+++ b/integrations/tor_analyzers/app.py
@@ -41,8 +41,5 @@ app.config.update(CONFIG)
 executor = Executor(app)
 shell2http = Shell2HTTP(app, executor)
 
-# binary paths
-onionscan_binary = "./bundled/onionscan"
-
 # with this, we can make http calls to the endpoint: /onionscan
-shell2http.register_command(endpoint="onionscan", command_name=onionscan_binary)
+shell2http.register_command(endpoint="onionscan", command_name="./bundled/onionscan")
--- a/integrations/tor_analyzers/app.py
+++ b/integrations/tor_analyzers/app.py
@@ -16,7 +16,7 @@ from flask_shell2http import Shell2HTTP
 logger = logging.getLogger("flask_shell2http")
 # logger config
 formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
-log_level = os.getenv("LOG_LEVEL", logging.INFO)
+log_level = os.getenv("LOG_LEVEL", "20")
 log_path = os.getenv("LOG_PATH", "/var/log/intel_owl/tor_analyzers")
 # create new file handlers, files are created if doesn't already exists
 fh = logging.FileHandler(f"{log_path}/tor_analyzers.log")
--- a/integrations/tor_analyzers/app.py
+++ b/integrations/tor_analyzers/app.py
@@ -1,0 +1,48 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+import logging
+
+# system imports
+import os
+
+# web imports
+from flask import Flask
+from flask_executor import Executor
+from flask_shell2http import Shell2HTTP
+
+# Logging configuration
+# get flask-shell2http logger instance
+logger = logging.getLogger("flask_shell2http")
+# logger config
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+log_level = os.getenv("LOG_LEVEL", logging.INFO)
+log_path = os.getenv("LOG_PATH", "/var/log/intel_owl/tor_analyzers")
+# create new file handlers, files are created if doesn't already exists
+fh = logging.FileHandler(f"{log_path}/tor_analyzers.log")
+fh.setFormatter(formatter)
+fh.setLevel(log_level)
+fh_err = logging.FileHandler(f"{log_path}/tor_analyzers_errors.log")
+fh_err.setFormatter(formatter)
+fh_err.setLevel(logging.ERROR)
+# add the handlers to the logger
+logger.addHandler(fh)
+logger.addHandler(fh_err)
+logger.setLevel(log_level)
+
+# Globals
+app = Flask(__name__)
+# Config
+CONFIG = {
+    "SECRET_KEY": __import__("secrets").token_hex(16),
+}
+app.config.update(CONFIG)
+
+executor = Executor(app)
+shell2http = Shell2HTTP(app, executor)
+
+# binary paths
+onionscan_binary = "./bundled/onionscan"
+
+# with this, we can make http calls to the endpoint: /onionscan
+shell2http.register_command(endpoint="onionscan", command_name=onionscan_binary)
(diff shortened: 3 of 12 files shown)


api_app/analyzers_manager/observable_analyzers/onionscan.py
```diff
@@ -0,0 +1,82 @@
...
+        self._verbose = params.get("verbose", True)
+        self._tor_proxy_address = params.get("torProxyAddress", None)
+
+    def run(self):
+        """
+        Run Onionscan against target onion domain
+        """
+        # Check for onionscan binary in path/pwd.
+        if which("onionscan"):
+            self.onionscan_binary = "onionscan"
```
eshaan7, line 40: is this really required ? we are not installing onionscan so this will never be true.

requirements.txt
```diff
@@ -30,7 +30,7 @@ peepdf-fork==0.4.3
 pdfid==1.0.4
 pefile-fork==2020.12.16
 psycopg2-binary==2.9.1
-pycti==5.1.2a0
+pycti==5.1.3
```
eshaan7, line 33: pls revert these 2 changes.

docker/Dockerfile
```diff
@@ -40,9 +40,14 @@ RUN touch ${LOG_PATH}/django/api_app.log ${LOG_PATH}/django/api_app_errors.log \
...
     && rm -rf /root/.local
 
+
+# install bundled tools
+COPY configuration/bundled /opt/deploy/bundled
+
 # download yara rules and quark engine rules
 RUN api_app/analyzers_manager/repo_downloader.sh
 
+
```
eshaan7, line 50: redundant empty line.

api_app/analyzers_manager/observable_analyzers/onionscan.py
CypherpunkSamurai: Yeah, that's true. As we are using a docker container and scripts I think its better to remove it.

requirements.txt
CypherpunkSamurai: Ok. I was actually trying to undo those changes in my commit (that's why the force push). I added that because it wasn't building on my machine.

Will be removed sir :) 👍🏼

api_app/analyzers_manager/repo_downloader.sh
```diff
@@ -7,6 +7,9 @@
 # Example: you may want to add a new repository. Add the clone here
 # Example: you may want to remove some of the rules available in the downloaded repositories. Remove them here.
 
+# Install pre-bundled Onionscan (linux-amd64)
+chmod 755 /opt/deploy/bundled/onionscan
```
mlodic, line 11: why this line is in this file?

docker/Dockerfile
```diff
@@ -40,6 +41,10 @@ RUN touch ${LOG_PATH}/django/api_app.log ${LOG_PATH}/django/api_app_errors.log \
     # this is cause stringstifer creates this directory during the build and cause celery to crash
     && rm -rf /root/.local
 
+
+# install bundled tools
+COPY configuration/bundled /opt/deploy/bundled
```
mlodic, line 46: is this really necessary? this file is copied anyway beforehand

docker/entrypoint_uwsgi.sh
```diff
@@ -20,6 +20,9 @@ echo "DEBUG: " $DEBUG
 echo "DJANGO_TEST_SERVER: " $DJANGO_TEST_SERVER
 echo "------------------------------"
 
+# start required daemons
+tor --DataDirectory $(mktemp -d) &
```
mlodic, line 24: honestly I do not like this to happen by default.

Tor should be started only if the user wants to use the OnionScan analyzer. Most of the time the user won't. 

This is also the reason why, in my opinion, this analyzer cannot be enabled by default. We should force the user to choose to activate it if he wants, and let him understand what he is doing. Open a port to Tor cannot just be taken so lightly.

docs/source/Usage.md
```diff
@@ -204,6 +204,7 @@ The following is the list of the available analyzers you can run out-of-the-box.
 * `AbuseIPDB`: check if an ip was reported on [AbuseIPDB](https://www.abuseipdb.com/)
 * `Fortiguard`: scan an observable with the [Fortiguard URL Analyzer](https://www.fortiguard.com/webfilter)
 * `TorProject`: check if an IP is a Tor Exit Node
+* `OnionScan`: check a Tor domain for misconfigurations and info leaks.
```
mlodic, line 207: For the reasons mentioned beforehand, I would like to have OnionScan as an analyzer that is optional.

I think one good way to do that is to transform it as a docker analyzer (see https://intelowl.readthedocs.io/en/latest/Advanced-Usage.html#optional-analyzers). Tor and the onionscan binary should be integrated there. They are not necessary in the default image.

However, that would mean to add another optional container. That is not a good thing because, if a user wants to enable all the analyzers, they would already need to run too many containers.

So I think this is the reason why we should integrate OnionScan requirements in an already existing container. We just need to choose which one.
@eshaan7 thoughts?

api_app/analyzers_manager/observable_analyzers/onionscan.py
```diff
@@ -0,0 +1,107 @@
...
+        check if tor bootstrap is complete
+        """
+        try:
+            url = "https://check.torproject.org/api/ip"
+            tor_proxy = {
+                "http": "socks5h://127.0.0.1:9050",
+                "https": "socks5h://127.0.0.1:9050",
+            }
+            r = requests.get(url, proxies=tor_proxy)
+            IsTor = json.loads(r.text)["IsTor"]
```
mlodic, line 98: better call varaible names without capital letters

api_app/analyzers_manager/observable_analyzers/onionscan.py
```diff
@@ -0,0 +1,107 @@
...
+            return IsTor
```
mlodic, line 99: just a single `return` for this function please...for better readibility

api_app/analyzers_manager/observable_analyzers/onionscan.py
```diff
@@ -0,0 +1,107 @@
...
+        except Exception as e:
+            logger.info("onionscan:check_tor error: \n%s", str(e))
```
mlodic, line 101: use the most recent f-format, not the old %s please

mlodic (changes requested):

api_app/analyzers_manager/repo_downloader.sh
CypherpunkSamurai: Sir, I added that to make sure the bundled tools have correct permissions. Should I move it to the Dockerfile instead?

docker/entrypoint_uwsgi.sh
CypherpunkSamurai: Right, open tor ports can serve as dangerous. Should we use the daemon from a subprocess in onionscan instead?

docs/source/Usage.md
CypherpunkSamurai: I think you're right. Having a separate container for all tor enabled tools along with the daemon would help keep the main app separate from the noise that happens to pass through tor.

Adding the daemon to a separate container would be an organized way to handle things, but hard on the CPU resources. Is it suitable if we make installer scripts for separate tools inside the container, and just pass them as command line args during install?

api_app/analyzers_manager/observable_analyzers/onionscan.py
CypherpunkSamurai: Ok, Sir.

CypherpunkSamurai:
I'm pushing changes, for now, to keep track of progress (also because I'm using an online vscode instance :P). The onionscan module happens to run ok but takes a lot of time, sometimes timeouts celery.

I agree with Mr. Matteo's suggestions about dockerizing tor enabled tools, so let's discuss it first :)

api_app/analyzers_manager/repo_downloader.sh
mlodic: yes, I think that line is unrelated to the that .sh file

docs/source/Usage.md
eshaan7: > Having a separate container for all tor enabled tools along with the daemon would help keep the main app separate from the noise that happens to pass through tor.

Yep, a new service in `integrations/` for TOR enabled tools makes sense. Similar to how we have one for static analyzers, other for APK analyzers.

eshaan7: I didn't know that onionscan requires an active tor daemon running otherwise I'd have suggested this from the very beginning. That's a mistake on my part for not properly reading through their README. :man_facepalming:

mlodic: As I mentioned, we have already a solution for this cases and I would not like to add other "tricks" to solve this issue.

The solution is: Docker-based analyzers (all the ones that are in the `integration` folder, the ones mentioned in the previous link I shared). 

There is also documentation about how to integrate them: https://intelowl.readthedocs.io/en/latest/Contribute.html#integrating-a-docker-based-analyzer

The concept is that you are separating "problematic" dependencies from the base image and add them in optional images. This is very useful because analyzers should not influence the base image because they are not mandatory to the overall application: most of the users would like to use just some of them and not everyone (they are more than 100).

So, as said, we already have some of this optional images and maybe they are already too many to be run together.
So, instead of creating a new docker-based integration, we could just add OnionScan dependencies and service in an already existing one (like `box-js`). The point is that this case someway is not related to the others. I mean, there are no other analyzers that needs Tor right now. So it would make sense to create a new docker-based analyzer for Tor-related services. If we do that (this is easier, you can do that), then we can try to reduce the amount of the overall number of docker-based analyzers by grouping some ot them together (`box-js` and `apk_analyzers` for instance)

eshaan7: So for this new `tor_analyzers`, you'd need to do 4 things:

- Create new directory inside `integrations/` named `tor_analyzers`. (Refer to `apk_analyzers` directory, you need the same files and same directory structure). This also means onionscan binary should now be kept inside `integrations/tor_analyzers/` as it will be executed from this new container using the flask-shell2http package.
- Convert the `Onionscan` class to inherit from `DockerBasedAnalyzer` class. (Refer to other docker-based observable analyzers such as thug_url.py, etc.)
- Add entry for the `tor_analyzers` integration in the `start.py`.

mlodic: created an issue for this too: https://github.com/intelowlproject/IntelOwl/issues/842

eshaan7: @CypherpunkSamurai so we discussed within ourselves and the final decision is to have a new container for tor enabled services (we will combine/compress other analyzers together for optimization)

So for this new `tor_analyzers`, you'd need to do 4 things:

- Create new directory inside `integrations/` named `tor_analyzers`. (Refer to `apk_analyzers` directory, you need the same files and same directory structure). This also means onionscan binary should now be kept inside `integrations/tor_analyzers/` as it will be executed from this new container using the flask-shell2http package.
- Modify the `Onionscan` class to inherit from `DockerBasedAnalyzer` class. (Refer to other docker-based observable analyzers such as thug_url.py, etc.)
- Add entry for the `tor_analyzers` integration in the `start.py` script.
- Edit `Advanced-Usage.md` add new row in the table under "Optional Analyzers".

CypherpunkSamurai: Ok Sir, On It! 😄

CypherpunkSamurai:
Ok, As the discussions seem to give a green flag to a new category of analyzers I'm overwriting current progress with a fresh new pull from `develop` to clean things up :)

CypherpunkSamurai:
Hello Mr. @mlodic and Mr. @eshaan7, can you please suggest to me how to patch `DockerBasedAnalyzers`?

I have tried the following but these ones aren't working as mock points
* `DockerBasedAnalyzer._docker_run`
* `DockerBasedAnalyzer.__enter__.return_value._docker_run`
* `DockerBasedAnalyzer.return_value._docker_run`
* `self._docker_run`

api_app/analyzers_manager/observable_analyzers/onionscan.py
```diff
@@ -0,0 +1,49 @@
...
+        self.args.extend(["-jsonReport", self.observable_name])
+
+        req_data = {
+            "args": self.args,
+        }
+
+        return self._docker_run(req_data=req_data, req_files=None)
+
+    @classmethod
+    def _monkeypatch(cls, *_) -> None:
```
eshaan7, line 42: Don't require monkeypatch in DockerBasedAnalyzer.

api_app/analyzers_manager/observable_analyzers/onionscan.py
```diff
@@ -0,0 +1,49 @@
...
+    def _onionscan_args_builder(params):
+        verbose = params.get("verbose", True)
+        tor_proxy_address = params.get("torProxyAddress", None)
+        # make request arguments
+        args = []
+        if verbose:
+            args.append("-verbose")
+        if tor_proxy_address:
+            args.extend(["-torProxyAddress", tor_proxy_address])
+        return args
```
eshaan7, line 29: No need for this function. Lines 21-22 should be part of `set_params` and line 23-28 part of `run`.

api_app/analyzers_manager/observable_analyzers/onionscan.py
```diff
@@ -0,0 +1,49 @@
...
+from tests.mock_utils import patch
+
+
+class Onionscan(ObservableAnalyzer, DockerBasedAnalyzer):
+    name: str = "Onionscan"
+    url: str = "http://tor_analyzers:4004/onionscan"
+    # http request polling max number of tries
+    max_tries: int = 1000
+    # interval between http request polling (in seconds)
+    poll_distance: int = 30
```
eshaan7, line 14: In your experience with using onionscan, did u notice it taking a lot of time to return result ? I ask because you've made max_tries 1000 and poll_distance 30 so that's 30,000 seconds or 8 hours :smiling_face_with_tear:

eshaan7: I'd ask you to run 2-3 scans and notice the average time it takes onionscan to return result.

eshaan7: Ideally, the values should be max_tries = 30, poll_distance = 10. So that is 300 seconds or 5 minutes. Which means the `soft_time_limit` should be made ~330.

api_app/analyzers_manager/observable_analyzers/onionscan.py
CypherpunkSamurai: Ok Sir

api_app/analyzers_manager/observable_analyzers/onionscan.py
CypherpunkSamurai: Ok Sir.

api_app/analyzers_manager/observable_analyzers/onionscan.py
CypherpunkSamurai: > In your experience with using onionscan, did u notice it taking a lot of time to return result ? I ask because you've made max_tries 1000 and poll_distance 30 so that's 30,000 seconds or 8 hours smiling_face_with_tear

Yikes, I added that for testing, seems I forgot to remove it.

> I'd ask you to run 2-3 scans and notice the average time it takes onionscan to return result.

> Ideally, the values should be max_tries = 30, poll_distance = 10. So that is 300 seconds or 5 minutes. Which means the `soft_time_limit` should be made ~330.

I have noticed it takes more than 10 minutes, I'll take an average and try if it works out.

CypherpunkSamurai: I just checked It took more than an hour, and celery just timed it out.

I'm leaving 3 consecutive ones running on a ssh server for a day and I'll return the results tomorrow.

CypherpunkSamurai: So, it looks like onionscan has a problem with v3 addresses, I checked out its issues and someone had a patch for it (onionscan#162) I added the patch and refactored the source, and it looks like its working. I'm now running tests on the ssh shell and will take an average for the timeouts :)

eshaan7: > I checked out its issues and someone had a patch for it (onionscan#162) I added the patch and refactored the source, and it looks like its working.

Really appreciate you going the extra mile with the effort.

So I guess we can make the soft_time_limit and the max_tries x poll_distance to be about 15 minutes.

CypherpunkSamurai: Sure Sir 👍

CypherpunkSamurai:
Sir(s), the DeepSource check seems to be linked to the `develop` branch. Please review my PR for any issues :)

integrations/tor_analyzers/app.py
```diff
@@ -0,0 +1,48 @@
...
+app.config.update(CONFIG)
+
+executor = Executor(app)
+shell2http = Shell2HTTP(app, executor)
+
+# binary paths
+onionscan_binary = "./bundled/onionscan"
+
+# with this, we can make http calls to the endpoint: /onionscan
+shell2http.register_command(endpoint="onionscan", command_name=onionscan_binary)
```
eshaan7, line 48: nitpick: directly assign `command_name="./bundled/onionscan"`.

eshaan7:
Hey, good work! Just one thing that you did not attach a screenshot of the GUI showing that it works.

CypherpunkSamurai:
> Hey, good work! Just one thing that you did not attach a screenshot of the GUI showing that it works.

Oh. My Apologies sir, I got a little sick and was taking time off. I promise I'll attach a screenshot within a day or 2. :P

eshaan7:
> Oh. My Apologies sir, I got a little sick and was taking time off. I promise I'll attach a screenshot within a day or 2. :P

That's not a problem at all. Take your time and be safe!

integrations/tor_analyzers/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ENV PROJECT_PATH /opt/deploy
+ENV LOG_PATH /var/log/intel_owl/tor_analyzers
+ENV USER tor-user
+
+# Add a new low-privileged user
+RUN useradd -r -s /sbin/nologin ${USER}
+
+# update and install packages
+RUN DEBIAN_FRONTEND=noninteractive apt-get update -qq \
+    && apt-get install --no-install-recommends -y git build-essential gcc pandoc curl tor \
```
eshaan7, line 12: I see we are installing `tor` here which makes sense but where/when is the tor daemon being launched ?

CypherpunkSamurai: We launch the tor daemon from the entry point script, before the python api starts.

integrations/tor_analyzers/app.py
CypherpunkSamurai: Ok Sir.

CypherpunkSamurai:
Hello Mr @eshaan7, I noticed onionscan tests we taking too long and celery timed out a few times. Turns out onionscan has bugs that don't return proper exit codes (onionscan keeps running after errors/scan completed). I have patched the bug and updated the binary.

I will be uploading screenshots soon.

Sir, I've noticed that having `verbose` set to true by default will cause the json output to get messed up, are you sure we should enable verbose by default?

integrations/tor_analyzers/Dockerfile
eshaan7: Oh! I missed that file. Thanks for the clarification.

CypherpunkSamurai:
Hello Mr @mlodic and Mr @CypherpunkSamurai. This PR is already completed, Please review this PR :D

api_app/analyzers_manager/observable_analyzers/onionscan.py
```diff
@@ -0,0 +1,35 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+from api_app.analyzers_manager.classes import DockerBasedAnalyzer, ObservableAnalyzer
+
+
+class Onionscan(ObservableAnalyzer, DockerBasedAnalyzer):
+    name: str = "Onionscan"
+    url: str = "http://tor_analyzers:4004/onionscan"
```
mlodic, line 9: 4004 is already used by APK_analyzers. Not a technical problem because it works but for clarity I would prefer this to have a different port from the other docker integrations. `4001` is free.

mlodic (changes requested):
Another change required is:
* update dependabot.yml

api_app/analyzers_manager/observable_analyzers/onionscan.py
CypherpunkSamurai: Sure Sir!

CypherpunkSamurai:
> Another change required is:
> 
> * update dependabot.yml

Hello Mr @mlodic, Please check.

.gitpod.yml
```diff
@@ -0,0 +1,10 @@
+# This configuration file was automatically generated by Gitpod.
+# Please adjust to your needs (see https://www.gitpod.io/docs/config-gitpod-file)
+# and commit this file to your remote git repository to share the goodness with others.
+
+tasks:
+  - init: pip install -r requirements.txt
+vscode:
+  extensions:
+    - ms-python.python
+
```
mlodic, line 10: can you please explain this addition?

Is this related to my request? Because I just asked to update the dependabot.yml. If you check that file, we added all the integrations directory so we need to add this new one too

CypherpunkSamurai: Ok Sir, I've removed it.

mlodic: yeah but still not changed the `dependabot.yml` :P

CypherpunkSamurai: I've updated it sir. Please check.

mlodic (approved):

Issue #722, opened 2021-10-05
CAPE sandbox analyzer


mlodic:
https://www.capesandbox.com

0x0elliot:
Hey! Can I take up this issue? I am free now and I wanted to contribute back again.

mlodic:
We are not sure that they still activate new accounts...

0x0elliot:
The only option we have is to wait.

kevoreilly:
We do activate new accounts when a manual request is made via Twiter - we cannot automate the activation as we have suffered so much abuse from anonymous/bogus accounts who only wish to scrape. This has caused denial of service for all users, I can only apologise for this but it is a necessary evil.

0x0elliot:
I will try making a manual request on Twitter then. Thanks kev!

0x0elliot:
I haven't been able to figure out where the API authentication token for https://www.capesandbox.com/ is? I looked through the documentation and couldn't find anything myself. I have gotten my account verified for context. Anyone else try their hand here and see if they can help?

mlodic:
Hi @kevoreilly , really appreciated! I'll ask now on Twitter too so I can help with the integration!

0x0elliot:
Matteo, you can use my account too! I don't plan on reusing it anyway.

mlodic:
https://capev2.readthedocs.io/en/latest/usage/api.html?highlight=api

You just need to call `curl -d "username=<USER>&password=<PASSWD>" https://capesandbox.com/apiv2/api-token-auth/` with your credentials and you'll get the token to use. 

Then the documentation and the examples of the API are [here](https://capesandbox.com/apiv2/)

0x0elliot:
Thanks!

0x0elliot:
@kevoreilly hey Kevin!

I noticed that whenever I was uploading a file using the API, it seemed stuck at `pending` while when I uploaded it through the GUI at https://www.capesandbox.com/analysis it moved from `pending` to `running` quicker.

Is there something regarding this that you're aware of? I have pretty much finished this one.

kevoreilly:
Hmm the only thing I can think of is that there have recently been issues (now should be fixed) that caused jobs to get stuck in pending - although it shouldn't be different for api to web ui... Is this something you have noticed recently on capesandbox.com?

0x0elliot:
[...]

PR #792, opened 2021-12-08
Fixes #440: new `Virushee_UploadFile` file analyzer

uzaxirr:
# Description
Integrated [Virusshee File Upload](https://api.virushee.com/#operation/upload) Analyzer 
Please include a summary of the change.

## Related issues
Fixes #440 (1/2 bullet from checklist)

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue).
- [x] New feature (non-breaking change which adds functionality).

# Checklist

- [x] The pull request is for the branch develop
- [x] A new analyzer or connector was added, in which case:
    - [x] [Usage](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Usage.md) file was updated.
    - [ ] [Advanced-Usage](./Advanced-Usage.md) was updated (in case the analyzer/connector provides additional optional configuration).
    - [x] Secrets were added in [env_file_app_template](https://github.com/intelowlproject/IntelOwl/blob/master/docker/env_file_app_template), [env_file_app_ci](https://github.com/certego/IntelOwl/blob/master/docker/env_file_app_ci) and in the [Installation](./Installation.md) docs, if necessary.
    - [x] If the analyzer/connector requires mocked testing, `_monkeypatch()` was used in it's class to apply the necessary decorators.
    - [x] If a File analyzer was added, it's name was explicitly defined in [test_file_scripts.py](https://github.com/intelowlproject/IntelOwl/blob/master/tests/analyzers_manager/test_file_scripts.py) (not required for Observable Analyzers).
- [ ] If external libraries/packages with restrictive licenses were used, they were added in the [Legal Notice](https://github.com/certego/IntelOwl/blob/master/.github/legal_notice.md) section.
- [x] The tests gave 0 errors.
- [x] `Black` gave 0 errors.
- [x] `Flake` gave 0 errors.
- [ ] The commits were squashed into a single one (optional, they will be squashed anyway by the maintainer)
  
# Real World Example

Screenshot from 2021-12-08 11-22-10

Screenshot from 2021-12-08 11-22-18

PR #836, opened 2022-01-07
Solves #722. Added CAPE Sandbox analyzer

0x0elliot:
# Description

Added CAPE Sandbox analyzer.

## Related issues

#722 

## Type of change

# Checklist

- [x] A new analyzer or connector was added, in which case:
    - [x] Secrets were added in [env_file_app_template](https://github.com/intelowlproject/IntelOwl/blob/master/docker/env_file_app_template), [env_file_app_ci](https://github.com/certego/IntelOwl/blob/master/docker/env_file_app_ci) and in the [Installation](./Installation.md) docs, if necessary.
    - [x] If the analyzer/connector requires mocked testing, `_monkeypatch()` was used in it's class to apply the necessary decorators.
    - [x] If a File analyzer was added, it's name was explicitly defined in [test_file_scripts.py](https://github.com/intelowlproject/IntelOwl/blob/master/tests/analyzers_manager/test_file_scripts.py) (not required for Observable Analyzers).

# Screenshot of the test run:

#### (Note that the API in use has been disabled for now)

image


0x0elliot:
I have to still run the necessary pre-commit tests. Till then I have made a draft PR for this issue.

0x0elliot:
I think I am not missing anything that is further required. The description for the analyzer might require changing. Please review.

configuration/analyzer_config.json
```diff
@@ -258,6 +258,34 @@
     "secrets": {},
     "params": {}
   },
+  "CapeSandbox" : {
+    "type" : "file",
+    "python_module" : "cape_sandbox.CAPEsandbox",
+    "description" : "Automatic scan of suspicious files using CapeSandbox API",
+    "disabled" : false,
+    "external_service" : false,
```
mlodic, line 266: even if this could be self-managed, I would set it `True` by default

docker/Dockerfile
```diff
@@ -41,7 +41,7 @@ RUN touch ${LOG_PATH}/django/api_app.log ${LOG_PATH}/django/api_app_errors.log \
     && rm -rf /root/.local
 
 # download yara rules and quark engine rules
-RUN api_app/analyzers_manager/repo_downloader.sh
+#RUN api_app/analyzers_manager/repo_downloader.sh
```
mlodic, line 44: reminder of removal

docs/source/Usage.md
```diff
@@ -168,6 +168,7 @@ The following is the list of the available analyzers you can run out-of-the-box.
 * `PEframe_Scan`: Perform static analysis on Portable Executable malware and malicious MS Office documents with [PeFrame](https://github.com/guelfoweb/peframe)
 * `Cymru_Hash_Registry_Get_File`: Check if a particular file is known to be malware by [Team Cymru](https://team-cymru.com/community-services/mhr/)
 * `Thug_HTML_Info`: Perform hybrid dynamic/static analysis on a HTML file using [Thug low-interaction honeyclient](https://thug-honeyclient.readthedocs.io/)
+* `cape_sandbox`: [CAPESandbox](https://capesandbox.com) automatically scans suspicious files using the CapeSandbox API.
```
mlodic, line 171: the name is different from the one in the `analyzer_config.json`. We could explicitly say that this analyzer can be used for private instances too.

tests/analyzers_manager/test_file_scripts.py
```diff
@@ -29,6 +29,7 @@ def get_params(cls):
                 "UnpacMe_EXE_Unpacker",
                 "PEframe_Scan",
                 "Capa_Info",
+                "CapeSandbox"
```
mlodic: comma is missing 😅 . Without it, the 2 strings are merged, causing havoc :P

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+from tests.mock_utils import MockResponse, if_mock_connections, patch
+
+
+class CAPEsandbox(FileAnalyzer):
+    def set_params(self, params):
+        self.__token = self._secrets["token"]
+        self.__vm_name = params.get("VM_NAME", "win7x64_8")
+
+    def run(self):
+        API_URL: str = "https://www.capesandbox.com/apiv2/tasks/create/file/"
```
mlodic, line 15: can we add the chance to customize the URL? in this way we can use this for private intances too

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+            "Authorization": f"Token {self.__token}",
+        }
+
+        files = {
+            "file": (self.filename, self.read_file_bytes()),
+        }
+
+        data = {"machine": self.__vm_name}
+
+        response = requests.post(API_URL, headers=headers, files=files, data=data)
```
mlodic, line 26: I think we should add a `response.raise_for_status()` and raise AnalyzerRunException in that case. There are a lot of other analyzers using this if you check

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+        response_json = response.json()
+        if not response_json["error"]:
```
mlodic, line 28: care, this key could not exist. Access it safely with `get`

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+            to_respond = {}
+            to_respond["result_url"] = response_json["url"]
```
mlodic, line 30: same for this key

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+            task_id = response_json["data"]["task_ids"][0]
+            status_api = "https://www.capesandbox.com/apiv2/tasks/status/" + str(
+                task_id
+            )
+            while True:
```
mlodic, line 35: we should control the polling. There are a lot of other analyzers who perform this kind of polling. Please go look into them. 

plus, `while True` should never be used because a simple error could bring infinite loops

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+                r = requests.get(status_api, headers=headers)
+                responded_json = r.json()
+                error = responded_json["error"]
+                data = responded_json["data"]
+                if (not error) and (data == "reported" or data == "completed"):
```
mlodic, line 40: parenthesis not required for the first condition. Second condition should use `in`

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+                    report_url = (
+                        "https://www.capesandbox.com/apiv2/tasks/get/report/"
+                        + str(task_id)
+                        + "/json"
+                    )
+                    final_request = requests.get(report_url, headers=headers)
+                    final_json = final_request.json()
+                    if final_json["error"]:
```
mlodic, line 48: key must be accessed safely

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,69 @@
...
+                        to_respond["response"] = final_json
+                        return to_respond
+                elif error:
+                    return responded_json
```
mlodic, line 52: this function should have single `return` for better readibility and maintenance

mlodic (changes requested):
thanks for the PR! the build failed for black. Do you have an example of a test run data to share please?

0x0elliot:
It seems like there are lots of changes to make! Getting to it.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
eshaan7: ^makes sense. @0x0elliot u can define the URL in the `docker/.env_template` and then define a variable in the `secrets`. see cuckoo analyzers for reference.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
eshaan7: Won't blame Aditya, since this is something I recommended in the other PR. https://github.com/intelowlproject/IntelOwl/pull/792#discussion_r779122902

my idea was that we can keep polling until celery's SoftTimeLimit exception is raised.

mlodic: makes sense but I think it is generally bad practice because we are assuming that another implicit mechanism would solve the eventual problem. I would say better safe and put a maximum number of possible loops

0x0elliot: Understood. Is there a particular analyzer you can redirect me to which does polling?

0x0elliot: Found it! https://github.com/intelowlproject/IntelOwl/blob/5998d1cbc0fec5ed4c72cc2fa3e052b0e0c05261/api_app/analyzers_manager/observable_analyzers/pulsedive.py#L77

configuration/analyzer_config.json
0x0elliot: Noted

docs/source/Usage.md
0x0elliot: Noted.

tests/analyzers_manager/test_file_scripts.py
0x0elliot: Noted and fixed.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: Noted

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: Why `in` matteo?

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: Noted and taken care of.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: i will be instead raising the error returned here.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: Noted.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: i did it the way cuckoo did it.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: As in, It always returns the whole string "reported" or "completed" accordingly. I usually use `in` only when i need to check if one string co-insides with another. I will definitely add it as well but I wanted to understand why did you feel the need for it?

0x0elliot: ahh i see what you meant! https://deepsource.io/gh/intelowlproject/IntelOwl/run/55667e73-a3d2-4e5c-94ee-b8d92e1f4da6/python/PYL-R1714

0x0elliot:
Ran the pre-commit before pushing. Everything must be good now!
Screenshot 2022-01-09 at 4 47 08 PM

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
mlodic: yeah coming late to your questions but I see you solved your questions by yourself. :P That's appreciated.

mlodic:
thanks, if you need to have a review, please re-ask it to us. In this way we can better filter and manage that request from normal comments. 

There is a still an error in the build if you check the logs of the action:

```
>>> Failed analyzers: [('CapeSandbox', ['401 Client Error: Unauthorized for url: https://www.capesandbox.com/apiv2/tasks/create/file/'])] 
>>> Failed connectors: []
```
I think some more mockups are required to work it properly

0x0elliot:
Hey Matteo. I noticed that as well in the tests but that is simply failing because it requires authorization. Is there something I can do to specify that it would fail till then in the testing script?

docker/env_file_app_ci
```diff
@@ -88,6 +89,7 @@ TEST_MD5=446c5fbb11b9ce058450555c1c27153c
 TEST_GENERIC=email@example.com
 
 # URLs
+CAPESANDBOX_URL=https://www.capesandbox.com
```
mlodic, line 92: please set a fake url here. This file is used by the Github Action

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        files = {
+            "file": (self.filename, self.read_file_bytes()),
+        }
+
+        data = {"machine": self.__vm_name}
+
+        try:
+            response = requests.post(
```
mlodic, line 31: you mocked only the `get` and not the `post` method

mlodic (changes requested):

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: ahh right. didn't notice that.

docker/env_file_app_ci
0x0elliot: will do!

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: I might need a minute to implement proper mock tests. Will request a review when I am done.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class CAPEsandbox(FileAnalyzer):
+    max_tries: int = 25
+    to_respond = {}
+
+    def set_params(self, params):
+        self.__token = self._secrets["token"]
+        self.__vm_name = params.get("VM_NAME", "win7x64_8")
+        self.BASE_URL = self._secrets.get("url_key_name", "https://www.capesandbox.com")
```
eshaan7, line 16: rename to `self.__base_url`.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def run(self):
+        API_URL: str = self.BASE_URL + "/apiv2/tasks/create/file/"
```
eshaan7, line 19: rename to `api_url`.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+        self.headers = {
```
eshaan7, line 20: create a `self.session = requests.Session(headers={..})` inside `set_params` method and use this session object for making requests.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+            "file": (self.filename, self.read_file_bytes()),
+        }
+
+        data = {"machine": self.__vm_name}
+
+        try:
+            response = requests.post(
+                API_URL, headers=self.headers, files=files, data=data
+            )
+            response_json = response.json()
```
eshaan7, line 34: this statement should come _after_ the `response.raise_for_status`.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+            response.raise_for_status()
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
+
+        # since not None == True as well. This is a more precise way.
+        if response_json.get("error") is False:
+            self.to_respond["result_url"] = response_json.get("url")
+            self.task_id = response_json.get("data").get("task_ids")[0]
+            self.status_api = self.BASE_URL + "/apiv2/tasks/status/" + str(self.task_id)
```
eshaan7, line 43: pass the `task_id` as an argument to `__poll_for_result` and build the `status_api` URL inside that function.

We should avoid setting a variable to `self` as much as possible.

configuration/analyzer_config.json
```diff
@@ -258,6 +258,39 @@
...
     "params": {}
   },
+  "CapeSandbox" : {
+    "type" : "file",
+    "python_module" : "cape_sandbox.CAPEsandbox",
+    "description" : "Automatic scan of suspicious files using CapeSandbox API",
+    "disabled" : false,
+    "external_service" : true,
+    "leaks_info" : true,
+    "docker_based" : false,
```
eshaan7, line 268: the default is `false` so better to remove this from here.

configuration/analyzer_config.json
```diff
@@ -258,6 +258,39 @@
...
+    "config" : {
+      "soft_time_limit" : 500,
+      "queue" : "long"
+    },
+    "secrets" : {
+      "token" : {
```
eshaan7, line 274: rename `token` to `api_key_name` because that's the naming convention followed by other analyzers.

configuration/analyzer_config.json
```diff
@@ -258,6 +258,39 @@
...
+        "env_var_key" : "CAPESANDBOX_KEY",
+        "description" : "",
+        "required" : true
+      },
+    "url_key_name": {
```
eshaan7, line 279: Install a JSON formatter in your IDE, the indentation here is messed up.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+    def __poll_for_result(
+        self,
+    ):
+        for i in range(self.max_tries):
+            try:
+                r = requests.get(self.status_api, headers=self.headers)
+                r.raise_for_status()
+            except requests.RequestException as e:
+                if r.status_code == 429:
+                    time.sleep(30)
```
eshaan7, line 88: set this to a class variable `poll_distance`. Same as u have done with `max_tries.`

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def __poll_for_result(
+        self,
+    ):
+        for i in range(self.max_tries):
+            try:
+                r = requests.get(self.status_api, headers=self.headers)
+                r.raise_for_status()
+            except requests.RequestException as e:
+                if r.status_code == 429:
```
eshaan7, line 87: I think a better logic here would be:

```python3
if r.status_code == 429:
  time.sleep(self.poll_distance)
else:
  r.raise_for_status()
```

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+        API_URL: str = self.BASE_URL + "/apiv2/tasks/create/file/"
+        self.headers = {
+            "Authorization": f"Token {self.__token}",
+        }
+
+        files = {
+            "file": (self.filename, self.read_file_bytes()),
+        }
+
+        data = {"machine": self.__vm_name}
```
eshaan7, line 28: since these are used only once, can just pass them directly to the `requests.post` function instead of defining explicitly before.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+                API_URL, headers=self.headers, files=files, data=data
+            )
+            response_json = response.json()
+            response.raise_for_status()
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
+
+        # since not None == True as well. This is a more precise way.
+        if response_json.get("error") is False:
+            self.to_respond["result_url"] = response_json.get("url")
```
eshaan7, line 41: I don't like the use of the `self.to_respond`. There should be better segregation of responsibilities.

As u can see in most other analyzers, we maintain a `result` dictionary and the return value from `__poll_for_result` function and others are updated in `result` at the very end just before `return result` statement in `run` function.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+            data = responded_json.get("data")
+            if not error and (data in ("reported", "completed")):
+                report_url = (
+                    self.BASE_URL
+                    + "/apiv2/tasks/get/report/"
+                    + str(self.task_id)
+                    + "/json"
+                )
+                try:
+                    final_request = requests.get(report_url, headers=self.headers)
```
eshaan7, line 103: missing `raise_for_status`.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+            response = requests.post(
+                API_URL, headers=self.headers, files=files, data=data
+            )
+            response_json = response.json()
+            response.raise_for_status()
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
+
+        # since not None == True as well. This is a more precise way.
+        if response_json.get("error") is False:
```
eshaan7, line 40: here u are checking "error" key but on line 48 you are checking "errors". I assume it's a typo ? or does their API send both in response ?

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,145 @@
...
+
+            try:
+                final_request = requests.get(report_url, headers=self.headers)
+            except requests.RequestException as e:
+                raise AnalyzerRunException(e)
+
+            final_json = final_request.json()
+            self.to_respond["response"] = final_json
+
+            return self.to_respond
```
eshaan7, line 75: this whole part should be extracted to smaller functions. 

- Line 51 - 60 to something like `__search_by_md5`.
- and isn't Line 61 - 75 to similar to `__poll_for_result` ?

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
eshaan7: Same applies for all the different `_url` variables before making request. You can pass them directly to the function. It reduces the number of code lines and makes the code more self-explanatory.

eshaan7:
A lot of changes required but it's okay! You can do them on your own time and then re-request a review from me. The basic principles to follow when writing the analyzers are:
- The flow of control in the `run` method should be readable/easy to understand. To achieve this, there should be only one (never more than 2) `return` statements from the `run` method and if possible all `requests` statement should be extracted out to private methods inside the class (like u did with `__poll_for_result`).
- Try to not set too many variables to `self` object especially inside the `run` method. That's what the `set_params` method is for.
- Always call `.raise_for_status()` after making HTTP request and only call `.json()` after the try/catch.
- Try to not use `elif/else` statements. Meaning: a cleaner alternative `if data: return elif error: return else: return..` is something like 

```python3
if not error:
   do_something()
return result
```

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: oh right! This is a better way to proceed forward with multiple requests.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: It sends it in both responses. That's why I am using it constantly.

0x0elliot:
I will work on these! Thanks for the review.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: Took care of the first part for now. Would look into the second part after I have a stable working draft ready.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: Implemented it this very way.

configuration/analyzer_config.json
0x0elliot: Gotcha

configuration/analyzer_config.json
0x0elliot: Done this for `docker_based`

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,162 @@
...
+                except requests.RequestException as e:
+                    raise AnalyzerRunException(e)
+
+                final_json = final_request.json()
+
+                if final_json.get("error"):
+                    response_ = final_json
+                    return response_
+
+            elif error:
```
mlodic, line 130: it is always better to first manage clauses _positive_ conditions (without `not`) for better clarity

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,162 @@
...
+                raise AnalyzerRunException(error)
+
+            elif data == "running":
+                time.sleep(30)
```
mlodic, line 134: this number should not be hardcoded but a configurable variable. See other similar analyzers (when we used `poll_distance`)

mlodic (changes requested):
there are almost no logs in all the analyzer. This makes impossible to debug the application on production. Please set logs in critical parts of the analyzer

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: In that case, I will just check for any errors before and raise them earlier.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,162 @@
...
+            to_respond["result_url"] = response_json.get("url")
+            task_id = response_json.get("data").get("task_ids")[0]
+            result = self.__poll_for_result(task_id=task_id)
+            to_respond["response"] = result
+            return to_respond
+
+        elif (
+            list(response_json.get("errors")[0].values())[0]
+            == "Not unique, as unique option set on submit or in conf/web.conf"
+        ):
```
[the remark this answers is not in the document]
0x0elliot:
```
{'error': True, 'error_value': 'Error adding task to database', 'errors': [{'60684641.png': 'Not unique, as unique option set on submit or in conf/web.conf'}]}
```

This is what `response_json` looked like for my test image which I had already uploaded before. I am adding the necessary comments which explain everything nicely.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,162 @@
...
+        db_search_url = self.__base_url + "/apiv2/tasks/search/md5/" + self.md5
+
+        try:
+            q = self.__session.get(db_search_url)
+            q.raise_for_status()
+
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
+
+        status_id = str(q.json().get("data")[0].get("id"))
```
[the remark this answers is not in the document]
0x0elliot: > Plus, there is no check that the status_id is an integer as expected.

I don't exactly get it Matteo. Whatever comes out of this  `str(q.json().get("data")[0].get("id"))` is being converted into a `string` by the `str()` function. So why should I check?

> too complex this line + does not handle the case that the data key is not a populated list 

I will try breaking this line down into simpler lines in that case.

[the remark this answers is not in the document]
mlodic: yeah first point my mistake

0x0elliot:
It seems like this should be ready. Please review once and let me know. @mlodic

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+                files={
+                    "file": (self.filename, self.read_file_bytes()),
+                },
+                data={"machine": self.__vm_name},
+            )
+            response.raise_for_status()
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
+
+        logger.info(f"[INFO] (Job: {self.job_id} -> Uploaded the file to CAPESandbox")
```
mlodic, line 42: there is no need to add the level of the log in the message. This is already done by the formatter.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+        self.__session = requests.Session()
+        self.__session.headers = {
+            "Authorization": f"Token {self.__token}",
+        }
+
+    def run(self):
+        api_url: str = self.__base_url + "/apiv2/tasks/create/file/"
+        to_respond = {}
+
```
mlodic, line 29: another log here too please?

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+        logger.info(f"[INFO] (Job: {self.job_id} -> Uploaded the file to CAPESandbox")
+        response_json = response.json()
+
+        if response_json.get("error") is False:
+            to_respond["result_url"] = response_json.get("url")
+            task_id = response_json.get("data").get("task_ids")[0]
+            result = self.__poll_for_result(task_id=task_id)
+            to_respond["response"] = result
+            return to_respond
```
mlodic, line 50: you fixed below the `return` but you could also do it here

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+            )
+            response.raise_for_status()
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
+
+        logger.info(f"[INFO] (Job: {self.job_id} -> Uploaded the file to CAPESandbox")
+        response_json = response.json()
+
+        if response_json.get("error") is False:
+            to_respond["result_url"] = response_json.get("url")
```
mlodic, line 46: another log here too please?

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+                        f"Sleeping for {self.poll_distance} seconds."
+                    )
+                    time.sleep(self.poll_distance)
+                else:
+                    r.raise_for_status()
+            except requests.RequestException as e:
+                if i == self.max_tries - 1:
+                    raise AnalyzerRunException(e)
+                    # ^ Unhandled exception at the last trial raised.
+                logger.info(
```
mlodic, line 151: I do not understand this log. I think it should be before the previous check and should not say that there is an "unhandled exception"

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+                    logger.info(
+                        f"[INFO] (Job: {self.job_id} -> "
+                        "Rate limited by CAPESandbox API. "
+                        f"Sleeping for {self.poll_distance} seconds."
+                    )
+                    time.sleep(self.poll_distance)
+                else:
+                    r.raise_for_status()
+            except requests.RequestException as e:
+                if i == self.max_tries - 1:
```
mlodic, line 148: honestly I do not like such kind of inline calculation. This could better be handled by a bool flag that is checked at the end of the loop

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+    ) -> dict:
+        status_api = self.__base_url + "/apiv2/tasks/status/" + str(task_id)
+        for i in range(self.max_tries):
+            logger.info(
+                f"[POLLING] (Job: {self.job_id} -> "
+                f"CAPEsandbox __poll_for_result #{i + 1}/{self.max_tries}"
+            )
+            try:
+                r = self.__session.get(status_api)
+                if r.status_code == 429:
```
mlodic, line 138: a question: so either you sleep or you raise an exception? So where is the "normal and right" case with the 200 status code?

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+                    final_request = self.__session.get(
+                        report_url,
+                    )
+                    final_request.raise_for_status()
+                except requests.RequestException as e:
+                    raise AnalyzerRunException(e)
+
+                final_json = final_request.json()
+                response_ = final_json
+                return response_
```
mlodic, line 186: again I don't like that it is not clear which are the possible returns in this function. Please just use a single return at the end of the function that manages all cases

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+
+            elif data in ("pending", "running", "processing"):
+                if i == self.max_tries - 1:
```
mlodic, line 189: no inline calculation please, use variables

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
mlodic: plus, why there is a sleep here and then you do not continue the loop? I mean, the logic is really difficult to follow in this for clause. You also do a sleep later. Please manage those sleeps with a single custom exception that, when is triggered, it performs the sleep and continues the loop

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+            q.raise_for_status()
+
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
+
+        data_list = q.json().get("data")
+        if data_list:
+            status_id_int = data_list[0].get("id")
+            status_id = str(status_id_int)
+            return status_id
```
mlodic, line 120: just a single `return` at the end of the code also here please

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,232 @@
...
+                    """
+                        At times, What I have noticed is that
+                        CAPESandbox likes to leave your submission pending.
+                        It has a proper pending queue.
+
+                        Usually the status changes like this:
+                        pending -> running -> processing -> completed -> reported
+                    """
+                logger.info(
+                    f"[INFO] (Job: {self.job_id} ->"
```
mlodic, line 200: if you put the log in this case, you should put it also in the previous one. 

You need to imagine that, with the logs, it is possible to completely understand the flow of the code. If there are possible paths without logs, it is not acceptable and it would create confusion to the person who is debugging the code

mlodic (changes requested):

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
mlodic: "a single return at the end of the **function**.

The point is that if you put multiple `return`, it is really difficult to quickly understand the possible outputs of  this function.

On the contrary, if you just use a single return at the end of the function with a variable, you can just follow the variable through the code.

Also, you made a mix between variables. I would have expect that the `to_respond` would have been the return value but then you overwrite the JSON response `response_json = to_respond`. This is all confusing.

You should just use a variable that you leverage as a single return value and then assign to it the correspondent value based on the conditions.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: Oh I now remember why I added it there. You recommended me to check only if there was an exception happening and not stop till the last poll. So I do not want to raise an exception when it comes but rather log it. Maybe <code>logger.error()</code> is the better way to go about this? I will remove "unhandled exception" and instead just log the error <code>e</code>

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: > Please manage those sleeps with a single custom exception that, when is triggered, it performs the sleep and continues the loop

Good suggestion. Will do.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: Understood. I left it there assuming that celery logs would anyway report the log as completed right after this because the last condition is the condition where the outputs are supposed to be returned. But I will add it there anyway.

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
mlodic: Now is better than before but there are still some problems. My point was: when something happens and you need to continue the loop, you should handle it appropriately at that specific time.

For instance, here (with the 429 status code) you want to sleep and continue the loop but...it does not happen exactly here, while it should. On the contrary you wait for the end of the code to do that. That is not how it should work. You should trigger a Custom Exception here that would bring you directly at the end of the code of the loop where there is a log + sleep.

Same thing for what happens after the `RequestException`. Also there you don't manage the problem directly. In both cases, you should raise the same Custom Exception that would bring you here:
```
            logger.info(
                f"Job: {self.job_id} -> "
                f"Analysis status: {data}. "
                f"Sleeping for {self.poll_distance} seconds before next attempt."
            )
            time.sleep(self.poll_distance)
```

I mean, you need to separate the part where you handle the loop for the first request and the part where you handle the results and do not mix them

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -0,0 +1,246 @@
...
+                else:
+                    r.raise_for_status()
+
+                error_in_poll = False
+
+            except requests.RequestException as e:
+                error_in_poll = True
+                error_details = e
+
+                logger.exception(
```
mlodic, line 159: this cannot be at `exception` level because this is not something not problematic. It is normal that this would happen

mlodic (commented):
thanks for updating the code.

Do you see the errors in the Codefactor linter? Can u fix them? 

one of the problem is that usually `"""` are used for docstrings and not for generic comments (you should use `#`). I know it's annoying but that is the normal code style

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
0x0elliot: Okay I will simply log it then?

0x0elliot:
@mlodic I have tested the changes. They seem to be working fine at my end. Thank you for the help! Please review it once and have it pulled accordingly.

mlodic (approved):

Issue #804, opened 2021-12-16
add Github action tests to check if the sphinx build did work before releasing


mlodic:
Right now, a bug to a dependency for the docs caused the build of the docs to fail in ReadTheDocs.

I noticed it before making a new release randomly. We need to have an automatic and reliable way to check if the docs will work in the releases just before creating them.

devmrfitz:
We can create a test that gets triggered based on [releases](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release). Should I make a PR to that effect?

devmrfitz:
@mlodic Should we mark this issue as resolved?

PR #837, opened 2022-01-09
Sphinx Build Github Action on Release

devmrfitz:
# Description
Added github action to build docs on release

## Related issues

#804 

## Type of change

# Checklist

### Important Rules

### DIFF
--- a/.github/workflows/sphinx_build_on_release.yml
+++ b/.github/workflows/sphinx_build_on_release.yml
@@ -1,6 +1,9 @@
 name: Sphinx Build on Release
 
-on: push
+on:
+  pull_request:
+    branches:
+      - master
 
 jobs:
   build:
--- a/.github/workflows/sphinx_build_on_release.yml
+++ b/.github/workflows/sphinx_build_on_release.yml
@@ -1,6 +1,6 @@
 name: Sphinx Build on Release
 
-on: release
+on: push
 
 jobs:
   build:
@@ -16,11 +16,6 @@ jobs:
       - name: Install docs requirements
         run: |
           pip install -r docs/source/requirements.txt
-
-      # Why do we have to do this?
-      - name: Install additional requirements
-        run: |
-          pip install recommonmark
 
       - name: Build docs
         run: |
--- a/.github/workflows/sphinx_build_on_release.yml
+++ b/.github/workflows/sphinx_build_on_release.yml
@@ -1,0 +1,27 @@
+name: Sphinx Build on Release
+
+on: release
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Install docs requirements
+        run: |
+          pip install -r docs/source/requirements.txt
+
+      # Why do we have to do this?
+      - name: Install additional requirements
+        run: |
+          pip install recommonmark
+
+      - name: Build docs
+        run: |
+          make -C docs html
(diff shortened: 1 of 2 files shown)


.github/workflows/sphinx_build_on_release.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Install docs requirements
+        run: |
+          pip install -r docs/source/requirements.txt
+
+      # Why do we have to do this?
```
mlodic, line 20: why is this comment here?

mlodic (changes requested):
hey, thanks for the PR!

This action should work not on release trigger because it would be too late.

Consider that we make releases each time we open PRs from `develop` to `master`, this action would need to be executed only for the PRs that we do from `develop` to `master`. In this way we can fix the docs if something is wrong there, before creating a new release.

Then, I am not sure that this works as intended.

In RTD, after having built the requirements as you did here, it executed the build with this command:

``` 
/home/docs/checkouts/readthedocs.org/user_builds/intelowl/envs/develop/bin/python -m sphinx -T -E -b html -d _build/doctrees -D language=en . _build/html
```

Can you please make a test in your repository by triggering this action and then give us proof that this would work?

Thank you

devmrfitz:
@mlodic By a test, do you mean that I should re-introduce the error fixed in last release to verify if the test catches it or not?

.github/workflows/sphinx_build_on_release.yml
devmrfitz: Shouldn't the additional package also be specified in docs/requirements.txt ?

eshaan7: yes it should be in the requirements.txt file

devmrfitz:
Here's a link for the same: https://github.com/devmrfitz/IntelOwl/runs/4763166223?check_suite_focus=true

.github/workflows/sphinx_build_on_release.yml
mlodic: RTD already performs some installation by default:

```
 /home/docs/checkouts/readthedocs.org/user_builds/intelowl/envs/develop/bin/python -m pip install --upgrade --no-cache-dir mock==1.0.1 pillow==5.4.1 alabaster>=0.7,<0.8,!=0.7.5 commonmark==0.8.1 recommonmark==0.5.0 sphinx<2 sphinx-rtd-theme<0.5 readthedocs-sphinx-ext<2.2 
```

It does that before installing the requirements in the `docs` folder. So yeah, there works but in the github action is needed.

mlodic:
> @mlodic By a test, do you mean that I should re-introduce the error fixed in last release to verify if the test catches it or not?

No. Considering that this is something that works only on the condition that you specify in the Action, you need to have that action triggered so we can test if it works or not. 

Either you do it in your repo and show us the results or you momentarily set that this action should trigger on PR against develop so we can see if it works here

mlodic:
> Here's a link for the same: https://github.com/devmrfitz/IntelOwl/runs/4763166223?check_suite_focus=true

exactly, something like that but with the action working

devmrfitz:
> > Here's a link for the same: https://github.com/devmrfitz/IntelOwl/runs/4763166223?check_suite_focus=true
> 
> exactly, something like that but with the action working

Does this look good https://github.com/devmrfitz/IntelOwl/actions/runs/1677927980 ?

devmrfitz:
@mlodic Is it alright if the action runs on every PR to master?

.github/workflows/sphinx_build_on_release.yml
devmrfitz: @mlodic Is it alright to add it to docs/source/requirements.txt?

Issue #779, opened 2021-11-25
Prevent running `repo_downloader.sh` in `Dockerfile` for every build


eshaan7:
because it slows down the image building time and hurts development flow

devmrfitz:
@eshaan7 What is the expected behaviour then? I'd like to work on this issue

mlodic:
We need to find a trick to allow developers to do not execute that line (consumes network resources and time) each time they build an image. Right now there is no other way that commenting that line and then remember to uncomment it before pushing a Pull Request. That is obviously not practical.

devmrfitz:
Should we set an environment variable that denotes a development environment (and pass the same into the Dockerfile through start.py) ?

devmrfitz:
Or maybe we could set up a multi-stage build process where the repo_downloader script runs at the very last stage. For development, we won't need to build this stage itself.

@mlodic Should I try to implement one of these approaches and make a PR?

mlodic:
> Should we set an environment variable that denotes a development environment (and pass the same into the Dockerfile through start.py) ?

This could work when working on a local environment. But care: the same Dockerfile is used for automatic builds in [DockerHub](https://hub.docker.com/repository/docker/intelowlproject/intelowl) too. So, arguments must be handled carefully to allow it work there too.

> Or maybe we could set up a multi-stage build process where the repo_downloader script runs at the very last stage. For development, we won't need to build this stage itself.

I am not sure that this would work.

> @mlodic Should I try to implement one of these approaches and make a PR?

Sure, you can make some tries to see which one works. Thanks

eshaan7:
A very simple approach would be to define an `ARG` (build argument) only in the `test.override.yml` file and then in the `Dockerfile` we add an if statement check to only execute the `repo_downloader.sh` if this `ARG` was set.

We can even avoid having to set a new `ARG` and just use the existing `DEBUG` and `DJANGO_SERVER` environment variables but that is not very intuitive.

devmrfitz:
@eshaan7 If the ```--django-server``` flag is present, as indicated [here](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-start-setup-project-and-development-instance), it executes ```django-server.override.yml```, instead of ```test.override.yml```. Hence, I have added the argument in that file too

devmrfitz:
[...]

PR #848, opened 2022-01-12
Prevent repo_downloader in DEVELOPMENT mode

devmrfitz:
# Description

Prevent repo_downloader execution in DEVELOPMENT mode

## Related issues

#779 

## Type of change

# Checklist

### Important Rules


docker/Dockerfile
```diff
@@ -40,8 +42,9 @@ RUN touch ${LOG_PATH}/django/api_app.log ${LOG_PATH}/django/api_app_errors.log \
     # this is cause stringstifer creates this directory during the build and cause celery to crash
     && rm -rf /root/.local
 
-# download yara rules and quark engine rules
-RUN api_app/analyzers_manager/repo_downloader.sh
+# download yara rules and quark engine rules if not in development mode
+RUN if [ $DEVELOPMENT = "true" ]; then echo "Skipping repo_downloader.sh in DEVELOPMENT mode"; \
```
mlodic, line 46: proof of this line working? anyway I would prefer to add a separated bash file with this because one-liners like this are not easy to read.

docker/django-server.override.yml
```diff
@@ -5,6 +5,8 @@ services:
     build:
       context: ..
       dockerfile: docker/Dockerfile
+      args:
+        DEVELOPMENT: true
```
mlodic, line 9: honestly I do not like to have this set as default here. I mean, in this way it is not possible to customize it and effectively run the code in the `repo_downloader.sh` file. If we need to do that we would find ourselves in the same problem as before where we change manually a file with the risk to commit it unwillingly.

The only way to solve this issue would be to add the chance to choose whether to run that bash script or not via one of the file that are NOT commited (`env_file_app`, `.env`)

mlodic (changes requested):

docker/django-server.override.yml
devmrfitz: Even I thought that. However, in the same override file, we are also setting the ```DEBUG``` and ```DJANGO_TEST_SERVER``` variables too. Should we look towards migrating them too?

docker/Dockerfile
devmrfitz: @mlodic Should I upload a screenshot on imgur and share it here for the proof? Additionally, do you want me to move the condition to the ```repo_downloader``` script itself?

mlodic: yep to both! thanks!|

docker/django-server.override.yml
mlodic: no, I mean that is different. If you are in development mode, you always need those 2 variables set.

On the contrary, for the `repo_downloader.sh`, you could need to have it running if you are working on it but most of the time you won't.

docker/Dockerfile
devmrfitz: The condition has been moved to ```repo_downloader.sh```

docker/django-server.override.yml
devmrfitz: I have created a new file in the root called ```.env.test.template``` and added ```.env.test``` to gitignore. Now the developer can freely modify ```.env.test``` without having the risk of accidentally committing it to the repo. Does that look good?

devmrfitz: @mlodic I've also verified it to be working and will shortly post screenshots of the same

devmrfitz:
Script disabled:
Screenshot from 2022-01-16 14-05-57



Script enabled: 
 Screenshot from 2022-01-16 14-59-20

docker/django-server.override.yml
mlodic: there is no need to create new environement files and use the `dotenv` library. As I mentioned earlier, you could use the already existing ones...please check inside the `docker` folder

devmrfitz: @mlodic the ```docker/env_file_*``` files are used to set environment variables. However, here, we need to set a build argument (that determines if the script will be executed or not). The only plausible way to set this argument is from ```start.py```. Hence, I had to create a new .env file to be sourced by the same.

mlodic: thanks for the explanation. I got you. that0s ok so.

Just 2 other things:
* is it possible to add some documentation about this? (maybe in the "hint" section [here](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-start-setup-project-and-development-instance)?)
* other concern is that we are adding another new file to the main directory. We risk to clutter it too much so when it is not 100% needed, we should try to move it somewhere else. Even if it is not docker-related, I think it would make sense to move that new env file in the `docker` folder with the others and change its name in something more intuitive (otherwise we risk to confuse it with the .env file of docker). Maybe `env_start_test` and `env_start_test_template`?

devmrfitz: @mlodic I was wary of using underscore in the name as that could confuse it with docker-compose's ```env_file```s. How is ```.env.start.test```. ```start``` because it is sourced by ```start.py``` and ```test``` as it is meant for test mode

devmrfitz: Should the documentation be in this issue itself or should I create a separate issue for the same?

mlodic: related documentation is always part of the issue itself otherwise we forget to do that :P

devmrfitz: @mlodic resolved all concerns

devmrfitz:
@mlodic all tasks completed

mlodic (approved):

mlodic:
thanks! merged

devmrfitz:
@mlodic I had made a filename typo in this PR. I've made PR #866 for the fix. Is that appropriate or should I try to edit this PR itself?

mlodic:
it is appropriate and thanks for that. I have already merged it

Issue #561, opened 2021-07-22
[tests] API/models permissions, groups


eshaan7:
Inside the `test_api`, we are currently using the admin user so we are basically bypassing all permissions.

We should define test cases where we use normal users and verify their permissions/groups. For example, enabling/disabling certain permissions to a group/user and then making request to a protected endpoint and verify that the behavior changed.

-------
[back reference to previous issue](https://github.com/intelowlproject/IntelOwl/issues/229#issue-723352673)

uzair-ali10:
Hey!!
From what I've understood is that currently in [`test_api`](https://github.com/intelowlproject/IntelOwl/blob/master/tests/test_api.py) all the views are tested against a Super User. And it is desired to test them against a normal user correct??
If this is the case please assign this to me i could take this up

eshaan7:
You are right but the issue #341 needs to be solved prior to working on this.

Issue #341, opened 2021-02-17
Reduce complexity: django object/model permissions


eshaan7:
We are using django object permissions but we are also using django model permissions. At some places, these are intermixed and that is troubling.

m0mosenpai:
Hey, is it possible if you could specificy one or two places where this occurs and causes issues? I'm trying to understand more about the issue a bit, thanks.

Eshaan7:
**Clarification**: Yes, they are supposed to be intermixed and used together but it's complex to manage for any admin managing an IntelOwl instance. 

**Example**: It may be better to overwrite the `get_permissions` method on the viewset rather than using `django-guardian` to store the object permissions.

iharshit009:
Hi @eshaan7, I got some understanding of intelowl and have knowledge of Django Viewsets. Can I take this up and work on it?

eshaan7:
@iharshit009 This issue has by and large solved as part of #856 by making use of the _organizations_ pattern available in [certego_saas](https://github.com/certego/certego-saas).

iharshit009:
> @iharshit009 This issue is by and large solved as part of #856 by making use of the _organizations_ pattern available in [certego_saas](https://github.com/certego/certego-saas).

Okay no issues, Let me know if I could contribute to this in any way. Will look for other issues.

mlodic:
I consider this closed so we can start to better filter out issues for the v4 milestone

PR #856, opened 2022-01-18
initial v4 changes (frontend in reactjs, backend using certego-saas)

eshaan7:
**ToDo:**
- [x] add a data migration for durin's API access client.
- [x] Use "organizations" app for access control / job level permissions. <sup>[1]</sup>
- [x] Remove the usage of django-guardian.  <sup>[2]</sup>
- [x] Changes were made to API URLs so we should make corresponding changes in pyIntelOwl. (https://github.com/intelowlproject/pyintelowl/commit/90df9ec203da8c9052cc943d5d2f1b9a38ba5613)
- [ ] Write more tests and fix existing. We should write tests also to check the new permissions system (#561).

*requires further discussion:*
- [ ] Review using httpOnly flagged cookie for authentication on the frontend.

Other related:
- [x] perform some manual tests
- [x] update frontend dependencies, removing deprecated ones
- [ ] add basic tests suite for the frontend, see #929 
- [x] update all the documentation

-----------------------
**Notes:**

[1] Simply put: for a "private" job (i.e. TLP >= AMBER), the same organization members should be able view/retry all jobs created by its members but kill/delete permission should be restricted to only the job owner.
[2] See #341. We wish to change how the user level permissions on Job model are being handled so is a breaking change. Right now user's have granular control but is very complicated to use, this change will simplify the management but won't provide users with granular control. (Maybe we should take a poll on twitter asking our audience if they are okay with this just to err on the side of caution ?)


mlodic:
got these 2 errors after first manual launch:

```
django.db.migrations.exceptions.InconsistentMigrationHistory:` Migration admin.0001_initial is applied before its dependency certego_saas.0001_initial on database 'default'.
```

```
PermissionError: [Errno 13] Permission denied: '/tmp/dj_cache_intelowl_default/tmp0bl24ebu'
```

Issue #876, opened 2022-02-03
Remove non-standard library dependency in start.py script


mlodic:
With the #848 we unwillingly added a new dependency that is not included in the python stardard library

```
from dotenv import load_dotenv
```

We do not want to include this additional library as a requirement because it complicates the installation. We need to refactor this code before deploy a new version

eshaan7:
This change was already included in the [v3.2.4 release](https://github.com/intelowlproject/IntelOwl/blob/master/.github/CHANGELOG.md#v324).

mlodic:
oh no 🥲

mlodic:
I guess we'll need to push a patch soon

mlodic:
However I am thinking about the fact that in the future we could need also other libraries and we cannot just avoid to installing them for this reason.

So I was thinking that an alternative could be to provide a bash script that would install the already existing requirements, plus the additional python requirements in case (like `dotenv`. It could be interactive and it could check if there is already an installation of those requirements in the system.

Something like:
"This script will help you in the installation of IntelOwl dependencies" -> scripts check docker installation -> "you have (not) installed docker with the version X. You want to install the most recent version?" -> yes/no -> installation -> loop for the other requirements

devmrfitz:
@mlodic I was not aware of the restriction on third-party libraries. I apologise for this mishappening. However, I added the package in ```requirements.txt```. Won't the package automatically get installed when ```RUN pip3 install --no-cache-dir --compile -r requirements.txt``` is run (from ```docker/Dockerfile```.

mlodic:
noproblem man, was my fault during the review, not yours.

I saw that, the problem is that this is a requirement for the startup script. When you run IntelOwl, you do not need to install those requirements because they were already installed in the image that do not need the `dotenv` package. So, `dotenv` should be removed from requirements.

mlodic:
I mean, with the actual configuration, we would need to add that package to the initial requirements [here](https://intelowl.readthedocs.io/en/latest/Installation.html#requirements). That is bad because it would slow down the installation process

devmrfitz:
@mlodic Oh. Got it. Just to reiterate, we are free to add new packages inside the docker container but adding additional imports in ```start.py``` lengthens the manual installation process?

mlodic:
well, completely free no because we need to consider multiple factors before adding new dependencies. Is it a good written and maintained dep? Do we really need it?

[...]

PR #878, opened 2022-02-07
Added script to check and install initial requirements. Solves #876

devmrfitz:
# Description

Added script to check and install initial requirements. Solves #876

## Related issues
#876 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue).

# Checklist

### Important Rules

# Real World Example


devmrfitz:
@mlodic Please review

initialize.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+set -e
+
```
mlodic, line 4: is it possible to write something welcoming the user? 

"this is the helper script to help you install the OS requirements for IntelOwl" or something similar.

initialize.sh
```diff
@@ -0,0 +1,61 @@
...
+    sh get-docker.sh
+    # Check if docker is installed
+    if ! [ -x "$(command -v docker)" ]; then
+      echo 'Error: Could not install docker.' >&2
+      exit 1
+    fi
+  else
+    exit 1
+  fi
+fi
```
mlodic, line 23: is it also possible to write something if the requirements are actually installed?

Example:
"checking docker installation...."

then

"Docker is correctly installed with version X"

otherwise

what you have already done

initialize.sh
```diff
@@ -0,0 +1,61 @@
...
+  # Ask if user wants to install docker-compose
+  read -p "Do you want to install docker-compose? [y/n] " -n 1 -r
+  echo
+  if [[ $REPLY =~ ^[Yy]$ ]]; then
+    # Install docker-compose
+    sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+    sudo chmod +x /usr/local/bin/docker-compose
+    sudo ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose
+    # Check if docker-compose is installed
+    if ! [ -x "$(command -v docker)" ]; then
```
mlodic, line 37: `docker-compose` here

initialize.sh
```diff
@@ -0,0 +1,61 @@
...
+      echo 'Error: Could not install docker-compose.' >&2
+      exit 1
+    fi
+  else
+    exit 1
+  fi
+fi
+
+# Check if python is installed
+if ! [ -x "$(command -v python)" ]; then
```
mlodic, line 47: I would look for `python3` instead. Plus could you add the installation too? See: https://linuxize.com/post/how-to-install-python-3-9-on-ubuntu-20-04/

Then, if python is already installed we would need to check if the requirement is satisfied (v3.6+)

initialize.sh
```diff
@@ -0,0 +1,61 @@
...
+    sudo ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose
+    # Check if docker-compose is installed
+    if ! [ -x "$(command -v docker)" ]; then
+      echo 'Error: Could not install docker-compose.' >&2
+      exit 1
+    fi
+  else
+    exit 1
+  fi
+fi
```
mlodic, line 44: if it is installed, we should also check the version. See: https://intelowl.readthedocs.io/en/latest/Installation.html#requirements.

In the case the version is old, for docker-compose it would be enough to install the version you installed here over it and that would work

initialize.sh
mlodic: If the version is old, we won't manage this problem because too hard. But we need to tell them that they need to change the version so the users can do on their own

initialize.sh
mlodic: We should also check the version of docker and tell them that the version is too old in case. No need to manage the uninistall/reinstall because too complicated imho. But at least the users know the problem and can fix it on their own

initialize.sh
```diff
@@ -0,0 +1,61 @@
...
+if ! [ -x "$(command -v python)" ]; then
+  echo 'Error: python is not installed.' >&2
+  exit 1
+fi
+
+# Check if pip is installed
+if ! [ -x "$(command -v pip)" ]; then
+  echo 'Error: pip is not installed.' >&2
+  exit 1
+fi
```
mlodic, line 56: we could add the installation/update of pip

initialize.sh
```diff
@@ -0,0 +1,61 @@
...
+
+echo "Installing python dependencies using pip..."
+pip install -r pre-requirements.txt
+
+echo "Looks like you're ready to go!"
```
mlodic, line 61: We could also add here which is the next step (running start.py)

docs/source/Installation.md
```diff
@@ -18,7 +18,7 @@ In some systems you could find pre-installed older versions. Please check this a
 </div>
```
mlodic, line 18: we could also add in the requirements section:

* that we added this helper script (in the Notes)

* that, together with python, there are other dependencies to install

mlodic (changes requested):
I really appreciate the effort.

To really be helpful, we need to also manage all the possible cases appropriately.

initialize.sh
devmrfitz: > I would look for `python3` instead. Plus could you add the installation too? See: https://linuxize.com/post/how-to-install-python-3-9-on-ubuntu-20-04/
> 
> Then, if python is already installed we would need to check if the requirement is satisfied (v3.6+)

@mlodic I thought of doing this. But that would limit us to systems using apt. The other scripts (docker and docker-compose) I've used support a wider variety of distros

initialize.sh
devmrfitz: > we could add the installation/update of pip

@mlodic Same apt dependency as above

initialize.sh
mlodic: this makes sense, thank you for the explanation. so I think it is enough to just tell it to the user

initialize.sh
mlodic: ok

also noticed that it should be pip3 too here

initialize.sh
devmrfitz: @mlodic Overwriting a user's docker-compose (in ```/usr/local/bin/docker-compose```) could lead to unexpected issues with other software as they might be depending on older version

devmrfitz:
@mlodic Resolved the issues. Please review

initialize.sh
mlodic: ok so we might keep the same approach as the other cases (just telling the user what they should do then). In that way it makes more sense because there is the same behavior for every case.

initialize.sh
```diff
@@ -0,0 +1,130 @@
...
+            ver2[i]=0
+        fi
+        if ((10#${ver1[i]} > 10#${ver2[i]}))
+        then
+            echo 1
+            return
+        fi
+        if ((10#${ver1[i]} < 10#${ver2[i]}))
+        then
+            echo 2
```
mlodic, line 37: would it be better to use a string instead of an integer for better clarity?

mlodic (changes requested):
appreciated the efforts.

initialize.sh
```diff
@@ -0,0 +1,130 @@
...
+    # Check if docker-compose is installed
+    if ! [ -x "$(command -v docker-compose)" ]; then
+      echo 'Error: Could not install docker-compose.' >&2
+      exit 1
+    fi
+  else
+    exit 1
+  fi
+else
+  docker_compose_version=$(docker-compose --version| awk '{print $NF}')
```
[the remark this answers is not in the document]
devmrfitz: What does ```docker-compose --version``` print for you?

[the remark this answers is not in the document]
devmrfitz: @mlodic

initialize.sh
devmrfitz: @mlodic "eq", "lt", "gt" instead of 1,2,3?

mlodic: the full word even better I guess

initialize.sh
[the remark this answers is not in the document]
devmrfitz: image
Weird. This is what I get

initialize.sh
devmrfitz: done

initialize.sh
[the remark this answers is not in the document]
mlodic: you have a major version. I think that is why they are different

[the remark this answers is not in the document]
devmrfitz: I pushed a patch. It should work properly in both cases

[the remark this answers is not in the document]
mlodic: image
still getting same result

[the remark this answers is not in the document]
mlodic: it does! does it work for your major 2.x version?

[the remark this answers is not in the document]
devmrfitz: Yeah, it does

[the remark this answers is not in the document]
mlodic: screen? :P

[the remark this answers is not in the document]
devmrfitz: image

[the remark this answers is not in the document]
mlodic: really loved this! thank you!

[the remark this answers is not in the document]
devmrfitz: :)

Issue #882, opened 2022-02-09
Correctly manage docker-analyzer dependencies


mlodic:
Right now we rely on Dependabot to keep updated the high number of dependencies of this project.

Dependabot works by reading through "requirements" files that are standards of the chosen language (ex: "requirements.txt", "pyproject.toml", etc)

I noticed that there are some Docker-analyzers where we hardcoded some dependencies directly inside the Dockerfile. This causes the Dependabot to do not retrieve updates of those dependencies and, as a consequence, there is no way that they will be updated ever.

So, we should move them outside the Dockerfile when possible.

Python dependencies:
- [x] flare-capa
- [x] stringsifter

It would be great to solve this together with this issue: #842

Yshaq:
Hey, can I be assigned this issue?
I'd like to work on it along with issue #842

Yshaq:
For this issue, since I just have to remove the 2 installation lines from dockerfile and add the packages to requirements.txt,
would a single PR be good for this issue and #842 together, or should I make a seperate PR for this one first?

mlodic:
as you wish...anyway you can separate them without problems

Yshaq:
Ok i'll make a PR for this one then coz the other one will probably take longer to finish

Issue #708, opened 2021-09-17
Capa v3 update


mlodic:
https://www.fireeye.com/blog/threat-research/2021/09/elfant-in-the-room-capa-v3.html

Yshaq:
Hey, I'd like to pick up this issue.

Installing capa directly from pip won't work because that doesn't set up the default library identification signatures. This way, we would have to somehow download the sigs folder from the [capa repo](https://github.com/mandiant/capa) and also modify the capa command being run.

What I'm thinking of doing is cloning the capa repo, and then [installing capa with pip from the local source](https://github.com/mandiant/capa/blob/master/doc/installation.md#method-3-inspecting-the-capa-source-code). This way the sigs are installed and the current command running need not be modified. Shall I go ahead with this method?

mlodic:
Yes, remember that it is important to bring proof of a successful result so we can clearly understand that this method works! Thanks!

mlodic:
done with v3.3.0

PR #885, opened 2022-02-10
package installations moved from dockerfile to requirements.txt for dependabot

Yshaq:
# Description
suitable package installations moved from dockerfile to requirements.txt so that they can be managed by dependabot.

## Related issues
#882 

## Type of change

# Checklist

### DIFF
--- a/integrations/static_analyzers/requirements.txt
+++ b/integrations/static_analyzers/requirements.txt
@@ -1,2 +1,4 @@
 Flask-Shell2HTTP~=1.8.0
 gunicorn~=20.1.0
+flare-capa~=2.0.0
+stringsifter~=2.20201202
--- a/integrations/static_analyzers/Dockerfile
+++ b/integrations/static_analyzers/Dockerfile
@@ -15,8 +15,8 @@ RUN useradd -ms /bin/bash static_analyzers-user
 
 # Install capa
 WORKDIR ${PROJECT_PATH}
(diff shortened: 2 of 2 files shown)


integrations/static_analyzers/requirements.txt
```diff
@@ -1,2 +1,4 @@
 Flask-Shell2HTTP~=1.8.0
-gunicorn~=20.1.0
\ No newline at end of file
+gunicorn~=20.1.0
+flare-capa==2.0.0
```
mlodic, line 3: can you use ~= so we can get patches automatically?

mlodic: plus, can you check the status of the current available versions? I guess this is old

Yshaq: Hey @mlodic . I've made the requested change.

Also, I looked into the packages. stringsifter is the same version but yes Capa is in v3.
I've looked into the tool. It seems that Capa v2 isnt working well with the latest rule set.
But just changing the dependency here wouldnt work because installing capa via pip doesnt setup the default library identification [signatures](https://github.com/mandiant/capa/tree/master/sigs)

I see there's already an issue for this, #708 . I can pick that up, I have a fair idea on what to do.

mlodic: sure, it would be great! you can pick up that issue! Meanwhile I accept this PR so we can close the first issue

Issue #967, opened 2022-04-12
MISP better configurability


BenjaminH-Sogeti:
# Issue
MISP has actually a really low level of configurability
- Hardcoded time constraint (events not older than 90 days)
- Hardcoded results limitation (50)
- Type of attribute filter is mandatory
- Search is only made on events field and not on attributes

# Doc 
Search() method start [here](https://buildmedia.readthedocs.org/media/pdf/pymisp/latest/pymisp.pdf) at page 29 

# Improvements
Offer a better configurability by moving these value into analyzer_config.json params and make them optional in case of null value.

mlodic:
yeah, thanks for your interest, but policy wants that each person should just have one assigned task at a time.

BenjaminH-Sogeti:
After our discussion in #979 
And discovering the different types of misp [data models](https://www.misp-project.org/datamodels/), I think that it could be useful to add "generic ioc" search format without type-filter. 

It could be useful to search for IBAN, filename, ... (everything that could be set into misp)

https://github.com/intelowlproject/IntelOwl/blob/ceebcc5042885ba49af564106a6cf419f5a31f63/configuration/analyzer_config.json#L1561

I may handle it if it also make sense to you !

mlodic:
yeah it makes sense. Consider that it was not there before because we added the "generic ioc" feature later than that MISP analyzer. Would you like to make a PR for that? In case, I can wait for that to make a new release with everything you already made. Thank you!

BenjaminH-Sogeti:
> Would you like to make a PR for that?

Sure

PR #979, opened 2022-04-15
Better configurability into misp analyzer

BenjaminH-Sogeti:
# Description
Remove hardcode value for filter and add configurability 
## Related issues
#967 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue).).

# Checklist

### Important Rules

# Real World Example
## No filter 
image

## Days filter
image

## Remarks 
- I can't test type filter
- limit with value 0 return result (maybe already handle as "no limit" in MISP

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/misp.py
+++ b/api_app/analyzers_manager/observable_analyzers/misp.py
@@ -14,6 +14,10 @@ class MISP(classes.ObservableAnalyzer):
     def set_params(self, params):
         self.ssl_check = params.get("ssl_check", True)
         self.debug = params.get("debug", False)
+        self.from_days = params.get("from_days", 90)
+        self.limit = params.get("limit", 50)
+        self.enforce_warninglist = params.get("enforce_warninglist", True)
+        self.filter_on_type = params.get("filter_on_type", True)
         self.__url_name = self._secrets["url_key_name"]
         self.__api_key = self._secrets["api_key_name"]
 
@@ -25,21 +29,19 @@ class MISP(classes.ObservableAnalyzer):
             debug=self.debug,
             timeout=5,
         )
-
-        # we check only for events not older than 90 days and max 50 results
         now = datetime.datetime.now()
-        date_from = now - datetime.timedelta(days=90)
+        date_from = now - datetime.timedelta(days=self.from_days)
         params = {
-            # even if docs say to use "values",...
-            # .. at the moment it works correctly only with "value"
             "value": self.observable_name,
-            "type_attribute": [self.observable_classification],
-            "date_from": date_from.strftime("%Y-%m-%d %H:%M:%S"),
-            "limit": 50,
-            "enforce_warninglist": True,
+            "limit": self.limit,
+            "enforce_warninglist": self.enforce_warninglist,
         }
-        if self.observable_classification == self.ObservableTypes.HASH:
-            params["type_attribute"] = ["md5", "sha1", "sha256"]
+        if self.from_days != 0:
+            params["date_from"] = date_from.strftime("%Y-%m-%d %H:%M:%S")
+        if self.filter_on_type:
+            params["type_attribute"] = [self.observable_classification]
+            if self.observable_classification == self.ObservableTypes.HASH:
+                params["type_attribute"] = ["md5", "sha1", "sha256"]
         result_search = misp_instance.search(**params)
         if isinstance(result_search, dict):
             errors = result_search.get("errors", [])
--- a/api_app/analyzers_manager/observable_analyzers/misp.py
+++ b/api_app/analyzers_manager/observable_analyzers/misp.py
@@ -6,7 +6,7 @@ import datetime
 import pymisp
 
 from api_app.analyzers_manager import classes
-from api_app.exceptions import AnalyzerRunException
+from api_app.exceptions import AnalyzerConfigurationException, AnalyzerRunException
 from tests.mock_utils import MockResponseNoOp, if_mock_connections, patch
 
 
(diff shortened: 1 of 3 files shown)


api_app/analyzers_manager/observable_analyzers/misp.py
```diff
@@ -25,21 +29,19 @@ def run(self):
...
-            "limit": 50,
-            "enforce_warninglist": True,
+            "limit": self.limit,
+            "enforce_warninglist": self.enforce_warninglist,
         }
-        if self.observable_classification == self.ObservableTypes.HASH:
-            params["type_attribute"] = ["md5", "sha1", "sha256"]
+        if self.from_days != 0:
+            params["date_from"] = date_from.strftime("%Y-%m-%d %H:%M:%S")
+        if self.filter_on_type:
```
mlodic, line 41: considering that we analyze a single observable at a time, there is no need for this flag.

The rest is good! thank you!

mlodic: I'll merge after this change

BenjaminH-Sogeti: Which flag are you talking about ?

mlodic: `if self.filter_on_type:`

BenjaminH-Sogeti: I have to admit that I don't get your point. 
The goal of this flag is to give option for the user to search (or not) on the attribute type in their MISP instance.
In my case my MISP instance doesn't specify the type of observable for entries, thus if I specify the type_attribute in the MISP query it return no queries. 
If I am wrong could you explain broadly why the flag is useless? Thanks

mlodic: First, I have not been working with a MISP instance for quite some time, so thanks for helping me understanding the reason of this change.

The reason why, right now, the parameter `type_attribute` is mandatory is:
* all the attributes must have a category: url, domain, etc. So are you telling me that now this is no more mandatory? Can you ping me either documentation or a real use case where this actually happens?
* performance. This filter reduces the time required to perform the query. This is critical for scalability purposes. This is why I cannot easily accept this change without being sure 100% that we are doing the right thing.

BenjaminH-Sogeti: ## Mandatory 
[Here ](https://buildmedia.readthedocs.org/media/pdf/pymisp/latest/pymisp.pdf) (page 29) is the doc specifying than it is optional : 
`type_attribute (Optional[TypeVar(SearchParameterTypes, str,
List[Union[str, int]], Dict[str, Union[str, int]])]) – The attribute type, any
valid MISP attribute type is accepted.`

"Any valid MISP attribute type is accepted", [here](https://www.misp-project.org/datamodels/) are the misp attribute type and "ip" isn't an existing one. 

### Filter on type
No entry returned
image

### No filter on type 
One entry returned 
#### Pic
image

### Remark
As we can see in last picture the type is in my case "ip-dst", thus I believe than intelowl filter on "ip" and is a strict search, I can investigate on the use of wildcard for "ip*" if you want, or use each attribute type for IPs.  

## Performance
I got the point but by default the parameter is set to true. It wouldn't change anything to common user. (unless an if comparison which is negligible)

mlodic: thank you again for the time invested in this! all extremely useful.

Ok, so at the end the big problem is for the IP addresses that were *never* extracted. This is a pretty bad bug. I'll definitevely make a release out of that.

So I'd say that we could leave the code like you made it + just customize the `type_attribute` list in case the observable is an IP address.

```
            if self.observable_classification == self.ObservableTypes.IP:
                params["type_attribute"] = ["ip-dst", "ip-src", "ip-src|port", "ip-dst|port"]
```

What do you think?

BenjaminH-Sogeti: > Ok, so at the end the big problem is for the IP addresses that were _never_ extracted. 
At least not on this version of misp. 

> I'd say that we could leave the code like you made it
IMHO the more configurable it is for the end user the better

> just customize the type_attribute 
Yes totally, concerning the type _domain|ip_ would it go for ip, domain, both or none ? I would say both

mlodic: yes, I'd say both too. better adding than removing in any case I think.

BenjaminH-Sogeti: Ok I'll handle that

BenjaminH-Sogeti: Done but I messed up with the rebase... Hope you can catch up with my mistake, otherwise I have a clean branch that didn't rebase your commit, I can make a new PR if it is better for you !

mlodic (approved):

api_app/analyzers_manager/observable_analyzers/misp.py
mlodic: np I will squash the commits here anyway so the merge should be clean

integrations/malware_tools_analyzers/clamav/freshclam.conf
```diff
@@ -15,5 +15,4 @@ Foreground true
 DatabaseMirror db.local.clamav.net
 DatabaseMirror database.clamav.net
 ScriptedUpdates yes
-NotifyClamd /etc/clamav/clamd.conf
```
eshaan7, line 18: Was this an intentional change or got included by mistake ?

BenjaminH-Sogeti: I messed up my git rebase and included commit e037d32f44bc6c531e1eae9e6917142ff36ac03b in it

mlodic: yes that was mine

Issue #732, opened 2021-10-11
[feature] bulk analysis


eshaan7:
Bulk analyze of different types of IoCs in one shot. For example, could be done by pasting such lines in a box:

```
ip:192.168.1.25
domain:example.com
hash-sha256:ccadd99b16cd3d200c22d6db45d8b6630ef3d936767127347ec8a76ab992c2ea
```

mlodic:
closed via v4.0.0

PR #1032, opened 2022-06-06
Multiple IOC

devmrfitz:
# Description

Added support for multiple IOC analysis

## Related issues
#732 

## Type of change

# Checklist

### DIFF
--- a/api_app/serializers.py
+++ b/api_app/serializers.py
@@ -144,33 +144,6 @@ class _AbstractJobCreateSerializer(rfs.ModelSerializer):
                         f"{a_name} won't be run because you"
                         " filtered external analyzers."
                     )
-                if serialized_data["is_sample"]:
-                    if not config.is_type_file:
-                        raise NotRunnableAnalyzer(
-                            f"{a_name} won't be run because does not support files."
-                        )
-                    if not config.is_filetype_supported(
-                        serialized_data["file_mimetype"]
-                    ):
-                        raise NotRunnableAnalyzer(
-                            f"{a_name} won't be run because mimetype "
-                            f"{serialized_data['file_mimetype']} is not supported."
-                        )
-                else:
-                    if not config.is_type_observable:
-                        raise NotRunnableAnalyzer(
-                            f"{a_name} won't be run because "
-                            "it does not support observable."
-                        )
-
-                    if not config.is_observable_type_supported(
-                        serialized_data["observable_classification"]
-                    ):
-                        raise NotRunnableAnalyzer(
-                            f"{a_name} won't be run because "
-                            "it does not support observable type "
-                            f"{serialized_data['observable_classification']}."
-                        )
             except NotRunnableAnalyzer as e:
                 if run_all:
                     # in this case, they are not warnings but
--- a/api_app/views.py
+++ b/api_app/views.py
@@ -294,7 +294,7 @@ def analyze_multiple_files(request):
 
 @add_docs(
     description="This endpoint allows to start a Job related to an observable. "
-    "Retained for compatibility",
+    "Retained for retro-compatibility",
     request=ObservableAnalysisSerializer,
     responses={200: AnalysisResponseSerializer},
 )
--- a/docs/source/schema.yml
+++ b/docs/source/schema.yml
@@ -71,7 +71,7 @@ paths:
     post:
       operationId: analyze_observable_create
       description: This endpoint allows to start a Job related to an observable. Retained
-        for compatibility
+        for retro-compatibility
       tags:
       - analyze_observable
       requestBody:


api_app/serializers.py
```diff
@@ -77,18 +80,194 @@ class _AbstractJobCreateSerializer(rfs.ModelSerializer):
...
     def validate(self, attrs: dict) -> dict:
         # check and validate runtime_configuration
         runtime_conf = attrs.get("runtime_configuration", {})
         if runtime_conf and isinstance(runtime_conf, list):
             runtime_conf = json.loads(runtime_conf[0])
         attrs["runtime_configuration"] = runtime_conf
+        return attrs
+
+    @staticmethod
+    def filter_analyzers(serialized_data: Dict, warnings: List) -> List[str]:
```
0ssigeno, line 96: Since you have decided to move this stuff, i would love that we remove the fact that we modify the parameters of the function, inside the function itself.
What i mean is pretty much change the signature to 
```python3
def filter_analyzers(serialized_data: Dict) -> Tuple[List[str], List[str]]:
```
And concatenate the `warnings` in the caller

api_app/serializers.py
```diff
@@ -77,18 +80,194 @@ class _AbstractJobCreateSerializer(rfs.ModelSerializer):
...
+                else:
+                    logger.warning(e)
+                    warnings.append(str(e))
+            else:
+                cleaned_analyzer_list.append(a_name)
+
+        return cleaned_analyzer_list
+
+    @staticmethod
+    def filter_connectors(serialized_data: Dict, warnings: List[str]) -> List[str]:
```
0ssigeno, line 186: Since you have decided to move this stuff, i would love that we remove the fact that we modify the parameters of the function, inside the function itself.
What i mean is pretty much change the signature to 
```python3
def filter_analyzers(serialized_data: Dict) -> Tuple[List[str], List[str]]:
```
And concatenate the `warnings` in the caller

api_app/serializers.py
```diff
@@ -77,18 +80,194 @@ class _AbstractJobCreateSerializer(rfs.ModelSerializer):
...
+        attrs["analyzers_to_execute"] = self.filter_analyzers(
+            attrs,
+            warnings,
+        )
 
+        if not attrs["analyzers_to_execute"]:
+            raise ValidationError(
+                {"detail": "No Analyzers can be run after filtering."}
+            )
+
```
0ssigeno, line 258: Can't we raise the error in `filter_analyzers` directly?

api_app/views.py
```diff
@@ -107,7 +94,81 @@ def _analysis_request(
...
+        f"_analysis_request {serializer_class} received request from {request.user}."
+        f"Data:{dict(request.data)}."
+    )
+
+    # serialize request data and validate
+    serializer = serializer_class(data=request.data, many=True)
+    serializer.is_valid(raise_exception=True)
+
+    serialized_data = serializer.validated_data
+    print(serialized_data)
```
0ssigeno, line 127: Do not use `print` to debug

api_app/views.py
```diff
@@ -107,7 +94,81 @@ def _analysis_request(
...
+            for index, job in enumerate(jobs)
+        ],
+        many=True,
+    )
+    ser.is_valid(raise_exception=True)
+
+    response_dict = {
+        "count": len(ser.data),
+        "results": ser.data,
+    }
```
0ssigeno, line 171: I have to admit that I do not like to have 2 endpoints that share 95% of their code. The only difference that this endpoint has with `_analysis_request` is the `many=True` and the iteration on the `jobs`.
I think i would prefer a single method that manages both cases and, if necessary (meaning that you can't discern if you have `many=True` or not just by checking `request.data`) another argument of the function that is used to distinguish between the two cases

api_app/views.py
```diff
@@ -148,54 +209,63 @@ def ask_analysis_availability(request):
     )
 
     serializer = JobAvailabilitySerializer(
-        data=data_received, context={"request": request}
+        data=data_received, context={"request": request}, many=True
```
0ssigeno, line 212: This endpoints now expect a `list of jobs`. I can see stuff breaking with this change (pyintelowl).
Again, i would prefer that we check the input type and manage it internally the function, without  having breaking changes in the project.

0ssigeno (changes requested):

0ssigeno:
In any case, please do not open 2 PRs with the same code changes: open a single PR with both your changes, frontend + backend, so that we can actually focus on review the changes a single time AND we can actually test the new feature in our computers.

devmrfitz:
> In any case, please do not open 2 PRs with the same code changes: open a single PR with both your changes, frontend + backend, so that we can actually focus on review the changes a single time AND we can actually test the new feature in our computers.

Done

api_app/serializers.py
```diff
@@ -179,10 +358,53 @@ def validate(self, attrs: dict) -> dict:
         file_obj.seek(0)
         file_buffer = file_obj.read()
         attrs["md5"] = calculate_md5(file_buffer)
+        self.filter_analyzers_and_connectors(attrs)
```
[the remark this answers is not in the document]
devmrfitz: That's what I had initially planned. However, that can't be done as for filter_analyzers_and_connectors depends on attrs["file_mimetype"], which is not set until FileAnalysisSerializer.validate runs.

api_app/serializers.py
```diff
@@ -226,13 +449,14 @@ def validate(self, attrs: dict) -> dict:
             attrs["observable_name"] = attrs["observable_name"].lower()
         # calculate ``md5``
         attrs["md5"] = calculate_md5(attrs["observable_name"].encode("utf-8"))
+        self.filter_analyzers_and_connectors(attrs)
```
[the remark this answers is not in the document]
devmrfitz: That's what I had initially planned. However, that can't be done as for filter_analyzers_and_connectors depends on attrs["file_mimetype"], which is not set until FileAnalysisSerializer.validate runs.

api_app/views.py
devmrfitz: One endpoint is left as it is for backward compatibility to be removed in the future version

api_app/serializers.py
```diff
@@ -179,10 +358,53 @@ def validate(self, attrs: dict) -> dict:
...
+            item.pop("observables", None)
+            item["observable_name"] = name
+            item["observable_classification"] = classification
+            try:
+                validated = self.child.run_validation(item)
+            except ValidationError as exc:
+                errors.append(exc.detail)
+            else:
+                ret.append(validated)
+                errors.append({})
```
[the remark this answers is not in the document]
devmrfitz: > When you can, change the PR template too, you left the default one!

Done

devmrfitz:
> When you can, change the PR template too, you left the default one!

Done

api_app/serializers.py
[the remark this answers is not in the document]
devmrfitz: This is carried from from its original source

api_app/serializers.py
devmrfitz: There are 2 approaches that can be taken here:
1) Return `warnings` from the filtering functions.
2) Set `self.warnings` in the filtering functions after making them non-static.
Which one would you prefer?
@0ssigeno @mlodic

0ssigeno: I'm ok with both approach!
Probably remove the @statichmethod is slightly better, but both works for me

api_app/views.py
0ssigeno: For what reason we should remove the other endpoint? If we can avoid breaking changes at any time, is better.
Other than that, I talked about `endpoint`, but what i mean here are the functions `_multi_observable_analysis_request` and `_analysis_request`.

api_app/serializers.py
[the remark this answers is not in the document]
0ssigeno: So wait, we now have the method `filter_analyzers_and_connectors` in the class `_AbstractJobCreateSerializer` that checks "fields" (i.e. serialized_data["file_mimetype"]) that could not be set? You see that this logically makes no sense.
My approach is to actually remove that check in `filter_analyzers_and_connectors`, and in the class `FileAnalysisSerializer` override the method and _there_ add again this check

mlodic:
By closing the previous PR, we lost the steps of the review I previously done...

They still need to be addressed before closing this PR. https://github.com/intelowlproject/IntelOwl/pull/1031

Anyway, this is an inter-communication error that we need to address for the next tasks:
- Always use shared communication channels for coordination so all the mentors are aligned with the WIP.
- Just open a single PR for feature. Frontend/backend are dependant from each other cause one is useless without the other so it makes sense that they are always together. One PR against the branch of development should address the entire problem, not just a part of it.

api_app/views.py
mlodic: I agree with Simone. With a little more effort and following the implementation he suggested, we could have a more serious result.

frontend/src/components/scan/ScanForm.jsx
```diff
@@ -104,7 +106,7 @@ const observableType2PropsMap = {
 };
 const initialValues = {
   classification: "ip",
-  observable_name: "",
+  observable_names: [""],
```
mlodic, line 109: this is not empty.

frontend/src/components/scan/ScanForm.jsx
```diff
@@ -239,8 +249,8 @@ export default function ScanForm() {
         connectors: values.connectors.map((x) => x.value),
       };
       try {
-        const jobId = await createJob(formValues);
-        setTimeout(() => navigate(`/jobs/${jobId}`), 1000);
+        await createJob(formValues);
+        setTimeout(() => navigate(`/jobs/`), 1000);
```
mlodic, line 253: if the request would be a single request, the behaviour should be the same as before. This is correct only

frontend/src/components/scan/ScanForm.jsx
```diff
@@ -213,15 +215,23 @@ export default function ScanForm() {
...
+            }
+            return null;
+          }
         );
-        if (!pattern.test(values.observable_name)) {
-          errors.observable_name = `invalid ${values.classification}`;
-        }
+
+        if (ObservableNamesErrors.some((e) => e))
+          errors.observable_names = ObservableNamesErrors;
```
drosetti, line 232: I think a comment helps to make this code easier to understand: you create an error list and then check with `some()` if at least an error exist, this is fine but at first sight it's not so clear

mlodic:
Another thing...you requested the PR for the frontend code too but, right now, your code does break the "Scan form" completely at line 331...so you requested a review when you didn't even check whether it works manually? Honestly I am disappointed about that.

frontend/src/components/scan/ScanForm.jsx
```diff
@@ -295,29 +305,67 @@ export default function ScanForm() {
...
+                  name="observable_names"
+                  render={(arrayHelpers) => (
+                    <FormGroup row>
+                      <Label className="required" sm={3} for="observable_name">
+                        Observable Value(s)
+                      </Label>
+                      <Col sm={9}>
+                        {formik.values.observable_names &&
+                        formik.values.observable_names.length > 0
+                          ? formik.values.observable_names.map(
```
mlodic, line 325: not necessary the checks

mlodic:
Me and Daniele have fixed some problems regarding the management of the list of observables. There are still problems regarding the `_askAnalysisAvailability` endpoint which does not work from the GUI

mlodic:
The Scan Form was broken and did not load due to an error in the access of the first value of an empty list, caused by changes in your code. this is the critical part: https://github.com/intelowlproject/IntelOwl/pull/1032/commits/4d8c206a4f5fa880454cb58187dc9160eb10ae84#diff-b610d305059891e8cc0965d0fb97abed541d51557bd5c44b4b5ee927a19bcf0dL339-L340

devmrfitz:
> The Scan Form was broken and did not load due to an error in the access of the first value of an empty list, caused by changes in your code. this is the critical part: [4d8c206#diff-b610d305059891e8cc0965d0fb97abed541d51557bd5c44b4b5ee927a19bcf0dL339-L340](https://github.com/intelowlproject/IntelOwl/commit/4d8c206a4f5fa880454cb58187dc9160eb10ae84#diff-b610d305059891e8cc0965d0fb97abed541d51557bd5c44b4b5ee927a19bcf0dL339-L340)

Got it. The error was due to `formik.touched.observable_names[index]` specifically, right?
As of now, we are not tracking the "touched" state of an input field. Should I figure out a way to do that?

mlodic:
the problem was also in the `errors` attribute. Then we removed the `touched` for simplicity and because I don't think it is something necessary. The GUI felt good either way

frontend/src/components/scan/ScanForm.jsx
devmrfitz: Done. Does it look good?

frontend/src/components/scan/ScanForm.jsx
devmrfitz: Assuming you were referring to error detection, does it work better now?

mlodic:
I tried your change and that works correctly, the GUI feels good now. In the end, the validation of the existence of the list was the main problem.

Just one thing about the GUI: I think it would be better to leverage the space horizontally instead of vertically.

image

The new "Add" button is cool but takes up too much space, an entire row. Considering that, in my opinion, the "trash" button is useless for the first row, could we substitute it with an "Add" button and remove the other one that occupies the entire row?
Also, what do you think about a "dark grey" color for the trash button so it is easier to differentiate it from the other?

mlodic:
I don't get exactly what do you mean for the "difficulty for that index". 
Anyway, my idea was to remove the "trash" button in the first row only to gain the space to insert the "Add" button. The "Trash" button would be kept in all the other added rows. The point is that it makes no sense to add the chance to delete the first row so we could re-use that space for the "Add button"
Tell me if it sounds clearer now

devmrfitz:
image
Over here, if the user realizes that `8.8.8.8` does not need to be sent for scanning, the delete button in the first row becomes necessary

api_app/views.py
devmrfitz: Is the intended result having a single endpoint handle both cases? In that case, one endpoint would have two very different possible response structures. This seems to violate the REST API conventions as all documentation tools expect endpoints' responses to have a fixed structure

api_app/serializers.py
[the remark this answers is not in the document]
devmrfitz: If I override `filter_analyzers_and_connectors` in `FileAnalysisSerializer` and add the check there, won't I still have to keep the 2 calls to `self.filter_analyzers_and_connectors(attrs)` in child classes instead of parent class?

mlodic:
I got your point and makes sense. Still not sure about the entire row used by the "Add new value" button but I don't have any other ideas. For now I think it is ok like that

api_app/serializers.py
devmrfitz: Done

api_app/views.py
0ssigeno:
```python3
def _observable_analysis_request(request,
    serializer_class: ObservableAnalysisSerializer,
    many:bool=True
    # do stuff
    return something
    
@add_docs(
    description="This endpoint allows to start a Job related to an observable. "
    "Retained for compatibility",
    request=ObservableAnalysisSerializer,
    responses={200: AnalysisResponseSerializer},
)
@api_view(["POST"])
def analyze_observable(request):
    res =  _observable_analysis_request(request, ObservableAnalysisSerializer)
    # do parse
    return res


@add_docs(
    description="This endpoint allows to start Jobs related to multiple observables",
    request=ObservableAnalysisSerializer,
    responses={200: AnalysisResponseSerializer},
)
@api_view(["POST"])
def analyze_multiple_observables(request):
    res =  _observable_analysis_request(request, ObservableAnalysisSerializer)    
      # do parse
    return res
    
 ```
 This is the structure of what i expected

File: api_app/serializers.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (0ssigeno) [reply api_app/serializers.py:452]: No, you could just have it  called in the parent class and in the `FileAnalysisSerializer` call `super().validate()` **AFTER** the set up of `attrs["md5"]`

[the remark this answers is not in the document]
Comment (mlodic) [reply api_app/serializers.py:452]: No, you would need to leverage inheritance properly. After having removed the code related to the "file_mimetype" in the `filter_analyzers_and_connectors` you can add it in the overwrite of the same method in the `FileAnalysisSerializer` class:
```
def filter_connectors(...):
    super(FileAnalysisSerializer, ...
    # check mimetype here
```
In this way, first, you leverage the common code, then you perform the check related to files only. This approach could be applied also for observables (`serialized_data["observable_classification"]`)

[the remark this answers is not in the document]
Comment (0ssigeno) [reply api_app/serializers.py:452]:
```python3
class _AbstractJobCreateSerializer(rfs.ModelSerializer):
    """
    Base Serializer for ``Job`` model's ``create()``.
    """

    tags_labels = rfs.ListField(default=list)
    runtime_configuration = rfs.JSONField(required=False, default={}, write_only=True)
    analyzers_requested = rfs.ListField(default=list)
    connectors_requested = rfs.ListField(default=list)
    md5 = rfs.HiddenField(default=None)

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.warnings = []

    def validate(self, attrs: dict) -> dict:
        # check and validate runtime_configuration
        runtime_conf = attrs.get("runtime_configuration", {})
        if runtime_conf and isinstance(runtime_conf, list):
            runtime_conf = json.loads(runtime_conf[0])
        attrs["runtime_configuration"] = runtime_conf
        -------------------- HERE ----------------------
        self.filter_analyzers_and_connectors(attrs)
        return attrs

    @staticmethod
    def filter_analyzers(serialized_data: Dict, warnings: List) -> List[str]:
        # the same but removed the check on file

class FileAnalysisSerializer(_AbstractJobCreateSerializer):
    @staticmethod
    def filter_analyzers(serialized_data: Dict, warnings: List) -> List[str]:
        # added the check on the file
        
    def validate(self, attrs: dict) -> dict:
        logger.debug(f"before attrs: {attrs}")
        # calculate ``file_mimetype``
        attrs["file_mimetype"] = calculate_mimetype(attrs["file"], attrs["file_name"])
        # calculate ``md5``
        file_obj = attrs["file"].file
        file_obj.seek(0)
        file_buffer = file_obj.read()
        attrs["md5"] = calculate_md5(file_buffer)
        ---------------------------------- HERE --------------------------
        attrs = super(FileAnalysisSerializer, self).validate(attrs)
        logger.debug(f"after attrs: {attrs}")
        return attrs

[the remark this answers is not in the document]
Comment (devmrfitz) [reply api_app/serializers.py:452]: In `filter_analyzers_and_connectors`, there is a code snippet of this type:
```
if this_is_a_file:
    # Perform some file type checks (including file_mimetype
else:
    # Perform some observable type checks
```
There are 2 approaches I can follow here.
1) Leave this code block as it is.
2) Remove both, override `filter_analyzers_and_connectors` in both child analyzers and add them there.

Both these approaches would work equally well. Which one would you prefer me to take?
@mlodic @0ssigeno

File: api_app/views.py (hunk 5 above)
Comment (devmrfitz) [reply api_app/views.py]: @0ssigeno Pushed. Does it look good?

File: api_app/serializers.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (0ssigeno) [reply api_app/serializers.py:439]: Imho the secoond approach is the good one

File: api_app/views.py (hunk 5 above)
Comment (0ssigeno) [reply api_app/views.py]: I like it!

File: api_app/serializers.py (hunk 1 above)
Comment (devmrfitz) [reply api_app/serializers.py]: @0ssigeno @mlodic 
Alert: We can't normally use `self.warnings` here. This is because ListSerializer uses the same instance to serialize each of the entries. Hence, they all share a common `self.warnings`.

```
{
    "count": 2,
    "results": [
        {
            "job_id": 80,
            "status": "accepted",
            "warnings": [
                "Robtex_IP_Query won't be run because it does not support observable type domain."
            ],
            "analyzers_running": [
                "Classic_DNS",
                "Robtex_IP_Query"
            ],
            "connectors_running": []
        },
        {
            "job_id": 81,
            "status": "accepted",
            "warnings": [
                "Robtex_IP_Query won't be run because it does not support observable type domain."
            ],
            "analyzers_running": [
                "Classic_DNS"
            ],
            "connectors_running": []
        }
    ]
}
```

Comment (devmrfitz) [reply api_app/serializers.py]: Should I use method 1 now?

Comment (devmrfitz) [reply api_app/serializers.py]: Found a fix

File: api_app/serializers.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (devmrfitz) [reply api_app/serializers.py]: Done

**Comment by devmrfitz** [comment]
@0ssigeno @mlodic Could you help me with `test_analyze_file__corrupted_sample` failing on `api_app/views.py:252` (the first line of `analyze_file`)

File: api_app/views.py [hunk 14]
```diff
@@ -141,81 +203,126 @@ def _analysis_request(
...
-            & Q(analyzers_to_execute__contains=analyzers)
-        )
-
-    if minutes_ago:
-        minutes_ago_time = get_now() - timedelta(minutes=minutes_ago)
-        query = query & Q(received_request_time__gte=minutes_ago_time)
 
+@api_view(["POST"])
+def analyze_file(request):
+    data: QueryDict = request.data.copy()
```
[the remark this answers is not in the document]
devmrfitz: > This is the issue. `TemporaryFile` is not serializable, so the `.copy()` crashes. We could just remove the `.copy()` from here, don't we?

The copy is necessary as we need to manipulate the dictionary. And that can't be done without `.copy()`.
image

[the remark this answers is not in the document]
devmrfitz: If that was the issue, the other file tests would also fail. However, they do succeed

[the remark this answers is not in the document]
0ssigeno: https://code.djangoproject.com/ticket/29510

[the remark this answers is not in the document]
0ssigeno: Let me think about a solution that could with our case!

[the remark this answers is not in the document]
0ssigeno:
```python3

d: QueryDict = request.data
    data : QueryDict = QueryDict()    
    
    for key, value in d.items():
        if key == "file":
            data["files"]= [value]
        elif ...:
            ...
    if not "files" in data:
        raise ValidationError(...)
    
    response = _multi_analysis_request(request.user, data, FileAnalysisSerializer)[
        "results"
    ][0]
    return Response(
        response,
        status=status.HTTP_200_OK,
    )
```
Something like this?

api_app/serializers.py
```diff
@@ -141,14 +312,68 @@ def get_permissions(self, obj: Job) -> Dict[str, bool]:
...
+
+        if data.getlist("file_names", False) and len(data.getlist("file_names")) != len(
+            data.getlist("files")
+        ):
+            raise ValidationError("file_names and files must have the same length.")
+
+        for index, file in enumerate(data.getlist("files")):
+            # `deepcopy` here ensures that this code doesn't
+            # break even if new fields are added in future
+            item = data.copy()
```
0ssigeno, line 344: Same issue

api_app/views.py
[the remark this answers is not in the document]
devmrfitz: Does the fix look good?
@0ssigeno

devmrfitz:
@mlodic @0ssigeno Could you review the PR now?

devmrfitz:
> LGTM, we need to fix the last errors but i see no big issues with the current implementations!

Fixed :+1"

api_app/serializers.py
```diff
@@ -77,18 +83,183 @@ class _AbstractJobCreateSerializer(rfs.ModelSerializer):
...
+                        )
+
+                    if not config.is_observable_type_supported(
+                        serialized_data["observable_classification"]
+                    ):
+                        raise NotRunnableAnalyzer(
+                            f"{a_name} won't be run because "
+                            "it does not support observable type "
+                            f"{serialized_data['observable_classification']}."
+                        )
```
mlodic, line 173: could not we remove this part here considering that this is already handled by the specific versions of `filter_analyzers` of observables/files? so we have not duplicated code

api_app/views.py
```diff
@@ -141,81 +204,134 @@ def _analysis_request(
...
+    return Response(
+        response_dict,
+        status=status.HTTP_200_OK,
+    )
 
 
 @add_docs(
-    description="This endpoint allows to start a Job related to an observable",
+    description="This endpoint allows to start a Job related to an observable. "
+    "Retained for compatibility",
```
mlodic, line 297: "retrocompatibility"

tests/api_app/test_api.py
```diff
@@ -248,3 +308,81 @@ def test_download_sample_400(self):
             content["errors"],
             msg=msg,
         )
+
```
mlodic, line 311: thanks for adding additional tests. This is extremely useful

mlodic:
good job, it seems working well. With this implementation we have also full retrocompatibility with previous existing APIs so old clients can still be supported. Once solved the last things I wrote down we can definitively merge

docs/source/schema.yml
```diff
@@ -1,14 +1,24 @@
 openapi: 3.0.3
 info:
   title: IntelOwl API specification
-  version: 3.4.0
+  version: 3.3.2
```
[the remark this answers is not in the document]
mlodic: ah I think ithat is cause there is the wrong version also somewhere. I made a commit right now to fix that

devmrfitz:
Looks good now?

Issue #978, opened 2022-04-14
Allow plugin secrets to be stored and managed from GUI


eshaan7:
> At the moment, the API keys can be inserted only before starting the project, as environment variables.
It would be useful to let the user set up those keys and the other required configuration directly from the GUI.
This would imply a lot of changes. Also, we have already planned to work on the configuration (https://github.com/intelowlproject/IntelOwl/issues/306) so this idea could be considered as a later development

-- @mlodic in #433.

devmrfitz:
@drosetti  @0ssigeno @mlodic Do we intend to have global analyzer creds (like they are currently), or per-user credentials?

0ssigeno:
In my opinion, we could work in 2 steps: the first one, is to make the secret manageable by the GUI globally, so that we are just adding this new management possibility. After that, we can discuss if we want to have a per-user management, and we have any user case where this is needed.

mlodic:
closed with #1238

PR #1095, opened 2022-07-12
Allow customisation of plugin params from GUI

### LINKED ISSUE #433: Allow editing plugin `params` from the GUI
At the moment, the API keys can be inserted only before starting the project, as environment variables.

It would be useful to let the user set up those keys and the other required configuration directly from the GUI.

This would imply a lot of changes. Also, we have already planned to work on the configuration (#306) so this idea could be considered as a later development

devmrfitz:
# Description

Ability to set user-level and organization-level configuration

## Related issues
closes #433 

## Type of change

# Checklist

### Important Rules

# Real World Example


mlodic (commented):
some important things to keep in mind.

`get_verification` will need to be change and extract secrets no more from the env/AWS but from the DB. `get_secrets` will still be used for all the other non-analyzers/connectors secrets.

Also, there is a special case with a direct call to `MAXMIND_KEY` with the `get_secrets` that must be changed.

Also, remember that the issue you mentioned was splitted. So I think you are working on the #978 instead.

I'll write down the goals of this task from the proposal so it is easier to keep in mind the goals of this task:

```
Since the credentials are stored in env_file_app, which can
not be updated without restarting IntelOwl, the approach I plan to take is to store
these credentials in the database. This will also ensure that the credentials are
relatively secured from accidental exposure (as they are no longer in a simple
text file) and are containerized. To achieve storing in the database, we just need
to create a simple model (in models.py) and create a relevant DRF APIView for
this model. This APIView will not allow users to see the credentials once saved.
The model will also have a field to denote the scope of the credentials (user-level
or organisation-level). Only the organisation owner can edit organisation level
credentials. Now, we just need to add the UI to the frontend and call the created
APIView’s endpoints
```

devmrfitz:
@mlodic Actually I had #433 in mind.
`Create a new model for storing user-level and organization-level default param values. Modify code in api_app.api._analysis_request to read the user’s and organization’s param values from DB and apply them in order of priority. Create a menu in the frontend for users to save user-level params and for organization owners to save organization-level params.`

devmrfitz:
#978 is a separate entry in my timeline

mlodic:
Ah yeah, you are right, I got confused cause at the end the goals are pretty similar. In both cases we should provide to the GUI the ability to change the configuration, either `params` or `secrets`. I think that one can be the base for the other one so the following considerations I was writing and that you can find below are still valid.

Considerations:

We need to think about how to allow editing of the keys/params from the GUI.
One idea could be to add a new button in the Plugin view that could be called "Edit Secrets/Params": that button would open a modal (something like the "Run configuration" modal in the "Scan view"), where the people can change the secrets/params used by the selected analyzer. I think it is convenient for the user to get the chance to perform the change directly there because there you can find all the configuration values.

Then, I think that the secrets can be shown to the user but, by default they should not. We could do something like Eshaan did for the API tokens in the "API Access" section. With a click you can view the key. We could also add the time when it was added. Ex:
image

Then, we should handle there the chance to have both a "user" key and a "organization" key. Only admins can edit/view "organization" keys. If both are present, the user would leverage its own key by default.

devmrfitz:
The json view for params would present a problem as that would mean all params for a given analyzer are a single entry. Now, if we modify the analyzer to add new params, this would create an inconsistency. I was thinking of per-attribute overrides of params

api_app/models.py
```diff
@@ -181,3 +182,19 @@ def delete_file(sender, instance: Job, **kwargs):
...
+
+    type = models.PositiveSmallIntegerField(choices=Type.choices)
+    attribute = models.CharField(max_length=128, blank=False)
+    value = models.CharField(max_length=128, blank=False)
+    organization = models.ForeignKey(
+        Organization, on_delete=models.CASCADE, blank=True, null=True
+    )
+    owner = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.CASCADE, blank=False, null=False
+    )
```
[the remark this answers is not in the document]
devmrfitz: made some changes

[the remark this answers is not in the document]
devmrfitz:
```
name="EmailRep"
attribute="soft_time_limit"
value="90"
```

devmrfitz:
`POST` request at `/api/user-custom-config` with body:
```
{
    "name":"Classic_DNS",
    "attribute":"query_type",
    "value":"\"CNAME\"",
    "type": 1
}
```
gives response
```
{
    "id": 8,
    "owner": 1,
    "type": 1,
    "attribute": "query_type",
    "value": "\"CNAME\"",
    "name": "Classic_DNS",
    "updated_at": "2022-07-16T19:10:29.772962Z",
    "organization": null
}
```
and saves the corresponding `CustomConfig` entry.

api_app/helpers.py
```diff
@@ -130,3 +133,43 @@ def get_hash_type(hash_value):
             detected_hash_type = hash_type
             break
     return detected_hash_type  # stays None if no matches
+
+
+def get_custom_config_as_dict(user, entity_type, name=None) -> dict:
```
mlodic, line 138: I think this should be better placed as a `classmethod` in the corresponding CustomConfig class. In that way we don't separate the logic that leverages that class from the class itself. This is very important for code maintainability, in particular when the project grows up

api_app/helpers.py
```diff
@@ -130,3 +133,43 @@ def get_hash_type(hash_value):
...
+            organization=membership.organization,
+            type=entity_type,
+        )
+        if name is not None:
+            custom_configs = custom_configs.filter(name=name)
+        for custom_config in custom_configs:
+            custom_config: CustomConfig
+            if custom_config.name not in result:
+                result[custom_config.name] = {}
+            result[custom_config.name][custom_config.attribute] = json.loads(
```
mlodic, line 155: the `json.loads` call sounds wrong cause this means that we are erronously saving the `value` attribute as valid JSON while it should just be a simple string

api_app/helpers.py
```diff
@@ -130,3 +133,43 @@ def get_hash_type(hash_value):
...
             break
     return detected_hash_type  # stays None if no matches
+
+
+def get_custom_config_as_dict(user, entity_type, name=None) -> dict:
+    """
+    Returns custom config as dict
+    """
+    result = {}
+    try:
```
mlodic, line 143: while your intentions are clear to me, it would be useful to add more context (comments) to help the reader to understand why you check the org first

api_app/helpers.py
```diff
@@ -130,3 +133,43 @@ def get_hash_type(hash_value):
...
+    if name is not None:
+        custom_configs = custom_configs.filter(name=name)
+    for custom_config in custom_configs:
+        custom_config: CustomConfig
+        if custom_config.name not in result:
+            result[custom_config.name] = {}
+        result[custom_config.name][custom_config.attribute] = json.loads(
+            custom_config.value
+        )
+
```
mlodic, line 174: all of this code is mostly duplicated so please try to re-arrange this

api_app/helpers.py
mlodic: please add also logs in critical parts of this function

api_app/analyzers_manager/views.py
```diff
@@ -46,6 +48,16 @@ class AnalyzerListAPI(APIView):
     def get(self, request, *args, **kwargs):
         try:
             ac = self.serializer_class.read_and_verify_config()
+            custom_configs = get_custom_config_as_dict(
+                request.user, CustomConfig.Type.ANALYZER
+            )
+            for analyzer in ac.values():
```
mlodic, line 54: while your intentions are clear to me, it would be useful to add more context (comments) to help the reader to understand what is happening here

api_app/views.py
```diff
@@ -574,3 +587,37 @@ class TagViewSet(viewsets.ModelViewSet):
     queryset = Tag.objects.all()
     serializer_class = TagSerializer
     pagination_class = None
+
+
+class UserCustomConfigViewSet(viewsets.ModelViewSet):
```
mlodic, line 592: please add `@add_docs`

api_app/views.py
```diff
@@ -574,3 +587,37 @@ class TagViewSet(viewsets.ModelViewSet):
...
+            )
+        except CustomConfig.DoesNotExist:
+            return super().create(request, *args, **kwargs)
+        else:
+            raise ValidationError(
+                {
+                    "detail": "A custom config for this entity "
+                    "already exists with same scope."
+                }
+            )
```
mlodic, line 623: ok but this is a validation so this should be implemented in the related serializer

api_app/views.py
```diff
@@ -82,7 +83,19 @@ def _multi_analysis_request(
...
+            for analyzer in job.analyzers_to_execute:
+                # Appending custom config to runtime configuration
+                config = get_custom_config_as_dict(
+                    user, CustomConfig.Type.ANALYZER, name=analyzer
+                ).get(analyzer, {})
+                if analyzer in runtime_configurations[index]:
+                    config = config | runtime_configurations[index][analyzer]
+                if config:
+                    runtime_configuration[analyzer] = config
+            print(runtime_configuration, type(job.analyzers_to_execute[0]))
```
mlodic, line 97: -> logger.

api_app/views.py
```diff
@@ -82,7 +83,19 @@ def _multi_analysis_request(
...
+
+            for analyzer in job.analyzers_to_execute:
+                # Appending custom config to runtime configuration
+                config = get_custom_config_as_dict(
+                    user, CustomConfig.Type.ANALYZER, name=analyzer
+                ).get(analyzer, {})
+                if analyzer in runtime_configurations[index]:
+                    config = config | runtime_configurations[index][analyzer]
+                if config:
+                    runtime_configuration[analyzer] = config
```
mlodic, line 96: this is complex logic that requires proper testing. So please extract this part to a separate function and do the tests of what you want that this would behave.

api_app/analyzers_manager/views.py
mlodic: also, we need a test for this logic too which proves that it behaves accordingly

api_app/serializers.py
```diff
@@ -592,3 +597,74 @@ class EnvelopeSerializer(rfs.Serializer):
...
+
+class CustomConfigSerializer(rfs.ModelSerializer):
+    owner = PrimaryKeyRelatedField(queryset=User.objects.all(), required=False)
+
+    def validate_value(self, value):
+        try:
+            json.loads(value)
+        except json.JSONDecodeError:
+            raise ValidationError("Value is not JSON-compliant.")
+        return value
```
mlodic, line 610: why is this necessary?

api_app/serializers.py
```diff
@@ -592,3 +597,74 @@ class EnvelopeSerializer(rfs.Serializer):
...
+                    .params[attrs["attribute"]]
+                    .value
+                )
+                raise ValidationError(
+                    f"Connector {attrs['name']} attribute "
+                    f"{attrs['attribute']} has wrong type "
+                    f"{type(json.loads(attrs['value']))}. Expected: "
+                    f"{expected_type}."
+                )
+        return attrs
```
mlodic, line 666: please extract the duplicated code and leverage the right classes for each case

mlodic (changes requested):
when in doubt, please add one more log and/or comment. We will also need tests for all of these changes

api_app/helpers.py
devmrfitz: The json conversion is done as the param value it is overriding is guaranteed to be a JSON-compliant value. This allows us to store all the JSON-compliant types in the StringField

api_app/serializers.py
devmrfitz: The same explanation as the `json.loads` one

api_app/helpers.py
mlodic: Even if JSON is a string, StringField/CharField is not made to store JSON data. 
Anyway I don't even understand why you are using JSON instead of simple string for "value".

We have a structure like this:
```
 "Classic_DNS": {
    "type": "observable",
    "python_module": "dns.dns_resolvers.classic_dns_resolver.ClassicDNSResolver",
    "description": "Retrieve current domain resolution with default DNS",
    "disabled": false,
    "external_service": true,
    "leaks_info": false,
    "observable_supported": ["ip", "domain", "url"],
    "config": {
      "soft_time_limit": 30,
      "queue": "default"
    },
    "secrets": {},
    "params": {
      "query_type": {
        "value": "A",
        "type": "str",
        "description": "Query type against the chosen DNS resolver."
      }
    }
  }
```

The "value" attribute in CustomConfig should just be a simple string "A" in this case or "CNAME" in the example that you made.

mlodic: you got some problems on how you handle that parameter:

looking at what you have written, why CNAME is escaped and containers a string inside a string?
```
    "value": "\"CNAME\"",

```

this should just be
```
 "value": "CNAME",
```

devmrfitz: Let's say a param has type str and the value in it is `5`. Now, we'll store `"5"` in CustomConfig.value. However, if it is of type int and the value in it is `5`, we'll store `5`

devmrfitz: So, when I json.loads it, it will return the correct data type

mlodic: Ok now it starts to make sense why you did that but:
1 - that was completely implicit
2 - if we need to make data conversion, we could do that in another way. Here we are abusing JSON to get the result that we want. It is not meant to be used to do such kind of things. This is an hack we should avoid.

I think we should leverage what was already built inside IntelOwl to manage this properly. De facto what we want is to store a new configuration for a specific parameter in the analyzer config file. But we already have a Serializer for that config file (AbstractConfigSerializer).
If you check [here](https://github.com/intelowlproject/IntelOwl/blob/master/api_app/core/serializers.py#L95), there is already a `ParamSerializer` used for that structure that we want to override.

The right way to do that could be: first accept a value from the API. Verify that this value is complaint with the existing structure (if `type` is `str` it must be a string) and then save it as a BaseField (see [here](https://github.com/intelowlproject/IntelOwl/blob/master/api_app/core/serializers.py#L51)).
Then we wouldn't ever have any other problems with that parameter cause we can overwrite it here and there securely because we already managed the verification in the serializer properly so the data is always valid.

api_app/helpers.py
devmrfitz: Done

api_app/helpers.py
devmrfitz: Done

api_app/views.py
devmrfitz: Done

api_app/helpers.py
devmrfitz: Let's say we receive a `5` as a value by a `form-data` (or `x-www-form-payload type request. Unlike a `JSON` payload, there is no type information in a `form-data`, it's always a string.
image
Hence, value_type will always be str.

So now, we need to get the type of this analyzer attribute from the config file, then cast the payload to the said format and handle errors. Moreover, if (in the future) we want to support object-type values, we'll need to stringify them. 

However, considering that the `analyzer-config` file itself is a JSON file, it is a guarantee that the JSON-parsing mechanism will always continue to work.

mlodic: I am not getting why we should received a `form-data` payload in this API. All the other endpoints work by sending and receiving JSON. `Content-Type` of both answers and requests is always `application/json`.

I mean, you wrote this example:
```
POST request at /api/custom-config with body:

{
    "name":"Classic_DNS",
    "attribute":"query_type",
    "value":"\"CNAME\"",
    "type": 1
}
```

Where the payload is clearly JSON but the `value` is JSON itself. That's what I don't understand

api_app/models.py
```diff
@@ -181,3 +188,60 @@ def delete_file(sender, instance: Job, **kwargs):
...
+        """
+        Returns custom config as dict
+        """
+        custom_configs = QuerySet(cls)
+
+        # Since, user-level custom configs should override organization-level configs,
+        # we need to get the organization-level configs, if any, first.
+        try:
+            membership = Membership.objects.get(user=user)
+            custom_configs = custom_configs | cls.objects.filter(
```
mlodic, line 221: cool usage of bitwise operator for dictionaries. But we can make it even more cool. Try `|=` to avoid to repeat `custom_configs`. Please use that also in the other parts where you leveraged with operator.

api_app/serializers.py
```diff
@@ -38,8 +40,11 @@
     "ObservableAnalysisSerializer",
     "AnalysisResponseSerializer",
     "multi_result_enveloper",
+    "CustomConfigSerializer",
 ]
 
+# User = get_user_model()
```
mlodic, line 46: forgotten here?

api_app/views.py
```diff
@@ -82,7 +85,19 @@ def _multi_analysis_request(
...
+            for analyzer in job.analyzers_to_execute:
+                # Appending custom config to runtime configuration
+                config = CustomConfig.get_as_dict(
+                    user, CustomConfig.Type.ANALYZER, name=analyzer
+                ).get(analyzer, {})
+                if analyzer in runtime_configurations[index]:
+                    config = config | runtime_configurations[index][analyzer]
+                if config:
+                    runtime_configuration[analyzer] = config
+            logger.debug(runtime_configuration, type(job.analyzers_to_execute[0]))
```
mlodic, line 99: cool but there is no normal text. This is of critical importance when you need to search through your logs for the all the logs generated by this line. It is enough to write few words with this

api_app/views.py
```diff
@@ -574,3 +589,41 @@ class TagViewSet(viewsets.ModelViewSet):
     queryset = Tag.objects.all()
     serializer_class = TagSerializer
     pagination_class = None
+
+
+class CustomConfigViewSet(viewsets.ModelViewSet):
```
mlodic, line 594: please add @add_docs

api_app/views.py
```diff
@@ -82,7 +85,19 @@ def _multi_analysis_request(
...
+
+            for analyzer in job.analyzers_to_execute:
+                # Appending custom config to runtime configuration
+                config = CustomConfig.get_as_dict(
+                    user, CustomConfig.Type.ANALYZER, name=analyzer
+                ).get(analyzer, {})
+                if analyzer in runtime_configurations[index]:
+                    config = config | runtime_configurations[index][analyzer]
+                if config:
+                    runtime_configuration[analyzer] = config
```
mlodic, line 98: this is complex logic that requires proper testing. So please extract this part to a separate function and do the tests of what you want that this would behave.

api_app/serializers.py
```diff
@@ -592,3 +597,77 @@ class EnvelopeSerializer(rfs.Serializer):
...
+                self.instance = CustomConfig.objects.get(**params)
+                logger.info(f"CustomConfig {self.instance} found. Will be updated.")
+            except CustomConfig.DoesNotExist:
+                # No matching CustomConfig was found. A new entry will be created.
+                pass
+        return attrs
+
+    class Meta:
+        model = CustomConfig
+        fields = rfs.ALL_FIELDS
```
mlodic, line 673: that's cool, could you just move it at the start of the class CustomConfigSerializer? In that way a reader would read this first so he can understand which fields are available in this serializer before reading the rest of the code

api_app/serializers.py
```diff
@@ -592,3 +597,77 @@ class EnvelopeSerializer(rfs.Serializer):
...
+                raise ValidationError("User is not owner of the organization.")
+
+        config = None
+        category = None
+        if attrs["type"] == CustomConfig.Type.ANALYZER:
+            config = AnalyzerConfig
+            category = "Analyzer"
+        elif attrs["type"] == CustomConfig.Type.CONNECTOR:
+            config = ConnectorConfig
+            category = "Connector"
```
mlodic, line 631: cool optimization. could you also handle the impossible `else` case with a critical error log?

api_app/serializers.py
```diff
@@ -592,3 +597,77 @@ class EnvelopeSerializer(rfs.Serializer):
...
+        # check if owner is admin of organization
+        if attrs.get("organization", None):
+            # check if the user is owner of the organization
+            membership = Membership.objects.filter(
+                user=attrs.get("owner"),
+                organization=attrs.get("organization"),
+                is_owner=True,
+            )
+            if not membership.exists():
+                raise ValidationError("User is not owner of the organization.")
```
mlodic, line 622: could you also log the user name and the organization name here?

api_app/helpers.py
devmrfitz: Yeah, you're right. Currently, we're using `application/json` everywhere (except file uploads). But, in the future, if we want to mix this functionality with a file upload somehow, we would be unable to. So, I just wanted to be able to support it if needed in future.
Additionally, (for param values of types like `array` or `object`), we will need to stringify them in future. Thus, those doors are kept open too if we use JSON

api_app/serializers.py
```diff
@@ -592,3 +597,77 @@ class EnvelopeSerializer(rfs.Serializer):
...
+                f"{attrs['attribute']} has wrong type "
+                f"{type(json.loads(attrs['value'])).__name__}. Expected: "
+                f"{expected_type.__name__}."
+            )
+
+        if self.instance is None:
+            try:
+                params = attrs.copy()
+                params.pop("value")
+                print("params:", params)
```
mlodic, line 663: logger.debug?

api_app/serializers.py
```diff
@@ -592,3 +597,77 @@ class EnvelopeSerializer(rfs.Serializer):
...
+                self.instance = CustomConfig.objects.get(**params)
+                logger.info(f"CustomConfig {self.instance} found. Will be updated.")
+            except CustomConfig.DoesNotExist:
+                # No matching CustomConfig was found. A new entry will be created.
+                pass
```
mlodic, line 668: Ok this is to update an already existing entry with the new value, am I right? Is this necessary?

devmrfitz: This eliminates the need for front end to send PATCH requests. If a user has a lot of configs, they don't first have to find if a matching one exists, they can directly overwrite it.

api_app/serializers.py
```diff
@@ -5,12 +5,14 @@
 import logging
 from typing import Dict, List
 
+# from django.contrib.auth import get_user_model
```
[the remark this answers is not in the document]
devmrfitz: Left for easy future use. Should I remove?

api_app/serializers.py
devmrfitz: Left for easy future use. Should I remove?

api_app/helpers.py
mlodic: thanks for the explanation. 

I don't think that this specific API endpoint would ever support that. And even if that will happen, we'll make changes then. Right now I think it is better to keep things simple there.

Then the point regarding having array types that is true. We already have `type:list` params. I don't think we'll ever have `type:dict` because in that case it would make sense to just create another simple param.
Anyway, this was already handled previously by @eshaan7 when writing the code of the `AbstractConfigSerializer` that I linked you before. He leveraged the `BaseField` that basically allow to write every kind of type inside of it.
So, if the request payload is valid JSON and the analyzer config is valid JSON, then why both cases shouldn't be handled the same way?

In `CustomConfig`,
`type` could be `rfs.ChoiceField(choices=PARAM_DATATYPE_CHOICES)` and
`value` could be `BaseField()`

like it is in the `_ParamSerializer` serializer

api_app/models.py
```diff
@@ -181,3 +188,60 @@ def delete_file(sender, instance: Job, **kwargs):
...
+    type = models.PositiveSmallIntegerField(choices=Type.choices)
+    attribute = models.CharField(max_length=128, blank=False)
+    value = models.CharField(max_length=128, blank=False)
+    organization = models.ForeignKey(
+        Organization, on_delete=models.CASCADE, blank=True, null=True
+    )
+    owner = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.CASCADE, blank=False, null=False
+    )
+    name = models.CharField(max_length=128, blank=False)
```
mlodic, line 207: `plugin_name` is more understandable

api_app/views.py
devmrfitz: Most of the heavy lifting in done in `get_as_dict`. It's just 2 `if` conditions after that. Should I still attempt the refactor?

api_app/serializers.py
mlodic: nah, not needed

api_app/models.py
```diff
@@ -181,3 +188,60 @@ def delete_file(sender, instance: Job, **kwargs):
...
+
+    @classmethod
+    def get_as_dict(cls, user, entity_type, name=None) -> dict:
+        """
+        Returns custom config as dict
+        """
+        custom_configs = QuerySet(cls)
+
+        # Since, user-level custom configs should override organization-level configs,
+        # we need to get the organization-level configs, if any, first.
```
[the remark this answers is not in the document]
devmrfitz: Thank you :)

api_app/models.py
```diff
@@ -181,3 +188,60 @@ def delete_file(sender, instance: Job, **kwargs):
     if instance.file:
         if instance.file:
             instance.file.delete()
+
+
+class CustomConfig(models.Model):
+    class Type(models.IntegerChoices):
```
mlodic, line 194: we could call this `PluginType`

api_app/models.py
devmrfitz: Makes sense. However, I've used it in quite a few places now. Should I perform the refactor?

api_app/models.py
devmrfitz: done

api_app/views.py
mlodic: yeah, my point is to extract this to allow it to be unit-tested. Otherwise it is just not possible.

I mean, we need to test most the new things that we do but when we do changes to already existing APIs we should be careful even more. This is why extracting this and creating a simple unit test which proves that it works as you intend it to and that the normal behavior is preserved, would really help us in maintaining the application.

api_app/models.py
mlodic: yeah please, I mean, it lasts more this exchange of words than the refactor itself :P

api_app/helpers.py
devmrfitz: BaseField is not a valid db type. That's what I meant by "we'll need to stringify"... and destringify i.e. `json.dumps` and `json.loads`.

api_app/serializers.py
mlodic: What you said is clever and makes sense. Nevertheless this is a bit hacky, you know? We are basically circumventing the REST API rules to avoid to an additional HTTP request. Even if I appreciate the effort, I still prefer to follow the framework and the normal rules. It makes sense that if the config never existed then we use a PUT (`update`) and when it already exists we use a PATCH (`partial_update`). In that way we are complaint

api_app/models.py
devmrfitz: done

api_app/serializers.py
devmrfitz: If the config never existed, won't it be a `POST`?

devmrfitz: Also, so an error if a POST has a matching entry?

api_app/helpers.py
mlodic: ah you are right, damn the difference is that the analyzer config is not in the real DB so that association I pointed out is not possible.

ok so another option that could be more clear (I would like to get rid of that explicit json conversions if possible) is to use the Django [JSONField](https://docs.djangoproject.com/en/4.0/ref/models/fields/#jsonfield). What do you think?

api_app/serializers.py
mlodic: yeah you are right, I meant that, God I am melted 🤣  POST `create` 

Then yeah I think yes because after the first time we set it, we only want to update the value, nothing else, am I right?

api_app/views.py
devmrfitz: I could do that. But databases are widely varying in how they handle JSON Fields, and the explicit conversions will still remain (for type-checking purposes).

api_app/helpers.py
devmrfitz: I could do that. But databases are widely varying in how they handle JSON Fields, and the explicit conversions will still remain (for type-checking purposes).

api_app/serializers.py
devmrfitz: `DELETE` too, but that's anyways handled

api_app/models.py
```diff
@@ -181,3 +188,60 @@ def delete_file(sender, instance: Job, **kwargs):
...
+        """
+        custom_configs = QuerySet(cls)
+
+        # Since, user-level custom configs should override organization-level configs,
+        # we need to get the organization-level configs, if any, first.
+        try:
+            membership = Membership.objects.get(user=user)
+            custom_configs = custom_configs | cls.objects.filter(
+                organization=membership.organization,
+                type=entity_type,
```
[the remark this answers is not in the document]
devmrfitz: Done. Looks good?

api_app/models.py
```diff
@@ -181,3 +188,80 @@ def delete_file(sender, instance: Job, **kwargs):
...
+    )
+    plugin_name = models.CharField(max_length=128, blank=False)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        # unique_together
+        # index
+
+        constraints = [
+            models.UniqueConstraint(
```
mlodic, line 215: good thing this one.

Ah, remember that you'll need to merge the migrations in a single one at the end. Like we do for the commits in Git.

api_app/helpers.py
mlodic: yeah but if we are saving this as a valid JSON I think it makes sense to explicitly use this field considering that it is supported

devmrfitz: Alright. So everything else remains same, just the StringField changes to a JSONField, right?

devmrfitz:
@mlodic Could you re-review the backend as well as frontend?

devmrfitz:
image
image
image

mlodic:
The GUI seems cool :) I'll check closely later in the day

api_app/analyzers_manager/views.py
```diff
@@ -46,6 +47,7 @@ class AnalyzerListAPI(APIView):
     def get(self, request, *args, **kwargs):
         try:
             ac = self.serializer_class.read_and_verify_config()
+            CustomConfig.apply(ac, request.user, CustomConfig.PluginType.ANALYZER)
```
mlodic, line 50: cool, in this way you can do a unit test for this function.

* choose a possible input
* call the function
* check if the data generated/changes is the same of what you expect (use `self.assert`-like clauses from the `unittest` framework`

mlodic:
while the GUI is cool and while the separation between user config and org config make sense, it doesn't if we think about all the project and what already exists. I would like to underline what I shared with you beforehand:

> We need to think about how to allow editing of the keys/params from the GUI.
One idea could be to add a new button in the Plugin view that could be called "Edit Secrets/Params": that button would open a modal (something like the "Run configuration" modal in the "Scan view"), where the people can change the secrets/params used by the selected analyzer. I think it is convenient for the user to get the chance to perform the change directly there because there you can find all the configuration values.

This is important. I mean, we already have a "plugin" section and the users go there when they want to check the analyzers configuration. Adding a new section to perform only this kind of thing does not allow a good user experience. I mean, you implemented this without even having tried to discuss this before, while I clearly stated my point.

So, right now I don't see any valid reason to do not put this feature directly in the "Plugin" section. But I am available to hear other point of views of course.

devmrfitz:
> while the GUI is cool and while the separation between user config and org config make sense, it doesn't if we think about all the project and what already exists. I would like to underline what I shared with you beforehand:
> 
> > We need to think about how to allow editing of the keys/params from the GUI.
> > One idea could be to add a new button in the Plugin view that could be called "Edit Secrets/Params": that button would open a modal (something like the "Run configuration" modal in the "Scan view"), where the people can change the secrets/params used by the selected analyzer. I think it is convenient for the user to get the chance to perform the change directly there because there you can find all the configuration values.
> 
> This is important. I mean, we already have a "plugin" section and the users go there when they want to check the analyzers configuration. Adding a new section to perform only this kind of thing does not allow a good user experience. I mean, you implemented this without even having tried to discuss this before, while I clearly stated my point.
> 
> So, right now I don't see any valid reason to do not put this feature directly in the "Plugin" section. But I am available to hear other point of views of course.

That does make sense. It would be more convenient if one wants to override specific params. However, if an org admin wants to go and check all the overrides that have been applied, and say, remove them, it would be really inconvenient to click on every plugin to check for overrides. Hence, I implemented a central "override panel"

mlodic:
> That does make sense. It would be more convenient if one wants to override specific params. However, if an org admin wants to go and check all the overrides that have been applied, and say, remove them, it would be really inconvenient to click on every plugin to check for overrides. Hence, I implemented a central "override panel". And a similar situation for an individual user.
Moreover I considered the user-org separation to be necessary to ensure someone trying to apply, modify or delete a user-level override does not accidentally affect an entire organization. In the current implementation, the only way for a person to modify org-level config is to actually go through the org page, hence reducing accidents.

Thank you very much for your point of view. This is how we can get the best results. Combining our ideas together.

Everything you said makes sense to me. I think that a compromise must be chosen because both options are not perfect. But I think there's no alternatives.

Saying that, I think that we can keep your implementation. However we could add on top of it something else.
For example, first thing that comes to my mind is that we could add a link to the "Your custom config" and the "Org config" from the "Plugin" section. The second link would be accessible only to org users. In this way it is obvious to the users who go in the "Plugin" section that they need to go in the new sections to perform changes to the configuration.

devmrfitz:
> The second link would be accessible only to org users. 
only to org owners, right?

api_app/views.py
```diff
@@ -574,3 +597,47 @@ class TagViewSet(viewsets.ModelViewSet):
...
+        membership = Membership.objects.filter(
+            user=self.request.user, organization__isnull=False
+        )
+        if membership.exists():
+            result |= CustomConfig.objects.filter(
+                organization=membership[0].organization
+            )
+
+        # Adding CustomConfigs for user
+        result |= CustomConfig.objects.filter(owner=self.request.user)
```
0ssigeno, line 627: Why do you need to initialize this to an empty and use the `or` ? 
```suggestion
        # Adding CustomConfigs for user's organization, if any
        membership = Membership.objects.filter(
            user=self.request.user, organization__isnull=False
        )
        if membership.exists():
            result = CustomConfig.objects.filter(
                organization=membership[0].organization
            )
            else:

               result = CustomConfig.objects.filter(owner=self.request.user)
```

api_app/views.py
```diff
@@ -77,12 +80,32 @@ def _multi_analysis_request(
     logger.info(f"New Jobs added to queue <- {repr(jobs)}.")
 
     # Check if task is test or not
-    if settings.STAGE_CI:
+    if settings.STAGE_CI and not settings.FORCE_SCHEDULE_JOBS:
```
0ssigeno, line 83: Instead of having another variable and to run a job that you don't actually need to run (because, correct me if I'm wrong, but you want in the CI to verify the configuration), i think is a better solution to **always** do the configuration in the tests (hoping that this will not break other tests) and, `if not settings.STAGE_CI`, run the celery task.
In summary, I would remove this new settings variable, that we have already too many, and move the check CI/not CI before calling `celery_app.send_task`

0ssigeno (changes requested):
I won't comment any structural, since Matteo already approved everything that you did and I agree with its review. I just added small some opinion on small things that should not really impact anything important. GG on the tests and the code coverage!

api_app/views.py
devmrfitz: I need the actual `celery_app.send_task` to be executed, too, as only then would my `before_task` get triggered. So, stopping the call would prevent the test from running

api_app/views.py
devmrfitz: Because the first `or` is conditional in nature, while the other isn't.

0ssigeno: Fuck, i can't even read today. You are 100% right

api_app/views.py
0ssigeno: Ok, read the construct that you made, and even if I would have preferred to using the mockup package to mock `start_analyzer` and then verify that it was called with the correct args/kwargs, i do not actually dislike your solution. I think is very creative and imho we could adopt the approach even in other parts of the test suite to verify that the celery tasks have the correct arguments.

0ssigeno (approved):

api_app/analyzers_manager/views.py
0ssigeno: @devmrfitz I just checked the usages of `CustomConfig.apply` and I did not find anything on tests for that. Can you make a test as Matteo asked? After that, I think we are good to go with this PR

drosetti (changes requested):
The GUI is nice, but we have to improve the user experience:

* The user musts write JSON code, example: shodan_analysis field of Shodan_Honeyscore requires a string, and the user musts write something like `"value"`. Also, in case you store a value and then edit it, the value is saved but the user has to add again the `"`. This check should be performed via a function, in this way the user can simply write the value without thinking about the field's format (string, bool,...).
* I noticed that "runtime configuration" in Scan section updates its value with the custom config and org config and I like it! Unfortunately, sometimes it seems to happen a caching/update error and doesn't works properly: with both "custom configuration" and "organization custom config" if you add a value and you move to the Scan section you have to reload the scan page to see the custom value as default in the "runtime configuration" section.
* I also noticed a little problem in both "custom configuration" and "organization custom config" sections: after you delete an entry you can see the toast correctly shown, but the user needs to refresh the page to make the entry or the entry remains available in the GUI. I'm sorry, but this time I noticed this problem, but I didn't find the steps required to replicate it.

api_app/views.py
devmrfitz: Thanks a lot :)

devmrfitz:
> The GUI is nice, but we have to improve the user experience:
> 
>     * The user musts write JSON code, example: shodan_analysis field of Shodan_Honeyscore requires a string, and the user musts write something like `"value"`. Also, in case you store a value and then edit it, the value is saved but the user has to add again the `"`. This check should be performed via a function, in this way the user can simply write the value without thinking about the field's format (string, bool,...).
> 
>     * I noticed that "runtime configuration" in Scan section updates its value with the custom config and org config and I like it! Unfortunately, sometimes it seems to happen a caching/update error and doesn't works properly: with both "custom configuration" and "organization custom config" if you add a value and you move to the Scan section you have to reload the scan page to see the custom value as default in the "runtime configuration" section.
> 
>     * I also noticed a little problem in both "custom configuration" and "organization custom config" sections: after you delete an entry you can see the toast correctly shown, but the user needs to refresh the page to make the entry or the entry remains available in the GUI. I'm sorry, but this time I noticed this problem, but I didn't find the steps required to replicate it.

I've solved all the data out-of-sync issues. 
As for the input in JSON, I think we'll need to maintain that since we also need complex inputs like JSON, and you need the quotes to know the difference between [5] and ["5"], since they are both valid lists but can have very different results based on the analyzer.

> Also, in case you store a value and then edit it, the value is saved but the user has to add again the `"`.

Fixed this

api_app/analyzers_manager/views.py
devmrfitz: Done

0ssigeno (approved):

drosetti (changes requested):
You are right, some analyzers require JSON data, so it's needed the JSON input. I think it could be useful something that helps the user to write the values in correct format. A placeholder in the input element could be enough, but I'd like to see you point of view :)

I still have the sync problems, but only with the organization config:
* If I add to custom organization config and go in the Scan section I need to reload the page to make the update effective in the "runtime configuration".
* If I add a custom organization config, delete it and change page, when I come back to the organization config page I still see the config. It's required to refresh the page to make it disappear.

I found a bug: If you select shodan_honeyscore analyzer and set the value of the filed shodan_analysis to ["0"] or something that is not a string (the expected value type) the server response with a 500.

devmrfitz:
> You are right, some analyzers require JSON data, so it's needed the JSON input. I think it could be useful something that helps the user to write the values in correct format. A placeholder in the input element could be enough, but I'd like to see you point of view :)
> 
> I still have the sync problems, but only with the organization config:
> 
>     * If I add to custom organization config and go in the Scan section I need to reload the page to make the update effective in the "runtime configuration".
> 
>     * If I add a custom organization config, delete it and change page, when I come back to the organization config page I still see the config. It's required to refresh the page to make it disappear.
> 
> 
> I found a bug: If you select shodan_honeyscore analyzer and set the value of the filed shodan_analysis to ["0"] or something that is not a string (the expected value type) the server response with a 500.

I forgot to implement the changes onto org config. My apologies

devmrfitz:
> 

The placeholder sounds good. Just one issue: Let's say analyzer A needs a List of str and B needs a list of int. Now, all we know is that they both need a `list`. Do we show the same placeholder for them? If yes, then what?

devmrfitz:
> I found a bug: If you select shodan_honeyscore analyzer and set the value of the filed shodan_analysis to ["0"] or something that is not a string (the expected value type) the server response with a 500.

Thanks :). Fixed and wrote tests for it

devmrfitz:
@drosetti Does this look good?

frontend/src/components/me/config/Config.jsx
```diff
@@ -0,0 +1,296 @@
...
+    retrieveConnectorsConfiguration();
+  };
+
+  return (
+    <Container>
+      <h4>Your custom configuration</h4>
+      <Loader
+        render={() => (
+          <Formik initialValues={{ config: respData }} onSubmit={null}>
+            {({ values, setFieldValue }) => (
```
drosetti, line 104: I known "value" is a "standard" name, but I think something like "configurations"  is easier to understand

frontend/src/components/me/config/Config.jsx
```diff
@@ -0,0 +1,296 @@
...
+              <Form>
+                <FieldArray name="config">
+                  {({ remove, push }) => (
+                    <FormGroup row>
+                      <Col>
+                        {values.config && values.config.length > 0
+                          ? values.config.map((item, index) => {
```
drosetti, line 111: same logic of the comment above, please change "item" with "plugin"

drosetti (changes requested):
I focused on the frontend and I like a lot the GUI and the UX, (thanks for the fixes!), you did a good job, but I have some suggestions:
* `Config.jsx` and `OrgConfig.jsx` are too similar: we should have a single component with the common code and extend it, in this way we avoid copy-paste. React classes can help you to achieve this goal.
* I think the migration should have a name related to this feature

drosetti:
> @drosetti Do the tests look good?

Yes the user experience is very good! :D

drosetti:
> > You are right, some analyzers require JSON data, so it's needed the JSON input. I think it could be useful something that helps the user to write the values in correct format. A placeholder in the input element could be enough, but I'd like to see you point of view :)
> 
> The placeholder sounds good. Just one issue: Let's say analyzer A needs a List of str and B needs a list of int. Now, all we know is that they both need a `list`. Do we show the same placeholder for them? If yes, then what?

This is a very good question (and I forgot to answer to you in the review, my apologies).
I think you can put and empty list in the placeholder ([] or [...]). Maybe is not the best solution, but I found only very complex solutions and I think they don't worth the effort.

What do you think ?

frontend/src/components/me/config/Config.jsx
```diff
@@ -0,0 +1,296 @@
...
+export function filterEmptyParams(plugins) {
+  const objectFilter = (obj, predicate) =>
+    Object.keys(obj)
+      .filter((key) => predicate(obj[key]))
+      .reduce((res, key) => Object.assign(res, { [key]: obj[key] }), {});
+  return objectFilter(
+    plugins,
+    (item) => item.params && Object.keys(item.params).length > 0
+  );
+}
```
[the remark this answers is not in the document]
devmrfitz: Done. I'd made the `objectFilter` function as I thought that it could be a generic code snippet that could be reused in other parts of the codebase.

frontend/src/components/me/config/Config.jsx
devmrfitz: done

frontend/src/components/me/config/Config.jsx
devmrfitz: done

devmrfitz:
> > > You are right, some analyzers require JSON data, so it's needed the JSON input. I think it could be useful something that helps the user to write the values in correct format. A placeholder in the input element could be enough, but I'd like to see you point of view :)
> > 
> > 
> > The placeholder sounds good. Just one issue: Let's say analyzer A needs a List of str and B needs a list of int. Now, all we know is that they both need a `list`. Do we show the same placeholder for them? If yes, then what?
> 
> This is a very good question (and I forgot to answer to you in the review, my apologies). I think you can put and empty list in the placeholder ([] or [...]). Maybe is not the best solution, but I found only very complex solutions and I think they don't worth the effort.
> 
> What do you think ?

Sounds good. Done

devmrfitz:
> I focused on the frontend and I like a lot the GUI and the UX, (thanks for the fixes!), you did a good job, but I have some suggestions:
> 
>     * `Config.jsx` and `OrgConfig.jsx` are too similar: we should have a single component with the common code and extend it, in this way we avoid copy-paste. React classes can help you to achieve this goal.
> 
>     * I think the migration should have a name related to this feature

Done

drosetti (changes requested):
Good, I think there are few changes and then the pr is done:

* Please change the the migration name with something related to this feature.

frontend/src/components/me/config/UserConfig.jsx
```diff
@@ -0,0 +1,332 @@
...
+                              plugin.attribute &&
+                              plugins[plugin.plugin_name].params[
+                                plugin.attribute
+                              ]
+                            ) {
+                              const { type } =
+                                plugins[plugin.plugin_name].params[
+                                  plugin.attribute
+                                ];
+                              if (type === "str") placeholder = '"string"';
```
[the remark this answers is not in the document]
devmrfitz: Removing the `"` would lead to inconsistency. For instance, in a list, you'll have to write `["5", 5]` to differentiate between strings. Hence in some places, strings would have a `"`, whereas it won't in other places.

Moreover, if we support a generic `JSON` type param in the future, we'll need to use the `"` for sure.

> I think you should check in the backend the type (same as int).

I'm checking all types in the backend but also ensuring proper JSON formatting for all of them.

frontend/src/components/me/config/UserConfig.jsx
```diff
@@ -0,0 +1,332 @@
...
+                              else if (type === "int") placeholder = "1234";
+                              else if (type === "float") placeholder = "12.34";
+                              else if (type === "bool") placeholder = "true";
+                              else if (type === "json")
+                                placeholder = '{"key": "value"}';
```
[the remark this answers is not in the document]
devmrfitz: This is just included for future support

frontend/src/components/me/config/UserConfig.jsx
```diff
@@ -0,0 +1,332 @@
...
+                              else if (type === "list") placeholder = "[...]";
+                              else if (type === "dict") placeholder = "{...}";
+                            }
+                            const disabledSuffix = plugin.edit
+                              ? " input-dark "
+                              : " disabled text-dark input-secondary ";
+
+                            return (
+                              <Row className="py-2" key={`config.${index + 0}`}>
```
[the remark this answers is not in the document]
devmrfitz: It's needed to prevent `react/no-array-index-key`. That lint rule is irrelevant here as every entry is indeed uniquely identified by the index

frontend/src/components/me/config/UserConfig.jsx
```diff
@@ -0,0 +1,332 @@
...
+        <Formik initialValues={{ config: respData }} onSubmit={null}>
+          {({ values: configurations, setFieldValue }) => (
+            <Form>
+              <FieldArray name="config">
+                {({ remove, push }) => (
+                  <FormGroup row>
+                    <Col>
+                      {configurations.config && configurations.config.length > 0
+                        ? configurations.config.map((plugin, index) => {
+                            let plugins;
```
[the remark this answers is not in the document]
devmrfitz: If we replace `plugin` with `configuration`, we'll end up having a `configurations` and a `configuration`.

devmrfitz:
>     * Please change the the migration name with something related to this feature.

Done, assuming you're referring to the PR title

drosetti:
> > ```
> > * Please change the the migration name with something related to this feature.
> > ```
> 
> Done, assuming you're referring to the PR title

No I meant the name of the migration file in the folder `api_app/migrations`: `0010_auto_20220722_1611.py` -> `0010_custom_plugin_params.py`.

This can generate error in your environment: If you rename the migration django will see a new migration to apply, but when you apply it an error will be raised because migration `0010_custom_plugin_params` will try to change the db base on 0009, but you previously applied 0010_auto_20220722_1611, so you db is not in the state of 0009 and this will break your project. 

You have two choices:
* Rename the migration, and destroy the docker's volumes of the project, this is very fast, but you will lost the db data
* Unapply the migration `0010_auto_20220722_1611`, rename it and then apply `0010_custom_plugin_params`

devmrfitz:
> > > ```
> > > * Please change the the migration name with something related to this feature.
> > > ```
> > 
> > 
> > Done, assuming you're referring to the PR title
> 
> No I meant the name of the migration file in the folder `api_app/migrations`: `0010_auto_20220722_1611.py` -> `0010_custom_plugin_params.py`.
> 
> This can generate error in your environment: If you rename the migration django will see a new migration to apply, but when you apply it an error will be raised because migration `0010_custom_plugin_params` will try to change the db base on 0009, but you previously applied 0010_auto_20220722_1611, so you db is not in the state of 0009 and this will break your project.
> 
> You have two choices:
> 
>     * Rename the migration, and destroy the docker's volumes of the project, this is very fast, but you will lost the db data
> 
>     * Unapply the migration `0010_auto_20220722_1611`, rename it and then apply `0010_custom_plugin_params`

I've done that already. My bad for the confusion

frontend/src/components/me/config/UserConfig.jsx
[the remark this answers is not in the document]
drosetti: That's true, but I think is better `configuation` for thes two reasos:

* in fact `configurations` is a list and `plugin` is an element of the list, so it makes more sense to be `configuration`.
* In the other part of the code we have something like:
  ```
  if (
    plugin.plugin_name &&
    plugins[plugin.plugin_name]
  )
    attributeList = Object.keys(
      plugins[plugin.plugin_name].params
    );
  ```
  I think is harder to read because the two variables has only a "s" of difference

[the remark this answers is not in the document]
devmrfitz: @drosetti Done

drosetti (commented):
I think the work is good, even if I have some doubts about the "string" input format.

I think the best solution is removing the `"` from both the single field and also in the list fields.
In fact if the user puts [5, 10] as input, the backend validator can suppose is a list of `int`.

devmrfitz:
> I think the work is good, even if I have some doubts about the "string" input format.
> 
> I think the best solution is removing the `"` from both the single field and also in the list fields. In fact if the user puts [5, 10] as input, the backend validator can suppose is a list of `int`.

I think this could fire back, if let's say an analyzer was expecting a list of strings and we send `[5, 10]`. For instance (I don't know if this is a good example), maybe something like `recipe_code` of `Cyberchef`.

drosetti:
> > I think the work is good, even if I have some doubts about the "string" input format.
> > I think the best solution is removing the `"` from both the single field and also in the list fields. In fact if the user puts [5, 10] as input, the backend validator can suppose is a list of `int`.
> 
> I think this could fire back, if let's say an analyzer was expecting a list of strings and we send `[5, 10]`. For instance (I don't know if this is a good example), maybe something like `recipe_code` of `Cyberchef`.
> 
> Moreover, if this triggers a silent failure (no error is generated), and we chain analyzers (I'm assuming that's what Playbooks do), the silent error would cascade and finding the actual source would become quite difficult.
> 
> Moreover, similar problems occur in interpreting of `[true, false]`.

Talking about the first problem: With the current code with the input ["5", "10"] we have the same problem. I know, this suppose the user intentionally put the wrong input, but people can also make mistakes.
About the second problem: I'm not sure a field could have a list of bool as input (it's a bit weird) and I don't think if this is a "future case" or is simply not possible.

However: your points are very good and my "solutions" are not "solutions" because introduce other problems (the two you described). I think the changes you did are enough, I approve your pr.

drosetti (approved):

devmrfitz:
Thanks a lot for the extensive review @drosetti .

As for the list of `bool`, Cyberchef has an interesting recipe [here](https://github.com/mattnotmax/cyberchef-recipes#recipe-details-5). Have a look at the various `args`.

devmrfitz:
Additionally, is the PR ready to merge now, or are any steps left?

drosetti:
> Thanks a lot for the extensive review @drosetti .

Thanks for your work and also for your patience: I know sometimes I've been annoying, (for example with the variable names), but I think it will be easier for other people in future, but also for us in case we will need to work with this code in a distant future. 

> As for the list of `bool`, Cyberchef has an interesting recipe [here](https://github.com/mattnotmax/cyberchef-recipes#recipe-details-5). Have a look at the various `args`.

Thanks for finding a real scenario!

PR #1106, opened 2022-07-15
fixes #1105

### LINKED ISSUE #1105: Healthcheck bug for connectors.
As pointed out by @fear-the-reaper, The healthcheck has a bug where if you don't set up the MISP instance properly and leave the environment variables necessary as empty, then you get a 500 status code back. I can see how [this](https://github.com/intelowlproject/IntelOwl/blob/abd84544eaec6f049972f6ead57f77c7f5079671/api_app/connectors_manager/classes.py#L67) as a safety check. It didn't work well because `cc.read_secrets(secrets_filter="url_key_name")` returns ` {'url_key_name': ''}` so it bypasses the checks below it. An easy fix would be changing the check [here](https://github.com/intelowlproject/IntelOwl/blob/abd84544eaec6f049972f6ead57f77c7f5079671/api_app/connectors_manager/classes.py#L67) should be changed from `if url is not None:` to `if url:` only and we have to handle checking the format of the URL itself below in someway to verify that it has `http://` in the beginning or not. Or we can just let the requests library take care of that and take care of the exception raised below.

The error message itself looks like this: 

```
2022-07-14 17:53:17,757 - django.request - log_response - ERROR - Internal Server Error: /api/connector/YETI/healthcheck
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/usr/local/lib/python3.9/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/opt/deploy/intel_owl/./api_app/core/views.py", line 137, in get
    health_status = self.perform_healthcheck(name)
  File "/opt/deploy/intel_owl/./api_app/connectors_manager/views.py", line 76, in perform_healthcheck
    return connectors_controller.run_healthcheck(connector_name)
  File "/opt/deploy/intel_owl/./api_app/connectors_manager/controller.py", line 129, in run_healthcheck
    status = klass.health_check(connector_name)
  File "/opt/deploy/intel_owl/./api_app/connectors_manager/classes.py", line 69, in health_check
    requests.head(url, timeout=10)
  File "/usr/local/lib/python3.9/site-packages/requests/api.py", line 100, in head
    return request("head", url, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/requests/sessions.py", line 573, in request
    prep = self.prepare_request(req)
  File "/usr/local/lib/python3.9/site-packages/requests/sessions.py", line 484, in prepare_request
    p.prepare(
  File "/usr/local/lib/python3.9/site-packages/requests/models.py", line 368, in prepare
    self.prepare_url(url, params)
  File "/usr/local/lib/python3.9/site-packages/requests/models.py", line 439, in prepare_url
    raise MissingSchema(
requests.exceptions.MissingSchema: Invalid URL '': No scheme supplied. Perhaps you meant http://?
```

mlodic:


api_app/connectors_manager/classes.py
```diff
@@ -64,7 +64,7 @@ def health_check(cls, connector_name: str) -> Optional[bool]:
             url = cc.read_secrets(secrets_filter="url_key_name").get(
                 "url_key_name", None
             )
-            if url is not None:
+            if url and url.startswith("http://"):
```
0x0elliot, line 67: I think it should also be acceptable to add HTTPS support? I mean, You never know right? `if url and url.startswith(("http://", "https://")):` should do it.

mlodic: ah yeah you right 🤦🏻 thanks for pointing this out

PR #1119, opened 2022-07-22
Extends key (closes #752)

### LINKED ISSUE #752: `extends` key in plugin configuration
### Problem Description

Currently, there's no way to extend an existing configuration entry. 

For instance, take the case of `Shodan_Honeyscore` and `Shodan_Search` analyzer configurations. Both share most attributes differing in only few i.e. `description` and `params` so there is redundant copy-pasta between them.

https://github.com/intelowlproject/IntelOwl/blob/eaf95756d3af07c6604f8737896100cf22a5a68a/configuration/analyzer_config.json#L2413-L2470

Same problem applies to many other pairs too.

### Proposed Solution(s)

1. An `extends` attribute, like so:

```javascript
"Shodan_Honeyscore": {
    "type": "observable",
    "python_module": "shodan.Shodan",
    "description": "scan an IP against Shodan Honeyscore API",
    "disabled": false,
    "external_service": true,
    "leaks_info": false,
    "observable_supported": ["ip"],
    "config": {
      "soft_time_limit": 30,
      "queue": "default"
    },
    "secrets": {
      "api_key_name": {
        "env_var_key": "SHODAN_KEY",
        "description": "",
        "required": true
      }
    },
    "params": {
      "shodan_analysis": {
        "value": "honeyscore",
        "type": "str",
        "description": ""
      }
    }
  },
  "Shodan_Search": {
    "extends": "Shodan_Honeyscore",
    "description": "scan an IP against Shodan Search API",
    "params": {
      "shodan_analysis": {
        "value": "search",
        "type": "str",
        "description": ""
      }
    }
  },
```

i.e. `Shodan_Search` should,
- inherit all attributes from `Shodan_Honeyscore`
- overwrite for the attribute values defined explicitly in `Shodan_Search`
- overwrite-merge the `dict` attributes such as `params`, `secrets` and `config`.

devmrfitz:
# Description

Parsing of `extends` in plugin configuration

## Related issues
#752 

## Type of change

# Checklist

### DIFF
--- a/api_app/core/serializers.py
+++ b/api_app/core/serializers.py
@@ -175,20 +175,19 @@ class AbstractConfigSerializer(rfs.Serializer):
         if plugin_name in visited:
             raise RuntimeError(f"Circular dependency detected in {cls} config")
         visited.add(plugin_name)
+        result = config_dict[plugin_name]
         if plugin_name not in config_dict:
             raise RuntimeError(
                 f"Plugin {plugin_name} not found in {cls} config "
                 "but referenced in extends"
             )
         if "extends" in config_dict[plugin_name]:
-            temp = config_dict[plugin_name]
-            parent_plugin = temp.pop("extends")
-            config_dict[plugin_name] = deepcopy(
-                cls._complete_config(config_dict, parent_plugin, visited)
-            )
-            for key in temp:
-                config_dict[plugin_name][key] = temp[key]
-        return config_dict[plugin_name]
+            parent_plugin = config_dict[plugin_name]["extends"]
+            result = deepcopy(cls._complete_config(config_dict, parent_plugin, visited))
+            for key in config_dict[plugin_name]:
+                if key != "extends":
+                    result[key] = config_dict[plugin_name][key]
+        return result
 
     @classmethod
     @cache_memoize(
@@ -202,7 +201,7 @@ class AbstractConfigSerializer(rfs.Serializer):
         """
         config_dict = cls._read_config()
         for plugin in config_dict:
-            cls._complete_config(config_dict, plugin, set())
+            config_dict[plugin] = cls._complete_config(config_dict, plugin, set())
         serializer_errors = {}
         for key, config in config_dict.items():
             new_config = {"name": key, **config}
--- a/api_app/core/serializers.py
+++ b/api_app/core/serializers.py
@@ -6,6 +6,7 @@ import json
 import logging
 import os
 import sys
+from copy import deepcopy
 from typing import List, Optional, TypedDict
 
 from cache_memoize import cache_memoize
@@ -95,6 +96,7 @@ class AbstractConfigSerializer(rfs.Serializer):
     params = rfs.DictField(child=_ParamSerializer())
     # automatically populated fields
     verification = rfs.SerializerMethodField()
+    extends = rfs.CharField(allow_blank=True, required=False)
 
     def is_valid(self, raise_exception=False):
         ret = super().is_valid(raise_exception=raise_exception)
@@ -164,6 +166,31 @@ class AbstractConfigSerializer(rfs.Serializer):
         return md5hash
 
     @classmethod
+    def _complete_config(
+        cls, config_dict: dict, plugin_name: str, visited: set
+    ) -> dict:
+        """
+        Completes config by parsing extends configs
+        """
+        if plugin_name in visited:
+            raise RuntimeError(f"Circular dependency detected in {cls} config")
+        visited.add(plugin_name)
+        if plugin_name not in config_dict:
+            raise RuntimeError(
+                f"Plugin {plugin_name} not found in {cls} config "
+                "but referenced in extends"
+            )
+        if "extends" in config_dict[plugin_name]:
+            temp = config_dict[plugin_name]
+            parent_plugin = temp.pop("extends")
+            config_dict[plugin_name] = deepcopy(
+                cls._complete_config(config_dict, parent_plugin, visited)
+            )
+            for key in temp:
+                config_dict[plugin_name][key] = temp[key]
+        return config_dict[plugin_name]
+
+    @classmethod
     @cache_memoize(
         timeout=sys.maxsize,
         args_rewrite=lambda cls: f"{cls.__name__}-{cls._md5_config_file()}",
@@ -174,6 +201,8 @@ class AbstractConfigSerializer(rfs.Serializer):
         This function is memoized for the md5sum of the JSON file.
         """
         config_dict = cls._read_config()
+        for plugin in config_dict:
+            cls._complete_config(config_dict, plugin, set())
         serializer_errors = {}
         for key, config in config_dict.items():
             new_config = {"name": key, **config}
--- a/tests/api_app/test_config_parse.py
+++ b/tests/api_app/test_config_parse.py
@@ -1,7 +1,5 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
-import random
-from hashlib import md5
 from unittest.mock import Mock, patch
 
 from django.test import TestCase
@@ -10,9 +8,11 @@ from api_app.core.serializers import AbstractConfigSerializer
 
 
 class ConfigParseTests(TestCase):
-    @patch("api_app.core.serializers.AbstractConfigSerializer._md5_config_file")
+    @patch(
+        "api_app.core.serializers.AbstractConfigSerializer._md5_config_file", new=Mock()
+    )
     @patch("api_app.core.serializers.AbstractConfigSerializer._read_config")
-    def test_extends(self, mock_read_config: Mock, mock_md5_config_file: Mock):
+    def test_extends(self, mock_read_config: Mock):
         sample_analyzer_config = {
             "Shodan_Honeyscore": {
                 "type": "observable",
@@ -52,23 +52,60 @@ class ConfigParseTests(TestCase):
         }
 
         pre_parsed_sample_config = {
-            "Shodan_Honeyscore": sample_analyzer_config["Shodan_Honeyscore"],
+            "Shodan_Honeyscore": {
+                "type": "observable",
+                "python_module": "shodan.Shodan",
+                "description": "scan an IP against Shodan Honeyscore API",
+                "disabled": False,
+                "external_service": True,
+                "leaks_info": False,
+                "observable_supported": ["ip"],
+                "config": {"soft_time_limit": 30, "queue": "default"},
+                "secrets": {
+                    "api_key_name": {
+                        "env_var_key": "SHODAN_KEY",
+                        "description": "",
+                        "required": True,
+                    }
+                },
+                "params": {
+                    "shodan_analysis": {
+                        "value": "honeyscore",
+                        "type": "str",
+                        "description": "",
+                    }
+                },
+            },
             "Shodan_Search": {
-                **sample_analyzer_config["Shodan_Honeyscore"],
-                **sample_analyzer_config["Shodan_Search"],
+                "type": "observable",
+                "python_module": "shodan.Shodan",
+                "disabled": False,
+                "external_service": True,
+                "leaks_info": False,
+                "observable_supported": ["ip"],
+                "config": {"soft_time_limit": 30, "queue": "default"},
+                "secrets": {
+                    "api_key_name": {
+                        "env_var_key": "SHODAN_KEY",
+                        "description": "",
+                        "required": True,
+                    }
+                },
+                "description": "scan an IP against Shodan Search API",
+                "params": {
+                    "shodan_analysis": {
+                        "value": "search",
+                        "type": "str",
+                        "description": "",
+                    }
+                },
             },
         }
-        pre_parsed_sample_config["Shodan_Search"].pop("extends")
 
         mock_read_config.return_value = sample_analyzer_config
-        mock_md5_config_file.return_value = md5(
-            str(random.random()).encode("utf-8")
-        ).hexdigest()
-        parsed_config = AbstractConfigSerializer.read_and_verify_config()
+        # _refresh ensures that cache isn't used, if any
+        parsed_config = AbstractConfigSerializer.read_and_verify_config(_refresh=True)
 
         mock_read_config.return_value = pre_parsed_sample_config
-        mock_md5_config_file.return_value = md5(
-            str(random.random()).encode("utf-8")
-        ).hexdigest()
-        expected_config = AbstractConfigSerializer.read_and_verify_config()
+        expected_config = AbstractConfigSerializer.read_and_verify_config(_refresh=True)
         self.assertDictEqual(parsed_config, expected_config)
--- a/tests/api_app/test_config_parse.py
+++ b/tests/api_app/test_config_parse.py
@@ -2,26 +2,17 @@
 # See the file 'LICENSE' for copying permission.
 import random
 from hashlib import md5
+from unittest.mock import Mock, patch
 
 from django.test import TestCase
 
 from api_app.core.serializers import AbstractConfigSerializer
 
 
-class ModAbstractConfigSerializer(AbstractConfigSerializer):
-    input_config = None
-
-    @classmethod
-    def _md5_config_file(cls) -> str:
-        return md5(str(random.random()).encode("utf-8")).hexdigest()
-
-    @classmethod
-    def _read_config(cls):
-        return cls.input_config
-
-
 class ConfigParseTests(TestCase):
-    def test_extends(self):
+    @patch("api_app.core.serializers.AbstractConfigSerializer._md5_config_file")
+    @patch("api_app.core.serializers.AbstractConfigSerializer._read_config")
+    def test_extends(self, mock_read_config: Mock, mock_md5_config_file: Mock):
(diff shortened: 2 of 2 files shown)


mlodic:
ah one thing, please use `closes` as a word for the issue in the template cause otherwise they won't be automatically closed once merged

api_app/core/serializers.py
```diff
@@ -174,6 +200,8 @@ def read_and_verify_config(cls) -> dict:
         This function is memoized for the md5sum of the JSON file.
         """
         config_dict = cls._read_config()
+        for plugin in config_dict:
+            cls._complete_config(config_dict, plugin, set())
```
mlodic, line 204: is this loop needed here? I think you could just call the function with `config_dict` as a single param and then do the loop inside the function

devmrfitz:
@mlodic looks good?

devmrfitz:
image
Is `develop` the default branch?

api_app/core/serializers.py
```diff
@@ -163,6 +164,31 @@ def _md5_config_file(cls) -> str:
...
+        """
+        if plugin_name in visited:
+            raise RuntimeError(f"Circular dependency detected in {cls} config")
+        visited.add(plugin_name)
+        if plugin_name not in config_dict:
+            raise RuntimeError(
+                f"Plugin {plugin_name} not found in {cls} config "
+                "but referenced in extends"
+            )
+        if "extends" in config_dict[plugin_name]:
```
mlodic, line 182: shouldn't we add this as an optional attribute in the serializer?

mlodic:
> image Is `develop` the default branch?

no (`master`) but that is written down also in the merged commit if you write it in the description or in a commit in this PR

api_app/core/serializers.py
devmrfitz: Then I would have to create another function as the loop itself needs to call a function. For reference, this is just like Depth-First Traversal of a directed graph. The loop ensures that all components of the graph are traversed.

api_app/core/serializers.py
devmrfitz: done

devmrfitz:
> > image Is `develop` the default branch?
> 
> no (`master`) ~but that is written down also in the merged commit if you write it in the description or in a commit in this PR~
> 
> EDIT: we need to have that clause in the merged commit. So one thing to do is to write this down in the PR description. That will work once we squash the commits into a single one. In this way, once merging from develop to master we will have a commit with the right clause

I just wrote it into the commit message. I guess that should take care of it

api_app/core/serializers.py
mlodic: thanks for clarification, that makes sense

mlodic:
in the next weeks please refer to the other mentors to close the work on your PRs. I asked them a double check on all of the recent PRs too just to have another point of view.

Thanks

devmrfitz:
> in the next weeks please refer to the other mentors to close the work on your PRs. I asked them a double check on all of the recent PRs too just to have another point of view.
> 
> Thanks

Could you explain what "refer to the other mentors" means?

mlodic:
it means that you need to use the Slack shared chats :P

drosetti (commented):
I suggest to add some tests. You can write some tests of  `read_and_verify_config` and mock the config file output, in this way the extends function (`_complete_config`) will be tested.

api_app/core/serializers.py
```diff
@@ -163,6 +165,31 @@ def _md5_config_file(cls) -> str:
...
+            temp = config_dict[plugin_name]
+            parent_plugin = temp.pop("extends")
+            config_dict[plugin_name] = deepcopy(
+                cls._complete_config(config_dict, parent_plugin, visited)
+            )
+            for key in temp:
+                config_dict[plugin_name][key] = temp[key]
+        return config_dict[plugin_name]
```
0ssigeno, line 191: Fun of recursion, but not a fan of modifying inputs. I would have suggest that you take the input, deepcopy it and return the new dict as output. Then, where you can write  
```python3
        for key, plugin in config_dict.items():
            config_dict[key] = cls._complete_config(config_dict, plugin, set())
```
Now, i see that this is technically slower, because you are making _another_ copy of the dictionary, but Imho the readability/maintainability > 0.2% increased speed.

devmrfitz:
> I suggest to add some tests. You can write some tests of `read_and_verify_config` and mock the config file output, in this way the extends function (`_complete_config`) will be tested.

Done. Kindly re-review

api_app/core/serializers.py
```diff
@@ -175,20 +175,19 @@ def _complete_config(
         if plugin_name in visited:
             raise RuntimeError(f"Circular dependency detected in {cls} config")
         visited.add(plugin_name)
+        result = config_dict[plugin_name]
```
0ssigeno, line 178: If you are making this this assignment you are still changing the value of `config_dict[plugin_name]` because for python dict/lists are passed by reference, not value.
Example:
```python3

a = {"key": "value"}
b = a
b["key"] = "value2"
assert a["key"] == "value2"
```

tests/api_app/test_config_parse.py
```diff
@@ -0,0 +1,77 @@
...
+class ModAbstractConfigSerializer(AbstractConfigSerializer):
+    input_config = None
+
+    @classmethod
+    def _md5_config_file(cls) -> str:
+        return md5(str(random.random()).encode("utf-8")).hexdigest()
+
+    @classmethod
+    def _read_config(cls):
+        return cls.input_config
```
drosetti, line 20: This is a smart way to mock the function, but I prefer the use of the unittest library. You can use the `patch` decorator to mock the `_read_config` return.

tests/api_app/test_config_parse.py
```diff
@@ -0,0 +1,77 @@
...
+            },
+        }
+        pre_parsed_sample_config["Shodan_Search"].pop("extends")
+
+        ModAbstractConfigSerializer.input_config = sample_analyzer_config
+        parsed_config = ModAbstractConfigSerializer.read_and_verify_config()
+
+        ModAbstractConfigSerializer.input_config = pre_parsed_sample_config
+        expected_config = ModAbstractConfigSerializer.read_and_verify_config()
+        self.assertDictEqual(parsed_config, expected_config)
```
drosetti, line 77: With the mock of `_read_config` you should be able to compare via assertDictEqual the return of the function and the expected dict.

In the mock there is the dict with the "extends" keyword and in the assertDictEquals there is the comparison between the return of the function to test and the dict with the replacement of "extends" with the data

drosetti (commented):
I think is cleaner to use unittest library than extends the class and override the methods. However I appreciated a lot your idea!

tests/api_app/test_config_parse.py
devmrfitz: @drosetti If I mock only `_read_config` and not `_md5_config_file`, it will prevent `read_and_verify_config` from getting executed multiple times as, due to the actual file not changing, the md5 won't change and hence function won't get re-called

tests/api_app/test_config_parse.py
devmrfitz: Done

tests/api_app/test_config_parse.py
```diff
@@ -0,0 +1,74 @@
...
+        mock_read_config.return_value = sample_analyzer_config
+        mock_md5_config_file.return_value = md5(
+            str(random.random()).encode("utf-8")
+        ).hexdigest()
+        parsed_config = AbstractConfigSerializer.read_and_verify_config()
+
+        mock_read_config.return_value = pre_parsed_sample_config
+        mock_md5_config_file.return_value = md5(
+            str(random.random()).encode("utf-8")
+        ).hexdigest()
```
drosetti, line 72: In the tests is better to write verbose code, in this way you can understand quickly the test values or expected results. I suggest to replace computation like this with the dict.

tests/api_app/test_config_parse.py
```diff
@@ -0,0 +1,74 @@
...
+        pre_parsed_sample_config = {
+            "Shodan_Honeyscore": sample_analyzer_config["Shodan_Honeyscore"],
+            "Shodan_Search": {
+                **sample_analyzer_config["Shodan_Honeyscore"],
+                **sample_analyzer_config["Shodan_Search"],
+            },
+        }
+        pre_parsed_sample_config["Shodan_Search"].pop("extends")
+
+        mock_read_config.return_value = sample_analyzer_config
```
drosetti, line 63: The mock of _read_config (mock_read_config) should return the dict with the `extends` keyword, because the logic of replacement of the `extends` field with the "parent" analyzer's fields is done by your function (`_complete_config`) called inside `read_and_verify_config` after `_read_config`.

With `pre_parsed_sample_config` you don't test the logic, because the data are in the converted format in the mock of the file.

drosetti (changes requested):

tests/api_app/test_config_parse.py
```diff
@@ -0,0 +1,74 @@
...
+from unittest.mock import Mock, patch
+
+from django.test import TestCase
+
+from api_app.core.serializers import AbstractConfigSerializer
+
+
+class ConfigParseTests(TestCase):
+    @patch("api_app.core.serializers.AbstractConfigSerializer._md5_config_file")
+    @patch("api_app.core.serializers.AbstractConfigSerializer._read_config")
```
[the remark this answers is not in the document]
devmrfitz: If I specify `mock_read_config.return_value = sample_analyzer_config` inline, it makes us believe that `sample_analyzer_config` is the default return value of the mock. However, the mock is called twice and returns a different value, `pre_parsed_sample_config` (line 69) the second time. Hence, we'll be using two different ways of specifying return value, leading to confusion.

tests/api_app/test_config_parse.py
devmrfitz: Let's say we have 2 configs: 
```
A: {A.config},
B: {"extends": A, ...(B.config)}
```
and 
```
A: {A.config},
B: {...(A.config), ...(B.config)}
```
.
The goal of this PR is to ensure that these two configs are interpreted in the same manner. That's what I'm testing using `sample_analyzer_config` and `pre_parsed_sample_config`.

devmrfitz: @drosetti

tests/api_app/test_config_parse.py
devmrfitz: We just need to ensure that the two calls return different md5sums, so the caching doesn't get triggered. That's why I've used this. Does that sound good?

drosetti:
Ok, now I've got your point and you are totally right, excuse me for my misunderstanding of the test.

I suggest to split the test in two:
* The first one where there is the "extends" in the mock of the file.
* The second test doesn't have the "extends", but the replication of the code.

In this way you don't have to calculate the mock parameters (analyzer configs and md5s) because they are two different test: in the patch of one method you can pass the "extends" config with one md5 and in the other test you can mock the config without extends and a different md5.
In both the tests you can do an assert between the data returned by `read_and_verify_config` and the expected dict.

I know this is a lot more verbose, but it's also easier to read :)
Thanks for you patience

devmrfitz:
> Ok, now I've got your point and you are totally right, excuse me for my misunderstanding of the test.
> 
> I suggest to split the test in two:
> 
>     * The first one where there is the "extends" in the mock of the file.
> 
>     * The second test doesn't have the "extends", but the replication of the code.
> 
> 
> In this way you don't have to calculate the mock parameters (analyzer configs and md5s) because they are two different test: in the patch of one method you can pass the "extends" config with one md5 and in the other test you can mock the config without extends and a different md5. In both the tests you can do an assert between the data returned by `read_and_verify_config` and the expected dict.
> 
> I know this is a lot more verbose, but it's also easier to read :) Thanks for you patience

Our goal from this test is to verify that `extend`ing an analyzer (say X) and copying its configs (say Y) have the exact same effect i.e. we check if f(X) = f(Y). We don't care about their actual outputs. If we compare f(X) to an actual dictionary, it would place an unnecessary dependency on the actual return type and format of `read_and_verify_config`/

drosetti:
> > Ok, now I've got your point and you are totally right, excuse me for my misunderstanding of the test.
> > I suggest to split the test in two:
> > ```
> > * The first one where there is the "extends" in the mock of the file.
> > 
> > * The second test doesn't have the "extends", but the replication of the code.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this way you don't have to calculate the mock parameters (analyzer configs and md5s) because they are two different test: in the patch of one method you can pass the "extends" config with one md5 and in the other test you can mock the config without extends and a different md5. In both the tests you can do an assert between the data returned by `read_and_verify_config` and the expected dict.
> > I know this is a lot more verbose, but it's also easier to read :) Thanks for you patience
> 
> Our goal from this test is to verify that `extend`ing an analyzer (say X) and copying its configs (say Y) have the exact same effect i.e. we check if f(X) = f(Y). We don't care about their actual outputs. If we compare f(X) to an actual dictionary (the expected dict), it would place an unnecessary dependency on the actual return type and format of `read_and_verify_config`.

Ok, but with "expected dict" I don't mean "actual config" I mean the data we want as output and also we 

> > Ok, now I've got your point and you are totally right, excuse me for my misunderstanding of the test.
> > I suggest to split the test in two:
> > ```
> > * The first one where there is the "extends" in the mock of the file.
> > 
> > * The second test doesn't have the "extends", but the replication of the code.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this way you don't have to calculate the mock parameters (analyzer configs and md5s) because they are two different test: in the patch of one method you can pass the "extends" config with one md5 and in the other test you can mock the config without extends and a different md5. In both the tests you can do an assert between the data returned by `read_and_verify_config` and the expected dict.
> > I know this is a lot more verbose, but it's also easier to read :) Thanks for you patience
> 
> Our goal from this test is to verify that `extend`ing an analyzer (say X) and copying its configs (say Y) have the exact same effect i.e. we check if f(X) = f(Y). We don't care about their actual outputs. If we compare f(X) to an actual dictionary (the expected dict), it would place an unnecessary dependency on the actual return type and format of `read_and_verify_config`.

You point is good and I agree with you, I propose this changes:
* Remove the computations of the two md5s and put two hardcoded md5s (in this way we don't have the computation and they are different so we can run the function)
* Replace the computation of `pre_parsed_sample_config` with the raw dict
* Add a comment to the second md5 to explain that we need a different md5 to avoid to trigger the cache

In this way we have a more verbose test, easier to read, but we still have a single test with the comparison between the "classic" config and "extends" config. What do you think about these changes ?

devmrfitz:
>     * Remove the computations of the two md5s and put two hardcoded md5s (in this way we don't have the computation and they are different so we can run the function)
> 
>     * Replace the computation of `pre_parsed_sample_config` with the raw dict
> 
>     * Add a comment to the second md5 to explain that we need a different md5 to avoid to trigger the cache
> 
> 
> In this way we have a more verbose test, easier to read, but we still have a single test with the comparison between the "classic" config and "extends" config. What do you think about these changes ?

I've performed the instructed change to `pre_parsed_sample_config` and found a better way to bypass cache.
However, I feel that `pre_parsed_sample_config` was better in the computation form as, in future, if and when we change the `analyzer_config.json` format, we would only need to make one change, as `pre_parsed_sample_config` should ideally get automatically computed.

drosetti:
> > ```
> > * Remove the computations of the two md5s and put two hardcoded md5s (in this way we don't have the computation and they are different so we can run the function)
> > 
> > * Replace the computation of `pre_parsed_sample_config` with the raw dict
> > 
> > * Add a comment to the second md5 to explain that we need a different md5 to avoid to trigger the cache
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this way we have a more verbose test, easier to read, but we still have a single test with the comparison between the "classic" config and "extends" config. What do you think about these changes ?
> 
> I've performed the instructed change to `pre_parsed_sample_config` and found a better way to bypass cache, eliminating the md5's return value. However, I feel that `pre_parsed_sample_config` was better in the computation form as, in future, if and when we change the `analyzer_config.json` format, we would only need to make one change, as `pre_parsed_sample_config` should ideally get automatically computed.

I like a lot how you bypassed the cache, perfect!

Talking about the computation: I've got your point and this is totally correct in the codebase. I think in the tests the logic is a bit different because we want a test easy to read where the input and the expected output are well defined and not calculated, the only computations must be in the functions we want to test. From my point of view the test are written to test logic, but if you write logic in the tests, you should test the tests.

drosetti (approved):

devmrfitz:
> I like a lot how you bypassed the cache, perfect!
Thanks a lot :)
 
> Talking about the computation: I've got your point and this is totally correct in the codebase. I think in the tests the logic is a bit different because we want a test easy to read where the input and the expected output are well defined and not calculated, the only computations must be in the functions we want to test. From my point of view the test are written to test logic, but if you write logic in the tests, you should test the tests.

Oh. Got it. That explains the verbosity of tests. Thanks again!
Is there anything else left to be done towards this PR?

drosetti:
> Oh. Got it. That explains the verbosity of tests. Thanks again! Is there anything else left to be done towards this PR?

I hope I've been helpful :)

No the PR is done I proceed with the merge in the develop branch

Issue #123, opened 2020-07-15
[Feature Request] User groups and private jobs


colbyprior:
Hi,

I would like to extend the Job model to include the concept of a "private job" which references Django groups.

If a user was to create a "private job" then the job would only be visible to other users in the same group. This could be possibly tied to the "Force Privacy" attribute that already exists.

Django groups are multi-valued but it would still be possible. However it may be easier to instead extend the User object to include something like a "primary group" so that it would be force to a single value.

I would like to try and help contribute this functionality and I am interested to hear any feedback.

Eshaan7:
Hi. Thank you for your feedback! This indeed could be a nice feature.

If I understand you correctly, you basically want to categorize users into different "groups" so you can apply permissions, or some other label, to those users. Then, users can create Jobs which would be limited (in terms of view/edit permissions) to the belonging group only ?

Also, checking the `Force Privacy` attribute disables external analyzers which in some cases leak info, it's not related to user's permissions.

IMO, this can be easily integrated but most of the work is to the admin using Intel Owl him/herself. Let me explain,

- Intel Owl does not provide a registration page, only administrators are allowed to create new user accounts from the Django Admin interface. This is done with the thought in mind that Intel Owl is supposed to be an internal tool. We have plans to work on other authentication mechanisms to make this easier for larger organizations. See Issue #121.
- Django out-of-the-box has user level groups/permissions, which can be managed from the Django Admin Interface. The administrator can create groups, permissions and assign those to users, etc.
- So that said, what we as developers would need to do is,
     * provide something like a checkbox `[ ] check this to limit this analysis results to your group`, checking which will enforce an [_Object Level Permission._](https://docs.djangoproject.com/en/3.0/topics/auth/default/#permissions-and-authorization)
     * Since we use Django-Rest-Framework for the RESTful API, they provide support for [enforcing permissions](https://www.django-rest-framework.org/api-guide/permissions/#djangoobjectpermissions) too. So, like when the frontend's dashboard fetches all the Jobs, we need to only send those for which the authenticated user has the `view` permission. And they recommend using the package [drf-guardian](https://github.com/rpkilby/django-rest-framework-guardian) for this.

Eshaan7:
This was added in #135 and will be available with the next stable release `v1.3.0`.

PR #1120, opened 2022-07-22
updated PR template for automation

mlodic:


devmrfitz:
@mlodic For this to work, we necessarily need to PR to have more than 1 commit. Whereas our PR template actually recommends the user to have only 1 commit

mlodic:
you are right, thanks for pointing this out. I cry, I would like to automate all of this.

mlodic:
I removed the squash things because the maintainers would do that anyway so there is no need.

Plus, I added also a description about the case of the single commit.

.github/pull_request_template.md
```diff
@@ -1,10 +1,9 @@
+(Please add to the PR name the issue/s that this PR would close if merged by using a [Github](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) keyword. Example: `closes #999`. If your PR is made by a single commit, please add that clause in the commit too. This is all required to automate the closure of related issues.)
```
devmrfitz, line 1: We could just say "write `closes #123` in your first commit's message". That handles all cases

devmrfitz (changes requested):

Issue #121, opened 2020-07-15
Add support for external authentication methods (LDAP, GSuite, ...)


mlodic:
We should add support for some common authentication methods that could be leveraged to avoid the pain to create and manage new users directly in IntelOwl.

- [x] LDAP support
- [ ] GSuite via SAML
- [ ] others, if requested

cvdsouza:
Google OAuth2 :)

Rishabh-Kumar-07:
Hi! I was exploring the ways to authenticate through GSuite. OAuth2 appears to be a popular method to do so. Is there some specific case for which SAML is required or any of the method is fine ??

mlodic:
We can work on OAuth2 that it is easier to test and manage I guess. 

However we should keep in mind that we need the administrator to be able to apply a safelist/blocklist based on specific domains/emails, to keep control of which people can login or not.

mlodic:
This is something of really high priority. WE are looking for people who are willing to work on that :)

Eshaan7:
https://github.com/RealmTeam/django-rest-framework-social-oauth2
This package should solve all our requirements.
[List of supported backends](https://python-social-auth.readthedocs.io/en/latest/backends/index.html#supported-backends).

We wouldn't need to do any heavy lifting ourselves. 
We can just provide a `configuration/oauth_settings.py` file that is to be configured by an user however they want.

Note: this involves replacing the current token auth mechanism which wouldn't be so straight forward.

mlodic:
about the merged PR, I think we should make some minor adjustments:

- [ ]  Ouath docs should have been put in the "Authentication options" section.


- [ ] I would like to leverage the Google logo instead of the word Google. I'd put it in the header where we have the "Log In" text in the right side so it is clear that this is an alternate way to login, and the user is able to choose it before entering the data in the login form. I would also add a little link to the documentation with a "i" (that stays for "info") at the top right of that button so people understand that they have to do some work before being able to use that feature
![image](https://user-images.githubusercontent.com/30625432/184932145-defe2488-80ac-40a5-9f6d-d42e7e908933.png)

- [ ] button should be disabled if the environment variable are not set

Could you please @devmrfitz do those minor fixes?

Also, I tried to use that new functionality from a local IntelOwl instance, and, after I logged in with my Google Acoount with Oauth2 I got the following error:

```
IntegrityError at /api/auth/google-callback

null value in column "user_id" violates not-null constraint

[...]

PR #1129, opened 2022-08-01
Google OAuth (towards #121)

devmrfitz:
# Description

Added Google OAuth

## Related issues
#121 
## Type of change

# Checklist

### Important Rules

# Real World Example


devmrfitz:
Should I add the Google logo in "Google Sign In" button?

intel_owl/settings/security.py
```diff
@@ -15,7 +15,6 @@
 else:
     WEB_CLIENT_URL = f"http://{WEB_CLIENT_DOMAIN}"
 
-SESSION_COOKIE_SAMESITE = "Strict"
```
0ssigeno, line 18: @eshaan7 can we actually remove this without having any issue?

intel_owl/settings/auth.py
```diff
@@ -0,0 +1,18 @@
...
+    }
+}
+
+GOOGLE_CONF_URL = "https://accounts.google.com/.well-known/openid-configuration"
+oauth = OAuth()
+oauth.register(
+    name="google",
+    server_metadata_url=GOOGLE_CONF_URL,
+    client_kwargs={"scope": "openid email profile"},
+)
```
0ssigeno, line 18: Can we move the radius/ldap in this file too? So we have all the auth together

0ssigeno (commented):
Why didn't you use [django-allauth](https://github.com/pennersr/django-allauth)

intel_owl/settings/security.py
devmrfitz: I have sent a message on Slack regarding this as well

devmrfitz:
> Why didn't you use [django-allauth](https://github.com/pennersr/django-allauth)?

`allauth` (along with `dj-rest-auth`) is an extensive authentication solution in itself. However, we are already using `django-rest-durin` to handle the main authentication flow, and hence bringing the former would be an overkill and also require a lot of modifications to the core workings of `dj-rest-auth`.

intel_owl/settings/security.py
0ssigeno: Just adding another pointer :D

0ssigeno:
I'm reading a little bit about `allauth` to see which changes would that requires on our side. Your explanation makes sense, and I do not see any particular issue on following your implementation, just wanted to double check

intel_owl/settings/auth.py
devmrfitz: Done

devmrfitz:
> I'm reading a little bit about `allauth` to see which changes would that requires on our side. Your explanation makes sense, and I do not see any particular issue on following your implementation, just wanted to double check

The interesting thing is that I initially did a partial implementation using `dj-rest-auth` and `allauth` before going to the current implementation. xD

0ssigeno (commented):
Can we have some tests on the views/urls ?

0ssigeno:
I'm ok with this choice. If we will ever need to support _many different_ external auth, we can go on the `allauth` road

intel_owl/settings/auth.py
```diff
@@ -0,0 +1,57 @@
...
+
+    AUTHENTICATION_BACKENDS.append("intel_owl.backends.CustomRADIUSBackend")
+
+AUTHLIB_OAUTH_CLIENTS = {
+    "google": {
+        "client_id": secrets.get_secret("GOOGLE_CLIENT_ID"),
+        "client_secret": secrets.get_secret("GOOGLE_CLIENT_SECRET"),
+    }
+}
+
```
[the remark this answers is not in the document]
devmrfitz: Makes sense. What would be better, something to enable/disable all alternative authentications or individual controls (one for let's say Google OAuth, another for Github OAuth, etc)?

devmrfitz:
> Can we have some tests on the views/urls ?

That's something I'm having problems working out. How exactly do we automate the testing of Google (and possibly other) OAuths and is it not a violation of their policies?

0ssigeno:
For the urls, at least that a 200 code is returned.
For the `GoogleLoginCallbackView`, you can mock `oauth.google.authorize_access_token` and check the that the result matches that expectation. You can test the class itself instead of the `login` flow, at least we will have some kind of coverage

0ssigeno:
If you can test entire flow, mocking here and there, i will be very glad. But, as you said, do not query directly google for that, let's try to mock their answer

0ssigeno:
Dw about the `do not merge thing`; if a PR is in draft mode, we won't merge it. Once you change that, once all the comments have resolved and everyone had time to make their review, we will merge it. Just FYI for the future :D

devmrfitz:
I'm sorry for the weird PR title and message. The reason for that is the fact that when I use `DANGEROUSLY_DISABLE_HOST_CHECK=true npm start`, the backend is on port 80 while frontend is on port 3001, and the backend does not know this, hence the redirects are not working properly. I'll replace this with the proper relative URLs once the PR is ready to be merged.

devmrfitz:
> Dw about the `do not merge thing`; if a PR is in draft mode, we won't merge it. Once you change that, once all the comments have resolved and everyone had time to make their review, we will merge it. Just FYI for the future :D

Yeah. However, I was afraid that I might mark it `Ready for Review` without remembering this change, as it wouldn't pop up in normal testing. Hence I did something unusual so it hits me when I'm undoing that

intel_owl/settings/auth.py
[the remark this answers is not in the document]
devmrfitz: @0ssigeno

devmrfitz:
@0ssigeno I'm done with the tests. Any other changes you propose?

intel_owl/settings/auth.py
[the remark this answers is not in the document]
drosetti: I think individual controls is a better solution, it will require more env vars, but it will be more accurate

drosetti:
@devmrfitz

@0ssigeno  is out of office and he will come back next week, I'm reading your messages in this PR and I'm going to review you code, if you have questions or need some support feel free to tag me :D

drosetti (commented):
Good work! I have some questions/requests:

* Please add the `close` tag to close the related issue
* Currently the Google auth works in case the user already exist, if user is missing it raise an error. I think we could create the user in case it doesn't exist. I'm not but I think that also ldap authentication works in this way.
* I think we could move the `authentication` package outside `api_app`

What do you think ?

devmrfitz:
> * Currently the Google auth works in case the user already exist, if user is missing it raise an error. I think we could create the user in case it doesn't exist. I'm not but I think that also ldap authentication works in this way.

IntelOwl's current access model is that only pre-authorized people can login. Even in LDAP's case, only pre-authorized user's have their accounts in the LDAP server.
However, if we allow account creation through Google OAuth, anyone with a Gmail ID would get access to it.

tests/auth/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+            user = User.objects.get(username=username)
+            user.delete()
+        cls.user = User.objects.create_superuser(username, email, password)
+        cls.creds = {
+            "username": username,
+            "password": password,
+        }
+        # setup client
+        cls.client = APIClient()
+        return super().setUpClass()
```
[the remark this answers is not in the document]
devmrfitz: Done

devmrfitz:
> * Please add the `close` tag to close the related issue

That issue is for multiple external authentications. Should I create a separate one for Google?

devmrfitz:
> * I think we could move the `authentication` package outside `api_app`

Done

0ssigeno:
Is correct that the user is not created if the google auth fails.
Giving a fast double check, but imho is mergeable

0ssigeno (approved):

drosetti (approved):
> IntelOwl's current access model is that only pre-authorized people can login. Even in LDAP's case, only pre-authorized user's have their accounts in the LDAP server.
> However, if we allow account creation through Google OAuth, anyone with a Gmail ID would get access to it.

No you are right, you work is correct

> That issue is for multiple external authentications. Should I create a separate one for Google?

You are right again, we can leave the issue open, I will write something about the authentication  with Google

You PR is ok, I approve it and i'm going to merge it, good work!

0ssigeno:
@devmrfitz once the tests pass we are going to merge it

devmrfitz:
@0ssigeno Tests done

PR #1136, opened 2022-08-08
Storing plugin credentials in DB

devmrfitz:
# Description

Storing plugin credentials in DB

## Related issues
#978 

## Type of change

# Checklist

### Important Rules

# Real World Example


0ssigeno:
The issue talks about manage the secrets through a GUI. This allows to configure them in the Django Admin interface and, even if technically is a GUI, I do not think that this was what was wanted. I will give a code review in a sec but, before merging this, we will wait @mlodic review, so that we can confirm that solves the issue in a wanted way

devmrfitz:
@0ssigeno I was planning to implement a proper frontend for it. That's why the PR is a draft. Do you want me to not do that?

0ssigeno:
I saw the review pop up and though that you was almost done :D

Nono, having a complete frontend is absolutely better!

devmrfitz:
I tend to ask reviews at every step. Is that the preferred way or should I
avoid it?

devmrfitz:
image

devmrfitz:
@0ssigeno Could you re-review now?

0ssigeno (changes requested):
Like always, no review on the frontend, Daniele will make it.
I expect some tests on the changes, for the new stuff and, if possible, new tests for the changes (like the serializer).
Moreover, i think we should add some docs, since this PR expects that the secrets are stored in a different way and people need to be aware of that. We can even discuss if we can automatically launch the `migrate_secrets.py` during the docker startup if some api_key are found; this is just an idea, it has pros and cos.

api_app/analyzers_manager/observable_analyzers/maxmind.py
```diff
@@ -51,7 +52,11 @@ def run(self):
     @classmethod
     def updater(cls, params, db):
         # FIXME @eshaan7: dont hardcode api key name
-        api_key = secrets.get_secret("MAXMIND_KEY")
+        api_key = secrets.get_secret(
+            "api_key_name",
+            plugin_type=PluginCredential.PluginType.ANALYZER,
+            plugin_name="MaxMindGeoIP",
+        )
```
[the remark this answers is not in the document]
devmrfitz: Yes, but the method is static, and hence does not have access to `self`

api_app/signals.py
```diff
@@ -0,0 +1,25 @@
...
+        ConnectorConfigSerializer
+    )
+
+
+@receiver(post_delete, sender=PluginCredential)
+def post_delete_plugin_credential(*args, **kwargs):
+    AnalyzerConfigSerializer.read_and_verify_config.invalidate(AnalyzerConfigSerializer)
+    ConnectorConfigSerializer.read_and_verify_config.invalidate(
+        ConnectorConfigSerializer
+    )
```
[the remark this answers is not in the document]
devmrfitz: That "way" would have to read the state of the entire `PluginConfiguration` table to detect a change in it, and hence an entire table traversal on every call, hence removing the very point of performing the caching.

devmrfitz:
> We can even discuss if we can automatically launch the `migrate_secrets.py` during the docker startup if some api_key are found; this is just an idea, it has pros and cons.

@0ssigeno I think I had discussed this during my proposal formulation time, and it was decided that automatically running the script would make people dependent on the script i.e. they would just put the secrets in `env_file_app` and expect it to get loaded into the db automatically, hence removing the point of this migration-to-db in the first place

devmrfitz:
@0ssigeno @drosetti Any idea why, despite making the migrations, the table is not present in `backend-tests`?

api_app/analyzers_manager/observable_analyzers/maxmind.py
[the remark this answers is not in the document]
0ssigeno: Ah shit, 100% right. I hate that we had to use this trick here.

0ssigeno:
The error traceback is because you are importing `signals` in the `app.ready`. 
That `signals` manages stuff with models, that are not have already configured

0ssigeno:
Ok this is tricky. Long story short, django works this way:
For each app in settings, it calls the `ready` *before* doing the migration. Meaning that, if in the `ready` method you do something (like your case) with another DB table, it will not find that.
For the fixing, need a little bit more time :D

0ssigeno:
In this case, the error is caused by
```python3
    def ready(self):
        from .serializers import AnalyzerConfigSerializer  # to avoid import issue

        set_permissions(settings.DEFAULT_CACHE)
        # we "greedy cache" the config at start of application
        # because it is an expensive operation
        AnalyzerConfigSerializer.read_and_verify_config() # <----------- HERE
        ```

**Comment by 0ssigeno** [comment]
This should fix it. If you can, test it a little bit to ensure that it works, but I'm quite confident:
```python3
import sys

class AnalyzersManagerConfig(AppConfig):
    name = "api_app.analyzers_manager"

    def ready(self):

        set_permissions(settings.DEFAULT_CACHE)
        # we "greedy cache" the config at start of application
        # because it is an expensive operation
        if 'runserver' in sys.argv:

            from .serializers import AnalyzerConfigSerializer  # to avoid import issue
            AnalyzerConfigSerializer.read_and_verify_config()

```

devmrfitz:
> This should fix it. If you can, test it a little bit manually to ensure that it works, but I'm quite confident:
> 
> ```python
> import sys
> 
> class AnalyzersManagerConfig(AppConfig):
>     name = "api_app.analyzers_manager"
> 
>     def ready(self):
> 
>         set_permissions(settings.DEFAULT_CACHE)
>         # we "greedy cache" the config at start of application
>         # because it is an expensive operation
>         if 'runserver' in sys.argv:
> 
>             from .serializers import AnalyzerConfigSerializer  # to avoid import issue
>             AnalyzerConfigSerializer.read_and_verify_config()
> ```

I did a slight modification (as gunicorn doesn't use `runserver`). Does it look good?

0ssigeno:
Cool! I thought about some corner case, so we could even add the call `AnalyzerConfigSerializer.read_and_verify_config()` inside the file `wsgi.py`, that should be called once everything from Django has been set up. Maybe is even better? But as you prefer, no constraints here

0ssigeno:
Tests still fail 🤔 Let's try the wsgi.py road?

devmrfitz:
@0ssigeno looks good now?

api_app/admin.py
```diff
@@ -34,3 +34,12 @@ class JobAdminView(admin.ModelAdmin):
...
     search_fields = ("label", "color")
+
+
+@admin.register(PluginCredential)
+class PluginCredentialAdminView(admin.ModelAdmin):
+    list_display = ("id", "type", "attribute", "plugin_name")
+    search_fields = (
+        "attribute",
+        "plugin_name",
+    )
```
drosetti, line 45: Django also allows you to create filters over the fields. I think filters to select the "type" and the plugin name are useful.

intel_owl/secrets.py
```diff
@@ -63,12 +64,30 @@ def aws_get_secret(secret_name):
     return secret
 
 
-def get_secret(secret_name, default=""):
+def get_secret(secret_name, default="", plugin_type=None, plugin_name=None):
```
drosetti, line 67: Can we test this function ?

api_app/management/commands/migrate_secrets.py
```diff
@@ -0,0 +1,45 @@
...
+from api_app.analyzers_manager.serializers import AnalyzerConfigSerializer
+from api_app.connectors_manager.serializers import ConnectorConfigSerializer
+from api_app.models import PluginCredential
+
+
+class Command(BaseCommand):
+    help = "Migrates secrets from env_file_app (pre-v4.0) to database"
+
+    @staticmethod
+    def _migrate_secrets(plugin_list, plugin_type):
```
drosetti, line 17: Please add a test for this function

drosetti (changes requested):
Talking about the backend part I agree with Simone, so I don't add anything.

From the frontend prospective, I tested the UI/UX and there are several changes to do:

- [x] I noticed the button to access to the organization config in the "plugin" section appears after the organization owner create an org config. I think this button should always be available for the organization owner without hiding it at the beginning.
- [x] We should add in the "plugin" section a button to go in the secrets page (the same logic of the organization config, you can put the new button on the left of the organization config button)
- [x] The secret input don't load the secret: if you store a secret, save and then edit it again the UI doesn't load the secret.
- [ ] I like you hide the secret, but the input element seems disabled I prefer something like "password" -> "p***d"
- [ ] In order to make the GUI more similar to the exiting sections you could modify the new sections (user plugin, org plugin, secrets, for all of them you used the same base code, very good!) similar to the "Jobs": You could split in two sub-section (analyzer and connector) and also add a  bare where is possible to filter the elements.

devmrfitz:
> Talking about the backend part I agree with Simone, so I don't add anything.
> 
> From the frontend prospective, I tested the UI/UX and there are several changes to do:
> 
>     1. I noticed the button to access to the organization config in the "plugin" section appears after the organization owner create an org config. I think this button should always be available for the organization owner without hiding it at the beginning.
> 
>     2. We should add in the "plugin" section a button to go in the secrets page (the same logic of the organization config, you can put the new button on the left of the organization config button)
> 

Alright.

>     3. The secret input don't load the secret: if you store a secret, save and then edit it again the UI doesn't load the secret.
> 

That's as per the design.

>     4. I like you hide the secret, but the input element seems disabled I prefer something like "password" -> "p***d"

Alright.

> 
>     5. In order to make the GUI more similar to the exiting sections you could modify the new sections (user plugin, org plugin, secrets, for all of them you used the same base code, very good!) similar to the "Jobs": You could split in two sub-section (analyzer and connector) and also add a  bare where is possible to filter the elements.

Alright.

devmrfitz:
>     1. I noticed the button to access to the organization config in the "plugin" section appears after the organization owner create an org config. I think this button should always be available for the organization owner without hiding it at the beginning.

I was unable to reproduce this issue.

```
            {isUserOwner ? (
              <Link
                to="/me/organization/config"
                style={{ color: "inherit", textDecoration: "inherit" }}
              >
                <Row>
                  <Button className="my-2">
                    <BsPeopleFill className="me-2" /> Organization{" "}
                    {organization.name}&apos;s custom config
                  </Button>
                </Row>
              </Link>
            ) : null} 
```

The code also suggests that it shouldn't have occurred. Must have been a fluke.

drosetti:
> > ```
> > 1. I noticed the button to access to the organization config in the "plugin" section appears after the organization owner create an org config. I think this button should always be available for the organization owner without hiding it at the beginning.
> > ```
> 
> I was unable to reproduce this issue.

I tested with a clean environment, maybe is something that only happens at the startup. Another thing I noticed happen with the plugin config in case you delete the organization, the button is still active and the page show an unexpected error. I think the button should be disabled and the error be more specific.

drosetti (changes requested):
Thanks for you changes, I have these requests:
- [x] 1 - There is a sync problem: when you delete the org you need to reload the plugins page to make the button to go in the org config disappear
- [x] 2 - In case we delete the org and we go in the org config page there is an error, we need to handle it and show something prettier.
- [x] 3 - My fault, I didn't highlight it enough: I think is important the secret button is placed on the left of the plugin buttons and horizontally aligned with the button below, in this way we save vertical space and in the table can be show more elements. We could also put org plugin button above the user config button: in this way when an user create and org the user config is always below and we improve the UX (maybe i don't explain me well, in case tell me and i'll be more accurate).
 - [ ] 4 - My fault (again): Even if we don't have a class in the certego-ui for the table with inputs we need to improve the current section (with a sub-section for the analyzer and another for the connectors, but also with a table with an header row where users can filter the secrets). Maybe the best way could be add in the certego-ui an abstract class with a table and two class that extends it: one for the tables currently used and another for a table with inputs. In this way the job/plugins tables maintain the current behaviour (use the table currently use), we have a table with inputs for the plugin config/secrets and we also have a table class extended by the other two, to avoid code duplication (where is possible).

devmrfitz:
>     * [ ]  3 - My fault, I didn't highlight it enough: I think is important the secret button is placed on the left of the plugin buttons and horizontally aligned with the button below, in this way we save vertical space and in the table can be show more elements. We could also put org plugin button above the user config button: in this way when an user create and org the user config is always below and we improve the UX (maybe i don't explain me well, in case tell me and i'll be more accurate).
> 

image
image
Does this look good?

drosetti:
> > ```
> > * [ ]  3 - My fault, I didn't highlight it enough: I think is important the secret button is placed on the left of the plugin buttons and horizontally aligned with the button below, in this way we save vertical space and in the table can be show more elements. We could also put org plugin button above the user config button: in this way when an user create and org the user config is always below and we improve the UX (maybe i don't explain me well, in case tell me and i'll be more accurate).
> > ```
> 
> image image
> 
> Does this look good?

The two buttons in in the second row should have the same height, "your custom config" text should be in one line.
I'm not sure about the width of the org plugin config: maybe if is only above "you custom config" suggests to the user that there are two "columns", and in the right there are the plugin config buttons. I think we will wait @mlodic

devmrfitz:
>     * [ ]  4 - My fault (again): Even if we don't have a class in the certego-ui for the table with inputs we need to improve the current section (with a sub-section for the analyzer and another for the connectors, but also with a table with an header row where users can filter the secrets). Maybe the best way could be add in the certego-ui an abstract class with a table and two class that extends it: one for the tables currently used and another for a table with inputs. In this way the job/plugins tables maintain the current behaviour (use the table currently use), we have a table with inputs for the plugin config/secrets and we also have a table class extended by the other two, to avoid code duplication (where is possible).

It would be a table with filtering, dropdown fields and hidden as well as visible editable fields. Wouldn't it be quite an overcomplicated view for data with relatively (as compared to jobs) much fewer and less frequently changing entries?

mlodic:
I am trying to figure out the discussions and the possible design problems that arose.

1. buttons rendering
> The two buttons in in the second row should have the same height, "your custom config" text should be in one line.
I'm not sure about the width of the org plugin config: maybe if is only above "you custom config" suggests to the user that there are two "columns", and in the right there are the plugin config buttons. I think we will wait @mlodic

Right now how those buttons are rendered does not satisfy my expectations too.
One idea that just came up to my mind is to replicate the same button sections that IntelOwl uses in the "Job Result" page.
See this as an example:
image
image

We could put those 3 buttons at the top right using the same rendering used there. All the buttons would be in the same line (we have space) and it would be more intuitive for the user because he would expect the "action" buttons always in the same place and rendered in the same way. Thoughts about that?

2. general adjustments
image

- [ ] I noticed that some analyzers could have long names. Can we use more the available space in the Custom Config/Plugins Secrets sections? maybe doubling the space for analyzer names.

- [ ] I think that "your Custom Config" and "Plugin Secrets" are not the best names that we can give to those sections. I'd use something more intuitive and that binds them together logically. My idea is "Parameters Config" and "Secrets Config". (and "Organization Parameters Config"). In that it is clear what we are changing in those sections.

- [ ] Plus, it would make sense to add a "Note: ..." in this visualization, similarly to how Eshaan added a note in the "Plugin" section to explain clearly to the users how they can use that page. For example, we can say that the "custom config" basically allow the users to change the "Parameters" for each plugin + we can explicitly say that the User config overwrites the Org config (otherwise this is written only in the docs so it could not be found easily by users). Same thing for "Plugin Secrets".

- [ ] For "plugin secrets" we should not show the attribute key but the attribute value in my opinion. In that way it is clearer which variable we are changing. So, in that example, "api_key_name" would be "ANOMALI_THREATSTREAM_API_USER"

- [ ] While it is cool and right that the "plugin secrets" values are obscured, we should add a way to show them. Otherwise there is no way for the users to understand what he insterted. Eshaan already solved this problem in the "API access" section: 
image
We can re-use this.

- [ ] ("plugin secrets" section) If the user clicks on the modify button and tries to change a "pluigin secrets" already saved, if the new value of the secret would be empty, it would trigger a 500 error:
```
KeyError at /api/plugin-credential/2

'value'


/opt/deploy/intel_owl/api_app/serializers.py, line 737, in validate

                raise ValidationError(

                    f"{category} {attrs['plugin_name']} does not "

                    f"have attribute {attrs['attribute']}."

                )

            inclusion_params = attrs.copy()

            exclusion_params = {}

            inclusion_params.pop("value")

     
```

- [ ] I think that it makes no sense for an end user that could not even know what JSON is to get a "please insert a JSON valid value here" if the would just insert `test` instead of `"test"`. I know we already discussed about the JSON thing and backend side makes sense. The point is not changing that behavior but to allow the user to insert a string like `test` without the need to add `"`. (we could add `"` via javascript if we need it, solving the problem for the user) That would simplify everything from an user perspective
image


3. new sections' permissions

- [ ] If I login with an user that belongs to an org and then I go to the "Organization Config" section, I would see this:
image
But this is wrong because 1-I am already part of an org and right now users cannot be part of more than one org. 2-A normal org member should be able to see how its admin have configured the analyzers. What he should not do is to change them. So we should provide the same view but read-only.

- [ ] "Plugin secrets" section should be viewed and modified only by staff users. Is already like that?
Also, I think that the section should not be visible here 
image
by normal users because there is no need that they know that that section exists. Same thing for the "Django Admin Interface" link

4. Refactor of new sections

>  4 - My fault (again): Even if we don't have a class in the certego-ui for the table with inputs we need to improve the current section (with a sub-section for the analyzer and another for the connectors, but also with a table with an header row where users can filter the secrets). Maybe the best way could be add in the certego-ui an abstract class with a table and two class that extends it: one for the tables currently used and another for a table with inputs. In this way the job/plugins tables maintain the current behaviour (use the table currently use), we have a table with inputs for the plugin config/secrets and we also have a table class extended by the other two, to avoid code duplication (where is possible).

> It would be a table with filtering, dropdown fields and hidden as well as visible editable input fields. Wouldn't it be quite an overcomplicated view for data with relatively (as compared to jobs) much fewer and less frequently changing entries?

Even if I agree with the fact the those new sections are changed and used rarely if compared to other sections, on the other hand it is important to keep uniform the different sections of the application. In the end, those new sections contain tables so Daniele's point of view makes sense to me. The point is that these tables need to be _editable_ tables so that would require several changes, at least the ones that Daniele highlighted.
Then, about the "overcomplicated view": this may make sense in some cases but we should take into account IntelOwl deployments where a lot of configuration is added too. In that case, filtering capabilities would be really useful. Plus, as already said, I prefer to keep the frontend uniform between sections because this helps UX so I think it would make sense to do that work too.
However I'd do that once all the other things are solved, in a different PR that would be the refactor of these sections.
In this way, meanwhile, we could already start to use the new sections once this PR is solved.

Thoughts?

devmrfitz:
> I forgot that, obviously, also the secrets (like the params) should have 2 different types of configuration, one for the user and one for the org. Otherwise, all the multi-tenancy things won't work. So basically both cases should work the same.

I was thinking of modifying the CustomConfig model and merging PluginCredential into it, since they are becoming so similar. What's your perspective?

Also, weirdly didn't get an email of your last comment.

@mlodic

devmrfitz:
> I think that it makes no sense for an end user that could not even know what JSON is to get a "please insert a JSON valid value here" if the would just insert test instead of "test". I know we already discussed about the JSON thing and backend side makes sense. The point is not changing that behavior but to allow the user to insert a string like test without the need to add ". (we could add " via javascript if we need it, solving the problem for the user) That would simplify everything from an user perspective

That still maintains the case of: If a variable was initially `abc123` and I change it to `123`, is it now a string or number. Assumptions are risky because in some analyzers (like CyberChef), we know that the 2 will produce different behaviours

mlodic:
> That still maintains the case of: If a variable was initially abc123 and I change it to 123, is it now a string or number? Assumptions are risky because in some analyzers (like CyberChef), we know that the 2 will produce different behaviours

if the analyzer supports a string, then `123` would be a string and we do the `"` addition, while if the analyzer supports an int that would be converted to an int. This is what a simple user would expect from the application. We should solve the problem for the users, to keep things simple and intuitive.

> I was thinking of modifying the CustomConfig model and merging PluginCredential into it, since they are becoming so similar. What's your perspective?

This can make sense. We could also support them in the same visualization in the end. Like we have the following for "Plugins"
image
we could have the same for the "Config" ("Parameters", "Secrets")

devmrfitz:
Apologies for taking so long. The refactor needed to merge the two models is much more extensive and requires much more abstraction that I had initially assumed

devmrfitz:
@mlodic Do I seem to be going on the right path?

devmrfitz:
image
@mlodic @drosetti Does this look good?

devmrfitz:
> For "plugin secrets" we should not show the attribute key but the attribute value in my opinion. In that way it is clearer which variable we are changing. So, in that example, "api_key_name" would be "ANOMALI_THREATSTREAM_API_USER"

Since we are no longer using environment variables as plugin secrets, the `env_var_key` attribute is redundant. If we show it here, we would be introducing an unnecessary dependency over it. Instead, we could look into naming the actual secrets in a better way i.e. naming the secret something like `ANOMALI_THREATSTREAM_API_USER` instead of `api_key_name`.

devmrfitz:
@mlodic All done. Kindly re-review

mlodic:
> image @mlodic @drosetti Does this look good?

it does! but it would be even cooler if "analyzers/connectors" and "org plugin" would be in the same line

mlodic:
> plugin secrets

that could make sense but would require to do non-retrocompatible changes with the analyzer_config.json file. I don't know if we want to go in that path right now. I mean, it could be faster for you to just modify that in the frontend so we can move that refactor to a later time. Thoughts?

devmrfitz:
> > plugin secrets
> 
> that could make sense but would require to do non-retrocompatible changes with the analyzer_config.json file. I don't know if we want to go in that path right now. I mean, it could be faster for you to just modify that in the frontend so we can move that refactor to a later time. Thoughts?

The refactor we do in future would again be a db-level migration, as we'll need to go from picking `env_var_key` to `api_key_name`. Hence we risk getting locked into the old format

devmrfitz:
> > image @mlodic @drosetti Does this look good?
> 
> it does! but it would be even cooler if "analyzers/connectors" and "org plugin" would be in the same line

That would require modifying the RouterTabs component in `@certego/certego-ui`. Should I look into that?

mlodic:
> That would require modifying the RouterTabs component in @certego/certego-ui. Should I look into that?

that would be great! Otherwise another solution to save space could be to put that section (new buttons you made) in the line of the "Analyzers 166 total" name. Wouldn't it be easier?

devmrfitz:
image

I was able to achieve this without breaking backward compatibility in `@certego/certego-ui`. Does it look good?

mlodic:
it looks great! I'll merge that certego-ui PR and make a release so you can pin that release here, ok?

mlodic:
ah,. just a little thing here:
image
if you notice, there is extra space below the selected button, so just a little fix on sizes should help

devmrfitz:
> if you notice, there is extra space below the selected button, so just a little fix on sizes should help
Fixed

mlodic:
published certego 0.1.2 with your change

devmrfitz:
> published certego 0.1.2 with your change

:+1: Have already updated the `package.json`. Any other pending changes for this PR?

mlodic:
I'm checking the results:

New buttons "your plugin config": super cool
Overall user experience: cool, love it

Just very little adjustments:

- [ ] I noticed that the column where there is the name of the analyzer/connectors expands based on how much space it needs. Cool. However, sometimes (like the following example) the result could not be so friendly in the end if all the element in that column are all with different lenghts. Could it be possible to intervene in some way here frontend-side?
image

- [ ] there is still the old name here:
image

- [ ] can we find fancier icons? cause these are the same of the plugins (something like an investigator for secrets for instance)
image

- [ ] if I add two values for the same parameter, a validation error is correctly triggered but that is an unmanaged error frontend side. A proper toast should be shown instead
image
image

- [ ] the explanation is really useful but it would be even more if there would an embedded link to the organization's configuration
image
Plus, you could add the very same explicit section also in the org part here:
image


After those we can move on to either the point 4 or 5. I would focus on point 5 if possible because it would make the project more complete.

devmrfitz:
@mlodic Can you help me in understanding why I'm facing the error in `npm run build`?

mlodic:
> @mlodic Can you help me in understanding why I'm facing the error in `npm run build`?

sorry what do you mean?

devmrfitz:
@mlodic Could you review it now?

devmrfitz:
> > @mlodic Can you help me in understanding why I'm facing the error in `npm run build`?
> 
> sorry what do you mean?

The tests are failing because `npm run build` is failing

mlodic:
mmm that's strange, locally it works, right? In my pc it does work

mlodic:
I think I got this: I still don't understand why it is difficult to replicate it but the bug could have been caused by a recent update of a boostrap dependency: https://github.com/intelowlproject/IntelOwl/commit/e2f976400776c1d6546e42d84c479970cdbf70a3

mlodic:
I think we should:
* align this branch to develop (where there is the 5.2.1)
* update in certego-ui the dependency to 5.2.1
* in certego-ui remove the code that is generating the error that was introduced by a bug in bootstrap 5.1 (there is a comment about it) and was fixed in the library
* make a new release of certego-ui
* import the new release here and see if the bug still is available

mlodic:
You can pull from `develop` and get the fix for the problem

devmrfitz:
@mlodic All tests have passed now. Thanks a lot for your help. Could you kindly re-review it?

mlodic:
everything is really cool man! Just one little thing:

image

could you add the same "helper" into the organization's configuration? So there is a link there too back to the plugin configuration.

I noticed that you have also worked on the #1164! Great! Reviewing this now!

devmrfitz:
> 

Done :+1: @mlodic

mlodic:
that new feature is really cool teammate! And it seems to work correctly.

few details front-end side:
- [ ] the "enabled for organization" "button" is quite hidden but its view is different from the other columns (like "active" or "configured") that the user cannot modify and it makes sense to me. Otherwise it makes less sense if we compare it with the "health check" button. If we allow the users to perform an action on the table just by clicking on it, we should make that experience be the same over all the table so those 2 buttons should be similar. So, would it be possible to make the hidden "enabled for organization" button look like the "health check" button?
image

- [ ] when I click on the "enabled organization" button I would have expected that the "Active" flag would change but that changes only if a refresh the page.

- [ ] if I login with a non-admin user for an org with a disabled analyzer, I see this:
image
The analyzer is correctly not active but the "enabled for organization" flag is set to True while it should be False. I think this is only an error in the visualization.

- [ ] Last, it would be cooler if, in the "Scan page", the disabled analyzer would be correctly loaded as disabled. I mean, the filter works if I select a disabled analyzer but the user is not helped there. Then, I noticed that this is a problem that already existed before your PR (try GreynoiseAlpha for instance...it is disabled by default but in the "select analyzers" dropdown it is selectable)...but if you got some time to help for this it would be super appreciated.

api_app/analyzers_manager/observable_analyzers/maxmind.py
```diff
@@ -58,7 +59,11 @@ def updater(cls, params, db):
             logger.warning("No running updater for Maxmind, because it is disabled")
             return
         # FIXME @eshaan7: dont hardcode api key name
```
mlodic, line 61: we can remove this comment

api_app/signals.py
```diff
@@ -0,0 +1,29 @@
...
+    build_config_cache.delay()
+
+
+@receiver(post_delete, sender=PluginConfig)
+def post_delete_plugin_credential(*args, **kwargs):
+    AnalyzerConfigSerializer.read_and_verify_config.invalidate(AnalyzerConfigSerializer)
+    ConnectorConfigSerializer.read_and_verify_config.invalidate(
+        ConnectorConfigSerializer
+    )
+    build_config_cache.delay()
```
mlodic, line 29: I really like how you explored the usage of Signals here

api_app/views.py
```diff
@@ -668,3 +684,41 @@ def update(self, request: Request, *args, **kwargs):
             request._full_data = request.data.copy()
         request.data["owner"] = request.user.id
         return super().update(request, *args, **kwargs)
+
+
+@api_view(["DELETE", "POST"])
```
mlodic, line 689: in these new views, could you also add the @add_docs?
I would also love to see a logger.info message at the start of each method. Logging is something we always forget until the moment when we need logs and we don't have them and then we cry :P

authentication/views.py
```diff
@@ -78,10 +83,18 @@ def get_security_definition(auto_schema):
         }
 
 
-def google_login(request):
+@api_view(["GET"])
```
mlodic, line 86: docs and logger here too please

docs/source/Installation.md
```diff
@@ -100,6 +100,14 @@ In the `env_file_app`, configure different variables as explained below.
 * `DJANGO_SECRET`: random 50 chars key, must be unique. If you do not provide one, Intel Owl will automatically set a new secret on every run.
 * `INTELOWL_WEB_CLIENT_DOMAIN` (example: `localhost`/`mywebsite.com`): the web domain of your instance, this is used for generating links to analysis results.
 
+
+### Deprecated environment configuration
+The following variables are deprecated and will be removed in the future.
+The new way to configure plugin secrets is to use the `Plugin Secrets` page in the GUI.
+If you had previously specified any variables in the environment, run `docker exec -ti intelowl_uwsgi python3 manage.py migrate_secrets`
+and then remove those secrets from the env file.
```
mlodic, line 108: super cool this. The only thing not stated here is the version number where these variables start to be deprecated 🤔  but obviously you cannot know that in advance...

Plus, this is also a breaking change in the end if we put in this way. 🤔  I would really like to avoid to do a 5.x release just for this change.

One cool thing would be to run this automatically in the uwsgi entrypoint _only_ if this is the first time that this is executed. How to check this? Maybe just trying to see whether there is any element in the PluginConfig table? If there is already anything, that would mean that it was already populated so we could skip that. Does it make sense to you?

tests/api_app/test_api.py
```diff
@@ -23,6 +24,7 @@ def setUpClass(cls):
         cls.superuser = User.objects.create_superuser(
             username="test", email="test@intelowl.com", password="test"
         )
+        call_command("migrate_secrets")
 
     def setUp(self):
         self.client = APIClient()
```
mlodic, line 30: could you add tests for the new endpoints you created here please?

mlodic:
ok review ended. I think we are reaching a really good state.

api_app/signals.py
devmrfitz: Thanks a lot :)

devmrfitz:
>  the "enabled for organization" "button" is quite hidden but its view is different from the other columns (like "active" or "configured") that the user cannot modify and it makes sense to me. Otherwise it makes less sense if we compare it with the "health check" button. If we allow the users to perform an action on the table just by clicking on it, we should make that experience be the same over all the table so those 2 buttons should be similar. So, would it be possible to make the hidden "enabled for organization" button look like the "health check" button?

@mlodic Could you please explain this again? I'm unable to understand

frontend/src/components/plugins/utils/utils.jsx
```diff
@@ -186,6 +191,61 @@ export function PluginHealthCheckButton({ pluginName, pluginType }) {
...
+    )
+  );
+  let title = "";
+  if (!organization.name) title = "You're not a part of any organization";
+  else if (!isUserOwner)
+    title = `You're not an owner of your organization - ${organization.name}`;
+
+  return (
+    <div className="d-flex flex-column align-items-center" title={title}>
+      <InputCheckBox
```
mlodic, line 221: what I meant is that this should be a button and not an checkbox because checkbox usually do not contain actions while buttons do and in this case if you click you change a state so it should be a button. Does it make sense now?

devmrfitz: Done :+1:

tests/api_app/test_api.py
devmrfitz: I've made them in `tests/api_app/test_org_plugin_state.py`. Should I copy them here?

devmrfitz:
@mlodic All issues resolved except 1. Have posted a doubt in that one

devmrfitz:
@mlodic @0ssigeno @drosetti Could you help me in figuring out why the tests are failing?

tests/api_app/test_api.py
mlodic: ah no no it's better like you did. I did not notice.

mlodic:
```
test_start_analyzers (tests.analyzers_manager.test_observable_scripts.IPAnalyzersTestCase) ... Traceback (most recent call last):
  File "/opt/deploy/intel_owl/api_app/core/classes.py", line 123, in start
    _result = self.run()
  File "/opt/deploy/intel_owl/api_app/analyzers_manager/observable_analyzers/stratosphere.py", line 48, in run
    self.check_dataset_status()
  File "/opt/deploy/intel_owl/api_app/analyzers_manager/observable_analyzers/stratosphere.py", line 100, in check_dataset_status
    timestamp = os.path.getctime(db_loc0)
  File "/usr/local/lib/python3.9/genericpath.py", line 65, in getctime
    return os.stat(filename).st_ctime
FileNotFoundError: [Errno 2] No such file or directory: '/opt/deploy/files_required/stratos_ip_blacklist_last24hrs.csv'
[REPORT] (poll #0) 
>>> Job:4, status:'reported_with_fails' 
>>> analyzer_reports:{'all': 59, 'failed': 1, 'pending': 0, 'running': 0, 'success': 58, 'killed': 0} 
>>> connector_reports:{'all': 0, 'failed': 0, 'pending': 0, 'running': 0, 'success': 0, 'killed': 0}  
>>> Running/Pending analyzers: [] 
>>> Running/Pending connectors: []

>>> Failed analyzers: [('Stratosphere_Blacklist', ["[Errno 2] No such file or directory: '/opt/deploy/files_required/stratos_ip_blacklist_last24hrs.csv'"])] 
>>> Failed connectors: []
```

api_app/management/commands/migrate_secrets.py
```diff
@@ -22,6 +22,13 @@ def _get_env_var(name):
...
+        if PluginConfig.objects.filter(plugin_type=plugin_type).exists():
+            print(
+                f"Skipping {plugin_type} secrets migration because "
+                "there are already some secrets in the database."
+            )
+            return
+
         User = get_user_model()
         if not User.objects.filter(is_superuser=True).exists():
             raise Exception("Superuser must exist for secrets migration")
```
mlodic, line 34: I think this won't work in a clean environment because the superuser still needs to be created manually right now.

On the other hand, in a clean environment the user won't need to import anything so I think that the flow makes sense like it is. Am I right?

mlodic:
pushed a little commit. Let's if that fixes this

api_app/management/commands/migrate_secrets.py
devmrfitz: Yup. I assumed any working environment would always have a superuser.

devmrfitz:
@mlodic All done. Is there something else to be done before merging this PR?

mlodic:
I'll wait for the actual PR of @0x0elliot to be merged before merging this. This is because he already needed to manage a lot of conflicts for the changes of your previous PRs so it is right to me that this time his PR would be merged before yours. I'll update here as soon as I have news.

Then I'll make another final review as soon as I can

devmrfitz:
> I'll wait for the actual PR of @0x0elliot to be merged before merging this. This is because he already needed to manage a lot of conflicts for the changes of your previous PRs so it is right to me that this time his PR would be merged before yours. I'll update here as soon as I have news.
> 
> Then I'll make another final review as soon as I can

That sounds good :+1:

devmrfitz:
@mlodic I'm done

mlodic:
moved to #1237

Issue #1192, opened 2022-09-09
[Analyzer] Update GreyNoise to use SDK and Remove Alpha


bradchiappetta:
## Name:
GreyNoise

## Link: 
greynoise.io

## Type of analyzer
**this can be observable, file, and docker**
Observable

## Why should we use it
This is just a recommended update to use our SDK and remove the no longer available Alpha API

## Possible implementation

bradchiappetta:
I am happy to work on this to integrate the GreyNoise SDK and make the updates I've listed.

mlodic:
Thank you for listing out the updates to do. Feel free to open a PR with the updates! It would be appreciated!

bradchiappetta:
Hi @mlodic : I saw the updated PR merged into the `develop` branch.  Does that indicate everything was resolved and you could accept the update?  If so, just curious what your release process is for when that is available.  I have a client that I'd like to notify when they can use it.

mlodic:
We have no scheduled release right now because we are independent in that sense and we strongly rely on voluntarily help so it depends on when we get enough updates that justify a new release.

I am also very busy right now and Docker Hub deleted my autobuild configuration so that would require some time that I don't know when I will be able to invest. Please follow the project updates and you'll get the notification on the new release.

mlodic:
Obviously if Greynoise would choose to support the project for a donation, I can speed up the process :P Thank you for your understanding

mlodic:
closing because will be merged in the next release

mlodic:
> Obviously if Greynoise would choose to support the project for a donation, I can speed up the process :P Thank you for your understanding

well, considered that I helped you finish this task and basically that addition did not bring anything really valuable in the platform, I could at least deserve an answer. @bradchiappetta

mlodic:
and your release is available, you can freely gain money out of this open source project

PR #1193, opened 2022-09-09
1192 greynoise updates

bradchiappetta:
# Description

Please include a summary of the change.

- Updates GreyNoise to use the official SDK
- Removes the GreyNoise Alpha (v1) API, which no longer exists
- Requires v3/Community integration to enter an API key.  Keys are still free, you just need to sign up
- Updates v2/Paid lookup to pull from both the NOISE and RIOT datasets to pull back total info
- Renames analyzer class to remove conflict with SDK

## Related issues
Please add related issues.
#1192 

## Type of change

# Checklist

### Important Rules


mlodic:
it's great to see your PR here :) feel free to ask for a review once you are ready! thanks!

bradchiappetta:
@mlodic : I'm close. I'm struggling with how to get the failing test to work correctly.  I can't figure out how these function and if you use API keys built into the testing to validate these. The test returns a 401, which would be expected without an API key.

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def run(self):
+
+        if self.api_version == "v2":
+            api_key = self._secrets["api_key_name"]
+            session = GreyNoise(
+                api_key=api_key, integration_name="greynoise-intelowl-v1.0"
+            )
+            context_response = session.ip(self.observable_name)
+            riot_response = session.riot(self.observable_name)
```
mlodic, line 24: just to get a better understanding of this, could you please upload an example output of a successful query?

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
```diff
@@ -0,0 +1,58 @@
...
+            response = context_response.copy()
+            response.update(riot_response)
```
mlodic, line 26:
```suggestion
            response = session.ip(self.observable_name)
            response |= session.riot(self.observable_name)
```

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
```diff
@@ -0,0 +1,58 @@
...
+
+        elif self.api_version == "v3":
+            api_key = self._secrets["api_key_name"]
+            session = GreyNoise(
+                api_key=api_key,
+                integration_name="greynoise-community-intelowl-v1.0",
```
mlodic, line 32: how this integration_name is used in the pygreynoise library?

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
```diff
@@ -0,0 +1,58 @@
...
+from api_app.exceptions import AnalyzerRunException
+from tests.mock_utils import MockResponse, if_mock_connections, patch
+
+
+class GreyNoiseAnalyzer(classes.ObservableAnalyzer):
+    def set_params(self, params):
+        self.api_version = params.get("greynoise_api_version", "v3")
+        self.max_records_to_retrieve = int(params.get("max_records_to_retrieve", 500))
+
+    def run(self):
```
mlodic, line 16: an overall question. How the 4xx and 5xx status codes are handled by this library? What happens if one of them is triggered?

configuration/analyzer_config.json
```diff
@@ -1246,7 +1246,7 @@
       "api_key_name": {
         "env_var_key": "GREYNOISE_API_KEY",
         "description": "",
-        "required": false
+        "required": true
```
mlodic, line 1249: does this become mandatory for Community APIs?

mlodic: I read "Available to unauthenticated users with a limited number of lookups per day" in the documentation so this one should be kept as `false`

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
```diff
@@ -0,0 +1,58 @@
...
+                    "requests.get",
+                    return_value=MockResponse({}, 200),
+                ),
+                patch(
+                    "requests.post",
+                    return_value=MockResponse({}, 200),
+                ),
+            )
+        ]
+        return super()._monkeypatch(patches=patches)
```
mlodic, line 58: for the tests you should just mock the Classes/functions that you call in your code here with the results that you want.

This is an example: https://github.com/intelowlproject/IntelOwl/blob/master/api_app/analyzers_manager/file_analyzers/cape_sandbox.py#L209

mlodic: so you should mock `greynoise.Greynoise.<methods>` and add some possible output that you would get in real life

mlodic (changes requested):

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
bradchiappetta: I've added some screenshots to the issue, but essentially this merges the responses from both of our IP lookup endpoints and provides a consolidated output.

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
bradchiappetta: This is helpful for us, it is just updating a User-Agent value in the request so we can identify API usage based on integration.  It helps us understand the adoption of the integration and how well it is being used and consumed by our users.

configuration/analyzer_config.json
bradchiappetta: The experience with the API is better since the limits are higher, and it is preferred for integrations.  The API key is still totally free and takes only a second to generate, so it is always our desire that using an API key be required.

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
bradchiappetta: They are going to be returned as exceptions, so I've added in handling for that, so it returns the error back to the user.  Let me know if you'd like to see something different here.

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
mlodic: thank you!

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
mlodic: thanks for the clarification!

configuration/analyzer_config.json
mlodic: I completely understand this but, on the other hand, this limits the chance for the users who just use a tool like IntelOwl for educational purposes. I mean, if you want to have the users to always set an API key it makes completely sense to me but that restriction should be made on your side, not in IntelOwl. So, still, this should be kept as `false` if your service still provide open access. If you reconsider your policy, I'll change this.

bradchiappetta: Happy to change this back, no worries.  I'll include it in my next update.

api_app/analyzers_manager/observable_analyzers/greynoiseintel.py
bradchiappetta: @mlodic : I'm having an issue trying to make this work, I think I just don't have enough familiarity with this functionality.  This is what I've been trying, but I can't get it to work.  Any guidance you can provide?

```
    @classmethod
    def _monkeypatch(cls):
        patches = [
            if_mock_connections(
                patch(
                    "greynoise.GreyNoise.ip",
                    return_value=MockResponse(
                        {"noise": True}, 200
                    ),
                ),
                patch(
                    "greynoise.GreyNoise.riot",
                    return_value=MockResponse(
                        {"riot": True}, 200
                    ),
                ),
            )
        ]
        return super()._monkeypatch(patches=patches)
```

I keep facing this and I'm not sure how to work around it: `[('GreyNoise', ["GreyNoise Exception: unsupported operand type(s) for |=: 'MockResponse' and 'MockResponse'"])]`

mlodic: I wanted to fix this for you but I cannot push the changes in this PR, I don't know why exactly, maybe some permissions on your fork side...
try this please:


```
   @classmethod
    def _monkeypatch(cls):
        patches = [
            if_mock_connections(
                patch(
                    "greynoise.Greynoise",
                    return_value=MockGreynoise(),
                ),
            )
        ]
        return super()._monkeypatch(patches=patches)```

and add this class on top of the module:

`class MockGreynoise:
    def __init__(self):
        pass

    def ip(self, *args, **kwargs):
        return {"noise": True}

    def riot(self, *args, **kwargs):
        return {"riot": True}`

bradchiappetta: @mlodic : tried this, but now I'm back to getting a 401.  Would you like me to push the change up to the branch so you can see if I did anything wrong?

bradchiappetta: @mlodic : alright, latest commit is pushed so you can see where I'm at.  Appreciate the help.

mlodic:
I'll merge this to a branch local to IntelOwl and try to fix the problem there. Please follow that branch for updates

PR #1204, opened 2022-09-15
New greynoise integration

mlodic:


mlodic:
for reference @bradchiappetta

mlodic:
@bradchiappetta there is a problem in the Greynoise client. It does not allow to run queries without api keys set. For what we have already discussed in the other PR, we can't remove that functionality. So if you don't patch it to allow that, I'll have to refuse the integration of the client

bradchiappetta:
> @bradchiappetta there is a problem in the Greynoise client. It does not allow to run queries without api keys set. For what we have already discussed in the other PR, we can't remove that functionality. So if you don't patch it to allow that, I'll have to refuse the integration of the client

@mlodic : there are two different components here.  The v2/Paid version, requires an API key, as these are the paid endpoints being used.  The v3/Community version does not require an API, it is optional.  You'll note the "offering" attribute on line 46, which tells the SDK to use the v3 endpoint.  It will function without an API key.  Here is a shortened version for you to test to confirm

```
from greynoise import GreyNoise
api_key = ''
session = GreyNoise(
    api_key=api_key,
    integration_name="greynoise-community-intelowl-v1.0",
    offering="Community",
)
response = session.ip("8.8.8.8")
print(response)
```
```
{'ip': '8.8.8.8', 'noise': False, 'riot': True, 'classification': 'benign', 'name': 'Google Public DNS', 'link': 'https://viz.greynoise.io/riot/8.8.8.8', 'last_seen': '2022-09-15', 'message': 'Success'}
```

mlodic:
ok thank you for the answer. I'll make a test and continue the work

PR #1254, opened 2022-10-22
Closes #1225 Playbook support for ask availability endpoint.

### LINKED ISSUE #1225: Playbooks availability check
referring to PR #1123 :

```
I see that for playbooks the "Extra configuration" section was removed. This is because we are not doing a check for already existing playbooks running, right? I am talking about the ask_multi_analysis_availability function that we call before running analyzers/connectors normaly. Logically, it would make sense to have the same check for playbooks. But I understand that this would add additional burden. Let's ignore this for now. I'll open an issue
```

The point is that right now we don't check if a playbook was recently executed so we risk to multiply analysis in that sense. We should implement the same thing that we use for analyzers/connectors

@0x0elliot comment:
```
yes doing it right now wouldn't be right. but speaking of this, i wanted to bring this up to your attention but i think we should make it so that instead of it using cached results for a whole playbook, this feature takes the cached result for individual analyzers. this is a better solution now that we are moving in the direction of bulk searching. this is of course for whenever it is done.
```

0x0elliot:
# Description

Adding playbook support for Ask for Availability endpoint

## Type of change

# Checklist

- [x] A new analyzer or connector was added, in which case:
    - [x] Secrets were added in [env_file_app_template](https://github.com/intelowlproject/IntelOwl/blob/master/docker/env_file_app_template), [env_file_app_ci](https://github.com/certego/IntelOwl/blob/master/docker/env_file_app_ci) and in the [Installation](./Installation.md) docs, if necessary.
    - [x] If the analyzer/connector requires mocked testing, `_monkeypatch()` was used in its class to apply the necessary decorators.
    - [x] If a File analyzer was added, its name was explicitly defined in [test_file_scripts.py](https://github.com/intelowlproject/IntelOwl/blob/master/tests/analyzers_manager/test_file_scripts.py) (not required for Observable Analyzers).
- [x] If changes were made to an existing model/serializer/view, the docs were updated and regenerated (check [CONTRIBUTE.md](./Contribute.md)).
- [x] If the analyzer is free, Please add it in the `FREE_TO_USE_ANALYZERS` playbook in `playbook_config.json`

# Real World Example

Screenshot 2022-10-22 at 6 18 52 PM
Screenshot 2022-10-22 at 6 17 00 PM


0x0elliot:
one thing i am concerned about for this PR is that it avoids the existence of `analyzers` parameter for this API endpoint if playbooks are sent in.

0x0elliot:
oop. guess i pushed the `yarn.lock` file by mistake. explains the extra 10k lines :p

docs/source/schema.yml
```diff
@@ -235,9 +235,19 @@ paths:
...
-            status "running" or "reported_without_fails"
+            status "running" or "reported_without_fails" for analyzer checking.
+
+            However, for status being added in as a factor for Playbooks, 
+            IntelOwl also counts "running" (only if specifically selected), 
+            "reported_with_fails" or "reported_without_fails".
+
+            Also, if a playbook parameter is sent which isn't empty with/without an
+            analyzer, it completely ignores the existence of the analyzer paramter
+            and returns matching playbook jobs with the same observable.
```
mlodic, line 246: a more "clean" option could be to invalid the serializer in the case both playbooks and analyzers are set. That would help any integration in terms of understanding this case. Plus it would make sense in my opinion to notify the client that there's something wrong in the provided data.

To do that you could add a `validate` method to that serializer and just do that confrontation

0x0elliot: Good idea. let me quickly add that.

api_app/views.py
```diff
@@ -215,12 +216,33 @@ def _multi_analysis_availability(user, data):
                 Status.REPORTED_WITHOUT_FAILS,
             ]
 
-        if len(analyzers) == 0:
+        # this means that the user is trying to
+        # check avaibility of the case where all
+        # analyzers were run but no playbooks were
+        # triggered.
```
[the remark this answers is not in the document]
0x0elliot: learned it from the best :P

0x0elliot:
done. make your final review appropriately.

mlodic:
yeah I was checking the CI: the codecov error that is unrelated to your PR but the codefactor issue is a little thing to fix (2 spaces). Could you fix those spaces? otherwise I'll do it without problems. I just did not want to add my contribution in this commit just for removing 2 spaces 🤣

0x0elliot:
oop. didn't notice that. should be good now.

mlodic:
super!

PR #1324, opened 2022-12-05
Issue 1118-2 and 1118-3 solved

taher-mulla:
This PR is for Issue#1118. I have worked to fix problems numbers 2 and 3. Pls accept this and close the issue problems.

# Description

A string was changed, and for column management, I have added a constant space. Both these are separate commits.


mlodic:
please don't delete the PR template. It is there to be read, followed and compiled. I can't accept PRs without it

PR #1436, opened 2023-01-19
yaraify file analyzer fix/adjustment

mlodic:


api_app/analyzers_manager/file_analyzers/yaraify_file_scan.py
```diff
@@ -35,31 +39,66 @@ def run(self):
...
+                        )
+                        data = {"query": "get_results", "task_id": task_id}
+                        response = requests.post(self.url, json=data)
+                        response.raise_for_status()
+                        task_response = response.json()
+                        logger.debug(task_response)
+                        data_results = task_response.get("data")
+                        if isinstance(data_results, dict) and data_results:
+                            logger.info(f"found scan result for {self.md5}")
+                            break
```
0ssigeno, line 90: I would have expected that once we have the `data_results`, we always `break` , but if the type is not expected, an exception is raised

api_app/analyzers_manager/file_analyzers/yaraify_file_scan.py
```diff
@@ -35,31 +39,66 @@ def run(self):
...
+                    )
+                for _try in range(self.max_tries):
+                    try:
+                        logger.info(
+                            f"yara file scan md5 {self.md5} polling for"
+                            f" result try #{_try + 1}."
+                            f"task_id: {task_id}"
+                        )
+                        data = {"query": "get_results", "task_id": task_id}
+                        response = requests.post(self.url, json=data)
```
0ssigeno, line 83: Can we limit the scope of this `try except` clause? I think that the only point where an exception could be raised is the `raise_for_status` call

0ssigeno (approved):

api_app/analyzers_manager/file_analyzers/yaraify_file_scan.py
0ssigeno: GG

PR #1450, opened 2023-01-25
Updated initialize.sh to work with docker-compose v2 and above.

### LINKED ISSUE #1170: Failing initialize.sh with docker-compose-plugin (tested with v2.6.0)
## What happened

The docker-compose version control comparation fails with docker-compose-plugin installation

## Environment
1. OS: Ubuntu 20.04 LTS
2. IntelOwl version: 3.4.1
3. docker-compose-plugin: 2.6.0

It should work because the docker-compose version is greater then 1.23.2

Follow the installation steps and use docker-compose-plugin previously, instead to use older version 1.29.2 hardcoded in the initialize.sh script

## Output of docker-compose --version with docker-compose-plugin version 2.6.0
Docker Compose version v2.6.0
## Output expected with docker-compose-version 1.29.2
docker-compose version 1.29.2, build 5becea4c

## Error messages and logs

$ docker-compose version
Docker Compose version v2.6.0

$ ./initialize.sh 
This script will check (and possibly guide you through) the installation of dependencies for IntelOwl!
Docker version 20.10.17 detected
./initialize.sh: line 30: ((: 10#v2: value too great for base (error token is "10#v2")
./initialize.sh: line 35: ((: 10#v2: value too great for base (error token is "10#v2")
Error: Docker-compose version is too old. Please upgrade to at least 1.23.2.

### Sampe of update how to get the docker-compose version to work with docker-compose-plugin version v2.6.0 and docker-compose 1.y.z

docker_compose_version=$(docker-compose --version | tr '-' ' ' | cut -d' ' -f4 | cut -d'v' -f2 | cut -d',' -f1)

siddhanthrathod:
closes  #1170 

updated intaialize.sh to check the existence and version of docker-compse v2 and above as docker-compose 2.x is a sub command of docker now and added --user flag on pip to work with operating systems like gentoo which don't allow usage of pip without it.


initialize.sh
```diff
@@ -71,9 +71,8 @@ else
     echo "Docker version $docker_version detected"
   fi
 fi
-
 # Check if docker-compose is installed
-if ! [ -x "$(command -v docker-compose)" ]; then
+if ! [ -x "$(find /usr | grep "docker-compose$")" ]; then
```
mlodic, line 75: why this changed?

initialize.sh
```diff
@@ -92,7 +91,7 @@ if ! [ -x "$(command -v docker-compose)" ]; then
     exit 1
   fi
 else
-  IFS=',' read -ra temp <<< "$(docker-compose --version)"
+  IFS=',' read -ra temp <<< "$(docker compose version | cut -d 'v' -f3)"
```
mlodic, line 94: if you can launch `docker compose` it means that the version of docker compose plugin that you have installed is more recent than the minimum [required](https://intelowl.readthedocs.io/en/latest/Installation.html#requirements) (only from docker 2 there is this syntax). So it does not make sense to change this in my opinion.

This script should support both the existence of an old docker-compose and a new one, not just one of them.

The flow I think should be:
* verify that `docker compose` command run
* if yes, ok. If not, try `docker-compose` command run
* if yes, check version. If not, install `docker-compose` (this is already implemented)

mlodic (changes requested):

initialize.sh
```diff
@@ -126,7 +125,7 @@ else
 fi
 
 echo "Installing python dependencies using pip..."
-pip3 install -r requirements/pre-requirements.txt
+pip3 install --user -r requirements/pre-requirements.txt
```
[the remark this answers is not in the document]
siddhanthrathod: yes I'll do that

initialize.sh
siddhanthrathod: because docker-compose is stored in the cli-plugins dir in the usr dir(default:/usr/local/lib/docker/cli-plugins and in my case /usr/libexec/docker/cli-plugins/) but I should change this to work withe older version too

initialize.sh
siddhanthrathod: Yes it should support both  I'll change it

initialize.sh
siddhanthrathod: I was wrong here sorry

mlodic (approved):
do I consider this finished? It seems like it this

mlodic:
ty

PR #1452, opened 2023-01-26
updated to work with docker compose v2

siddhanthrathod:
# Description
  Used the try/catch to work with docker compose v2.


start.py
```diff
@@ -222,17 +222,30 @@ def start():
...
+                        env["BUILDKIT_PROGRESS"] = "plain"
+                subprocess.run(command, env=env, check=True)
+        #for docker compose v2
+        except OSError :
+                command = ["docker","compose"] + base_command[1:] + [args.docker_command] + unknown
+                env = os.environ.copy()
+                env["DOCKER_BUILDKIT"] = "1"
+                if args.debug_build:
+                        env["BUILDKIT_PROGRESS"] = "plain"
+                subprocess.run(command, env=env, check=True)
```
mlodic, line 246: this is cool but you have copy/pasted the code. Could you please optimize it with an external function?

mlodic:
you would also need to resolve merge conflicts before going forward

siddhanthrathod:
Okay I'll do that

start.py
siddhanthrathod: Yes I'll do that

start.py
```diff
@@ -222,6 +223,10 @@ def start():
         "--project-directory",
         "docker",
     ]
+    # for docker-compose v2
+    try:subprocess.run(base_command[0],stdout=subprocess.DEVNULL,stderr=subprocess.DEVNULL,check=True)
```
mlodic, line 227: please configure and use linters like stated in the documentation

mlodic (approved):

mlodic:
ty!

PR #1461, opened 2023-01-30
Fix secret permission

0ssigeno:


mlodic (approved):

intel_owl/settings/cache.py
```diff
@@ -1,15 +1,6 @@
-import os
-
-from ._util import set_permissions
-
-MEDIA_ROOT = "/opt/deploy/files_required"
```
mlodic, line 5: MEDIA_ROOT is used in another part of the application so I'll keep it there for now

PR #1471, opened 2023-02-01
Run on failure

### LINKED ISSUE #1431: Connectors configuration "run_on_failures"
with #1416 we are enabling connectors to execute even if there are some analyers who failed (status `reported_with_fails`), by default.

It would make sense to add an additional configuration to the connectors called "run_on_failures" which would allow the users to specify which connectors would need to execute in case of analyzers failure.
This can be useful because, for some connectors we would like to _always_ push the results at the end of the analysis, while for others we wouldn't

0ssigeno:
Closes #1431 

# Description

## Type of change

### DIFF
--- a/api_app/connectors_manager/classes.py
+++ b/api_app/connectors_manager/classes.py
@@ -1,6 +1,6 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
-
+import abc
 import logging
 from typing import Optional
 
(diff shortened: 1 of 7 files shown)


mlodic (approved):

mlodic:
I think we could take this opportunity to re-enable all the connectors by default. Right now the available connectors are disabled (see https://intelowl.readthedocs.io/en/latest/Usage.html#connectors-customization) but this makes no sense imo because, to enable them, the user would need to make changes to the json files and we are trying to avoid that as much as possible. There is no risk of pushing the information out without control because they all require manual configuration anyway.

It could fit the scope of this PR

PR #1472, opened 2023-02-02
Update quark engine

### LINKED ISSUE #1047: Use `freshquark` to keep Quark-Engine rules updated
see https://twitter.com/quarkengine/status/1539528874266005505 and their docs

0ssigeno:
# Description

Closes #1047 

## Type of change

### DIFF
--- a/api_app/analyzers_manager/file_analyzers/quark_engine.py
+++ b/api_app/analyzers_manager/file_analyzers/quark_engine.py
@@ -1,8 +1,8 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
-from django.conf import settings
 from quark import freshquark
+from quark.config import DIR_PATH
 from quark.report import Report
 
 from api_app.analyzers_manager.classes import FileAnalyzer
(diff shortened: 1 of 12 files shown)


api_app/analyzers_manager/file_analyzers/quark_engine.py
```diff
@@ -1,20 +1,23 @@
...
+from quark import freshquark
+from quark.report import Report
+
 from api_app.analyzers_manager.classes import FileAnalyzer
 from api_app.exceptions import AnalyzerRunException
 
 
 class QuarkEngine(FileAnalyzer):
+    def set_params(self, params: dict):
+        freshquark.download()
```
mlodic, line 14: I wouldn't execute this at every analyzer execution but periodically with a cron. Similarly to how we already update Yara Rules. In that way we don't slow the analysis

mlodic (changes requested):

api_app/analyzers_manager/file_analyzers/quark_engine.py
0ssigeno: This implementation is cluster safe 😿

0ssigeno: Plus the call of this method on an already updated repo requires ~2 seconds to execute. I think this will not be the bottleneck of the analyzer, while at the same time allowing us to have on **every** worker container an updated repository

mlodic: you brought out a point that makes sense to me: this won't work correctly in a cluster implementation.
However, to provide cluster support, we can't just do the same for all the Yara analyzers because that would be really strong in terms of wasted resources and would waste a lot of time during each analysis. See this issue: https://github.com/intelowlproject/IntelOwl/issues/1474

I think that for the issue of just updating quark-engine, it would make more sense to add it as a cron, because, once we find a centralized way to store the rules, those will be managed by a cron anyway. So there will be a day where both single and cluster deployments will leverage crons. So we can already create it from this PR

PR #1513, opened 2023-02-13
Fix plugins org

mlodic:


api_app/views.py
```diff
@@ -697,9 +697,15 @@ def get_queryset(self):
...
         if membership.exists():
-            result |= PluginConfig.objects.filter(
-                organization=membership[0].organization,
+            user_membership = membership[0]
+            # only the owner can view the secret
+            extracted = PluginConfig.objects.filter(
+                organization=user_membership.organization,
             )
+            if not user_membership.is_owner:
+                extracted.update(value="redacted")
```
0ssigeno, line 706: You are actually change the value in the db... we do not want that, we just want to change the value returned to the frontend

api_app/views.py
```diff
@@ -697,9 +697,15 @@ def get_queryset(self):
             user=self.request.user, organization__isnull=False
         )
         if membership.exists():
-            result |= PluginConfig.objects.filter(
-                organization=membership[0].organization,
+            user_membership = membership[0]
```
0ssigeno, line 700: if you are using the `[0]`, you can use `try/catch Membership.objects.get(query)`

0ssigeno (changes requested):

api_app/urls.py
```diff
@@ -54,6 +54,6 @@
     path("", include("certego_saas.apps.notifications.urls")),
     # organization sub-app
     path("me/", include("certego_saas.apps.organization.urls")),
-    path("plugin-disable/", plugin_state_viewer, name="plugin_state_viewer"),
+    path("plugin-disable", plugin_state_viewer, name="plugin_state_viewer"),
```
0ssigeno, line 57: You can revert this

0ssigeno (changes requested):

PR #1560, opened 2023-02-25
added user agent for crowdsec integration

rr404:


0ssigeno:
Can you point me where in their documentation is required a custom user agent to request their feeds?

mlodic:
> Can you point me where in their documentation is required a custom user agent to request their feeds?

@0ssigeno that was an explicit request from CrowdSec, which I am in contact with. In any case, that does make sense for any service because it allows the service to track the IntelOwl users. As integrators of external services, that is the minimum we should guarantee when we can.

mlodic:
Anyway, I'll close this because it has some problems (linters/version) and I'll make a new PR with the fixes

PR #1599, opened 2023-03-15
Fix "value too great for base" error

kyletaylored:
I keep running into this issue:

```
./initialize.sh: line 30: 10#v2: value too great for base (error token is "10#v2")
./initialize.sh: line 35: 10#v2: value too great for base (error token is "10#v2")
```

It looks like the issue is also mentioned in #1170, but I did not see the fix in the commit that closed it. I'm attaching a PR that would help filter out the v character when comparing versions.


mlodic:
thank you!

PR #1600, opened 2023-03-15
Fix logrotate copy command

kyletaylored:
Running into this:

```
cp: /etc/logrotate.d is not a directory
```

It looks like when calling the script, it's not aware of the context of it's location. Adding in a directory reference helps. I did have to add `sudo`, because without it would not let me create the `/etc/logrotate.d` directory. (see below)

```
cp: /etc/logrotate.d: Permission denied
cp: /Users/<USER>/IntelOwl/docker/scripts/logrotate: unable to copy extended attributes to /etc/logrotate.d: Permission denied
cp: /etc/logrotate.d/docker: No such file or directory
cp: /etc/logrotate.d/nginx: No such file or directory
cp: /etc/logrotate.d/application: No such file or directory
```


mlodic:
hey, thank you again for reporting!

The installer should be executed from the context of IntelOwl directory .

From there, the script was meant to be executed from the `./docker/scripts` subdirectory, as stated in the docs: https://intelowl.readthedocs.io/en/latest/Installation.html#logrotate-configuration-strongly-recommended

In that case there should no need to change the directory.

Plus, the `/etc/logrotate.d` should exist too, otherwise you would need to install Logrotate first, right?

kyletaylored:
Right, when I ran the `./initialize.sh` script, I saw the reference to the `./install_logrotate.sh` script, and when that ran, it failed. If I run the init script fresh, this is the output:

```
➜  IntelOwl git:(develop) ./initialize.sh
This script will check (and possibly guide you through) the installation of dependencies for IntelOwl!
CARE! This script is delivered AS IS and could not work correctly in every possible environment. In the case you face any error, you should just follow the official documentation and do all the required operation manually.
Docker version 20.10.22 detected
Docker-compose version 2.15.1 detected
Python3 version 3.10.10 detected
pip3 is installed
Installing python dependencies using pip...
... (things in here for requirements)

[notice] A new release of pip available: 22.3.1 -> 23.0.1
[notice] To update, run: /opt/homebrew/opt/python@3.11/bin/python3.11 -m pip install --upgrade pip
Python dependencies installed!
Adding Logrotate configuration to Systems logrotate
cp: /etc/logrotate.d is not a directory
```

If I run the `install_logrotate` script directly, I get the same output:
```
➜  scripts git:(develop) ./install_logrotate.sh
cp: /etc/logrotate.d is not a directory
```

So then if I manually create the directory and run it again...
```
➜  scripts git:(develop) ./install_logrotate.sh
cp: /etc/logrotate.d is not a directory
➜  scripts git:(develop) mkdir /etc/logrotate.d
mkdir: /etc/logrotate.d: Permission denied
➜  scripts git:(develop) sudo mkdir /etc/logrotate.d
➜  scripts git:(develop) ./install_logrotate.sh
cp: /etc/logrotate.d/application: Permission denied
cp: /etc/logrotate.d/docker: Permission denied
cp: /etc/logrotate.d/nginx: Permission denied
```

This was after installing logrotate on my machine as well, so my only assumption for a fix would be to fix the `cp` command to copy all contents of the logrotate directory into `/etc/logrotate.d`, which required sudo to create a directory in the `/etc` path. But now that I'm looking at it, the directory context isn't required, it's just the path reference in the `cp` command itself.

```
sudo cp -r logrotate /etc/logrotate.d
```

mlodic:
merged!

PR #1625, opened 2023-03-31
Visualizer Framework

0ssigeno:


intel_owl/wsgi.py
```diff
@@ -1,6 +1,9 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
```
mlodic, line 5: there are several duplicates

PR #1649, opened 2023-04-11
Dnstwist update

mlodic:
* updated dependencies
* moved from subprocess to library
* adjusted all the parameters
* other tweaks


mlodic:
to adjust the migraiton number

api_app/analyzers_manager/observable_analyzers/dnstwist.py
```diff
@@ -49,50 +41,34 @@ def run(self):
...
 
     @classmethod
     def _monkeypatch(cls):
         patches = [
             if_mock_connections(
                 patch(
-                    "subprocess.Popen",
-                    return_value=MockPopen(),
+                    "dnstwist.run",
+                    return_value=MockUpResponse({}, 200),
```
0ssigeno, line 71: You should just return a `dict`, not a `MockUpResponse`

0ssigeno (changes requested):

PR #1679, opened 2023-05-02
Param as model

0ssigeno:


api_app/core/serializers.py
```diff
@@ -35,63 +21,155 @@ class _ConfigSerializer(rfs.Serializer):
...
+            else:
+                disabled = True
+            plugin_representation["disabled"] = disabled
+            plugin_representation["verification"] = {
+                "configured": configured,
+                "details": details,
+                "missing_secrets": parameter_required_not_configured,
+            }
+            result.append(plugin_representation)
+
```
mlodic, line 162: log of the number of generated plugin representations per user?

api_app/core/serializers.py
```diff
@@ -35,63 +21,155 @@ class _ConfigSerializer(rfs.Serializer):
...
+            )
+        )
+        parsed = defaultdict(list)
+        # populate the result for every plugin (even the ones without parameters)
+        # parsed[plugin]= [parameter1]
+        for parameter in params:
+            parsed[getattr(parameter, self.child.Meta.model.snake_case_name)].append(
+                parameter
+            )
+
```
mlodic, line 110: log of user and number of params?

intel_owl/settings/__init__.py
```diff
@@ -41,9 +41,9 @@
     # auth
     "rest_email_auth",
     "drf_recaptcha",
+    # performance debugging
```
mlodic, line 44: leftover?

api_app/core/serializers.py
0ssigeno: It is always the number of configurations present in the db. I just that some of them are disabled

PR #1739, opened 2023-06-15
Pivot framework

0ssigeno:


mlodic (approved):
Great! I created a Trello card for the frontend/docs part so the next Sprint we'll think about adding that

PR #1740, opened 2023-06-15
More status mgmt

### LINKED ISSUE #1707: Final Job Statuses are not enough for the Plugin pipeline
Originally, Connectors and Visualizers did not exist. Now, we have different plugins and we'll add more in the future.

The original list of available statuses for a Job is not enough to represent the required statuses of the job. Right now there is a final status once the analyzers have finished to be executed. In this way, we lose the information regarding the fact that the connectors and the visualizers (or any other future plugin) have finished, independently from their order.

Actual order is:
analyzers -> connectors -> visualizers

Actual list:

```
class Status(models.TextChoices):
    PENDING = "pending", "pending"
    RUNNING = "running", "running"
    REPORTED_WITHOUT_FAILS = "reported_without_fails", "reported_without_fails"
    REPORTED_WITH_FAILS = "reported_with_fails", "reported_with_fails"
    KILLED = "killed", "killed"
    FAILED = "failed", "failed"

    @classmethod
    def final_statuses(cls) -> typing.List["Status"]:
        return [
            cls.REPORTED_WITHOUT_FAILS,
            cls.REPORTED_WITH_FAILS,
            cls.KILLED,
            cls.FAILED,
        ]
```

An important thing to keep in mind is that it would be cool to maintain retrocompatibility with old clients to avoid to break them with this change.

0ssigeno:
Closes #1707


api_app/core/views.py
```diff
@@ -59,14 +61,32 @@ def perform_kill(report: AbstractReport):
...
+        # clean up job
 
-    def perform_retry(self, report: AbstractReport):
-        raise NotImplementedError()
+        job = Job.objects.get(pk=report.job.pk)
+        job.set_final_status()
+
+    @staticmethod
+    def perform_retry(report: AbstractReport):
+        signature = report.config.get_signature(
```
mlodic, line 78: can we add a line of log?

api_app/models.py
```diff
@@ -232,106 +232,111 @@ def retry(self):
...
-        )
+        for config in [AnalyzerConfig, ConnectorConfig, VisualizerConfig]:
+            reports = self._get_config_reports(config).filter(
+                status__in=[
+                    AbstractReport.Status.PENDING,
+                    AbstractReport.Status.RUNNING,
+                ]
+            )
+
+            ids = list(reports.values_list("task_id", flat=True))
```
mlodic, line 334: can we add a line of log?

api_app/choices.py
```diff
@@ -32,11 +32,24 @@ def get_priority(cls, tlp):
 class Status(models.TextChoices):
     PENDING = "pending", "pending"
     RUNNING = "running", "running"
+
+    ANALYZERS_COMPLETE = "analyzers_complete", "analyzers_complete"
```
mlodic, line 36: is maybe better to use `..ers_completeD`? Same for all the others.

Plus, would make sense to add also "CONNECTORS_RUNNING" and "VISUALIZERS_RUNNING" or do you think that this is implicit in the previous statuses?

0ssigeno: It is kinda implicit, but I can see users that does not know the framework well enough to ask themselves "what is going on?"
For this reason i think that adding the intermediary status, its actually a good idea, and does not provoke any side effect at all.

api_app/core/views.py
0ssigeno: The log is inside the view that call this method. I think that it is enough

api_app/models.py
0ssigeno: Agree

PR #1755, opened 2023-06-22
Visualizer connected to playbook

### LINKED ISSUE #1704: Visualizers should be connected to Playbooks instead of Analyzers and Connectors separately
Working with Visualizers, we have reached the conclusion that each visualizer should be the representation of a single playbook, and not the visualization of analyzers and connectors. For this reason, we should change the mapping in the backend

0ssigeno:
Closes #1704


mlodic (approved):
I have just added little adjustments and tweaked a migration which is still not in `master`.

Issue #1470, opened 2023-02-01
Allow chance to have more than one org owner


mlodic:
Right now Organization can have only one owner (the creator), who is able to delete the organization, invite and remove members.

We should allow organizations to:
* have one or more owners
* owners can promote other members to owner or demote owners to members

Most of the changes (the backend part) should be done in the [certego-saas](https://github.com/certego/certego-saas) package

mlodic:
added with #51

PR #1837, opened 2023-08-09
added support admins pluginconfig managemnent #1470

federicofantini:
# Description

Added support to admin permission management plugin config: now you don't need to be an owner but just be an admin of an organization.
NOTE: this update do not cover the frontend.

## Type of change

# Checklist

- [X] I have added tests for the feature/bug I solved (see `tests` folder). All the tests (new and old ones) gave 0 errors.

### DIFF
--- a/api_app/permissions.py
+++ b/api_app/permissions.py
@@ -19,11 +19,15 @@ class IsObjectRealOwnerPermission(BasePermission):
 class IsObjectAdminPermission(BasePermission):
     @staticmethod
     def has_object_permission(request, view, obj):
-        if request.user.has_membership():
-            return request.user.membership.is_admin
-        return False
-
-
-IsObjectRealOwnerOrAdminPermission = (
-    IsObjectRealOwnerPermission | IsObjectAdminPermission  # pylint: disable=E1131
-)
+        obj_owner = getattr(obj, "owner", None)
+        # if the object was not made for an organization, we return false
+        if not obj_owner or not obj_owner.has_membership():
+            return False
+        else:
+            # if we are admin we can e have to check that is our org
+            return (
+                request.user.has_membership()
+                and request.user.membership.is_admin
+                and obj_owner.membership.organization
+                == request.user.membership.organization
+            )
--- a/api_app/permissions.py
+++ b/api_app/permissions.py
@@ -14,3 +14,16 @@ class IsObjectRealOwnerPermission(BasePermission):
         if obj_owner := getattr(obj, "owner", None):
             return obj_owner == request.user
         return False
+
+
+class IsObjectAdminPermission(BasePermission):
+    @staticmethod
+    def has_object_permission(request, view, obj):
+        if request.user.has_membership():
+            return request.user.membership.is_admin
+        return False
+
+
+IsObjectRealOwnerOrAdminPermission = (
+    IsObjectRealOwnerPermission | IsObjectAdminPermission  # pylint: disable=E1131
+)
--- a/api_app/permissions.py
+++ b/api_app/permissions.py
@@ -8,14 +8,6 @@ from rest_framework.permissions import BasePermission
 logger = getLogger(__name__)
 
 
-class IsObjectRealOwnerPermission(BasePermission):
-    @staticmethod
-    def has_object_permission(request, view, obj):
-        if obj_owner := getattr(obj, "owner", None):
-            return obj_owner == request.user
-        return False
-
-
 class IsObjectAdminPermission(BasePermission):
     @staticmethod
     def has_object_permission(request, view, obj):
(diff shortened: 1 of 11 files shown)


mlodic:
I leave the review to @0ssigeno only. Thanks for the contribution :)

api_app/queryset.py
```diff
@@ -210,11 +210,17 @@ def visible_for_user(self, user: User = None) -> "PluginConfigQuerySet":
...
+                    return self.filter(Q(for_organization=True) | Q(owner__isnull=True))
+                else:
+                    return self.filter(
+                        (
+                            Q(for_organization=True)
+                            & Q(owner=membership.organization.owner)
+                        )
+                        | Q(owner=user)
+                        | Q(owner__isnull=True)
+                    )
```
0ssigeno, line 223: Unfortunately, with this condition, if you are an `admin` you can see the configuration of **every** other organization user values. This is obviously not intended. 
1 - Can we add a test to verify or refute this claim?
2 - If the claim is valid, I would suggest something like
```python3
if membership.is_admin:

    Q(for_organization=True, owner__membership__organization=membership.organization) | 
    Q(owner=user) | Q(owner_isnull=True)
```
The test that you will make should fail before and pass after this change

api_app/permissions.py
```diff
@@ -14,3 +14,16 @@ def has_object_permission(request, view, obj):
...
             return obj_owner == request.user
         return False
+
+
+class IsObjectAdminPermission(BasePermission):
+    @staticmethod
+    def has_object_permission(request, view, obj):
+        if request.user.has_membership():
+            return request.user.membership.is_admin
+        return False
```
0ssigeno, line 24: If you are admin you can retrieve the value of another organization.
The check should be something like
```python3
 obj_owner = getattr(obj, "owner", None):
 # if the object was not made for an organization, we return false
if not obj_owner or not obj_owner.has_membership():
    return False
 else:
    # if we are admin we can e have to check that is our org
    return request.user.has_membership() and request.user.membership.is_admin and obj_owner.membership.organization == request.user.membership.organization
```

api_app/views.py
```diff
@@ -760,7 +760,12 @@ class AbstractConfigViewSet(viewsets.ReadOnlyModelViewSet, metaclass=ABCMeta):
     def disable_in_org(self, request, pk=None):
         logger.info(f"get disable_in_org from user {request.user}, name {pk}")
         obj: AbstractConfig = self.get_object()
-        if not request.user.has_membership() or not request.user.membership.is_owner:
+        if request.user.has_membership():
+            if not (
+                request.user.membership.is_owner or request.user.membership.is_admin
```
0ssigeno, line 765: You can remove the `is_owner` chec, since the owner is **always** admin

api_app/serializers.py
```diff
@@ -921,9 +929,18 @@ def validate(self, attrs):
...
+            org = attrs.pop("organization")
+            if not (
+                org.owner == attrs["owner"]
+                or (
+                    self.context["request"].user.has_membership()
+                    and self.context["request"].user.membership.organization.pk
+                    == org.pk
+                    and self.context["request"].user.membership.is_admin
+                )
+            ):
```
0ssigeno, line 941: Ok we starts to need some sort of comment in this boolean expression because it is breaking my mind.

This should mean
```

we raise ValidationError() when
1 - we are not owner  OR
2 - we are not admin of the same org
```
If I did not fucked up the boolean expression, the check seems valid to me, I would just suggest to add a comment because it took me 5 minutes to understand what was going on

api_app/serializers.py
```diff
@@ -882,7 +882,15 @@ def get_attribute(self, instance: PluginConfig):
...
-                and self.context["request"].user.pk != instance.owner.pk
+                and not (
+                    self.context["request"].user.pk == instance.owner.pk
+                    or (
+                        self.context["request"].user.has_membership()
+                        and self.context["request"].user.membership.organization.pk
+                        == instance.owner.membership.organization.pk
+                        and self.context["request"].user.membership.is_admin
+                    )
+                )
```
0ssigeno, line 893: We need a comment here too, this is fucking me up.
```
We return `redacted` when
1) is a secret AND
2) is a value for the organization AND
3) we are not its owner  OR
4) we are not an admin of the same organization
```
if I did not fuck up the boolean expression, it seems good to me, just add a comment for `future us`

0ssigeno (changes requested):

api_app/queryset.py
federicofantini: I was reasoning that in this case being admin/owner/user is indifferent to the level of visibility of the configs in the org...
I think the following code should be enough:
```python
# If you are member of an organization you should see the configs.
return self.filter(
    Q(for_organization=True, owner__membership__organization=membership.organization)
    | Q(owner=user)
    | Q(owner__isnull=True)
)
```

0ssigeno (approved):

0ssigeno:
LGTM. The only thing missing is to allow the org owner to add/remove admins from the org panel in the frontend.
I will open a different issue to keep track of this.

PR #1881, opened 2023-09-06
fixed recent scans 

carellamartina:
# Description
fixed recent scans 

## Type of change

# Checklist

### Important Rules


frontend/src/components/scan/utils/RecentScans.jsx
```diff
@@ -94,39 +91,60 @@ RecentScansCard.defaultProps = {
...
-  const [recentScans, LoaderRecentScans] = useAxiosComponentLoader({
-    url: JOB_RECENT_SCANS,
-    method: "POST",
-    data: { md5: fileMd5.length ? fileMd5 : md5(param) },
-  });
+
+  React.useEffect(() => {
+    fetchRecentScansUser();
+    fetchRecentscans(fileMd5.length ? fileMd5 : md5(param));
+  }, [fetchRecentScansUser, fetchRecentscans, fileMd5, param]);
```
drosetti, line 131: I suggest to split this useEffect in two: we will have one useEffect only for the intial fetching (user) and another use effect only for the observable. In this way we will reduce the number of requests to the backend. I think there shouldn't be drawbacks: in case you start a scan and come back to the scan page the initial request should be done the same. What do you think ?

frontend/src/stores/useRecentScansStore.jsx
```diff
@@ -0,0 +1,39 @@
+import { create } from "zustand";
+import axios from "axios";
+
+import { JOB_RECENT_SCANS, JOB_RECENT_SCANS_USER } from "../constants/api";
+
+const useRecentScansStore = create((set, _get) => ({
+  loadingScansUser: true,
+  loadingScans: true,
```
drosetti, line 8: Silly requests: I suggest to rename this and the other field to `loadingScansInsertedAnalyzable` or something like this

frontend/src/components/scan/utils/RecentScans.jsx
```diff
@@ -94,39 +91,60 @@ RecentScansCard.defaultProps = {
...
 
-  const allRecentScans = recentScans.length
-    ? recentScans.concat(recentScansUser)
-    : recentScansUser;
+  // remove duplicate job
+  const allRecentScans = Array.from(
+    [...recentScans, ...recentScansUser]
+      .reduce((m, scan) => m.set(scan.pk, scan), new Map())
+      .values(),
+  );
```
drosetti, line 141: Awesome use of JS!  :)

frontend/src/components/scan/utils/RecentScans.jsx
drosetti, line 131: I have another request about this: currently we perform to the backend a request for each char inserted, to avoid to flood the backend with a lot of requests (and lot of them are useless: if the user it typing "google.com" we will made some requests for: g, go, goo, etc...) we could try to think about a mechanism to wait the user terminated to write and then perform the request. I'm thinking about something like a delay: wait 1,2 second before send the request and send only one request with all the text. We could implement it ore use a library (the service portal use lodash)

drosetti (changes requested):

frontend/src/components/scan/utils/RecentScans.jsx
mlodic: yep that is very important to save network and backend computation

drosetti (approved):

PR #1908, opened 2023-09-27
Playbook deletion + Actions column in Plugins page

carellamartina:
# Description

Added the button to delete playbooks for admin users.
Added "actions" column for plugins tables

## Type of change

# Checklist

### Important Rules


carellamartina:
Schermata del 2023-09-27 17-38-13
Schermata del 2023-09-27 17-38-32

tests/__init__.py
```diff
@@ -240,10 +240,14 @@ def test_update(self):
     def test_delete(self):
         plugin = self.model_class.objects.order_by("?").first().pk
         response = self.client.delete(f"{self.URL}/{plugin}")
-        self.assertEqual(response.status_code, 405, response.json())
+        if "playbook" in self.URL:
+            self.assertEqual(response.status_code, 204, response)
+        else:
+            self.assertEqual(response.status_code, 405, response.json())
```
0ssigeno, line 246: instead of having this `if` condition, just override the method in the `PlaybookConfigViewSet` test class and change the `assertEqual`

tests/__init__.py
```diff
@@ -240,10 +240,14 @@ def test_update(self):
...
         self.client.force_authenticate(self.superuser)
         response = self.client.delete(f"{self.URL}/{plugin}")
-        self.assertEqual(response.status_code, 405, response.json())
+        if "playbook" not in self.URL:
+            self.assertEqual(response.status_code, 405, response.json())
```
0ssigeno, line 250: instead of having this if condition, just override the method in the PlaybookConfigViewSet test class and change the assertEqual

api_app/playbooks_manager/views.py
```diff
@@ -20,7 +20,11 @@
 logger = logging.getLogger(__name__)
 
 
-class PlaybookConfigViewSet(AbstractConfigViewSet, mixins.CreateModelMixin):
+class PlaybookConfigViewSet(
+    AbstractConfigViewSet,
+    mixins.CreateModelMixin,
+    mixins.DestroyModelMixin,
+):
```
0ssigeno, line 27: If you do not change the permissions here, every user can call the API directly and delete every playbook.

0ssigeno (changes requested):

api_app/playbooks_manager/views.py
```diff
@@ -30,6 +34,12 @@ def get_queryset(self):
             self.request.user
         ).prefetch_related("analyzers", "connectors", "visualizers", "tags", "pivots")
 
+    def get_permissions(self):
+        permissions = super().get_permissions()
+        if self.request.method in ["DELETE"]:
+            permissions.append((IsAdminUser)())
```
0ssigeno, line 40: ```permissions.append(IsAdminUser())``` :P

0ssigeno (approved):

drosetti (changes requested):
i noticed user owner cannot disable plugins: I created it with superuser command, create the company with the UI and the added an user with Django Admin, but as superuser I cannot disable a plugin. Also the error is a 403 from network tab (correct), but didn't show an error popup, I think is know because the button should be disabled, so it shouldn't be possible to perform this request.

I also noticed a bug not strictly requested but I ask you to fix it: the "save as a playbook" button in the job reports return an error in case you try to add a playbook ( I think is more related to the backend part) and want a min length to the playbook name, but it's not write in the form.

api_app/playbooks_manager/serializers.py
```diff
@@ -38,6 +38,20 @@ class Meta:
...
+        if instance.owner:
+            # if is a custom playbook,
+            # it is deletable by the owner of the playbook or by an admin
+            if (
+                instance.owner != self.context["request"].user
+                and not self.context["request"].user.membership.is_admin
+            ):
+                return False
+            return True
+        return False
```
0ssigeno, line 54: Three things
- if the user is System administrator (not org admin), every playbook is deletable at the moment (he could do it in django admin anyway)
- the second `if` condition, write it in a positive direction, otherwise my brain will explode
- the second part of the second `if` condition is missing a condition: `AND` the user requesting playbook `must` be of the same organization of the owner of the playbook. Otherwise an organization owner can delete playbooks of other organizations

0ssigeno (changes requested):

frontend/src/stores/usePluginConfigurationStore.jsx
```diff
@@ -237,6 +237,21 @@ const usePluginConfigurationStore = create((set, get) => ({
       return null;
     }
   },
+  deletePlaybook: async (playbook) => {
+    let success = false;
+    try {
+      const response = await axios.delete(
+        `${PLAYBOOKS_CONFIG_URI}/${playbook}`,
+      );
+      success = response.status === 204;
```
drosetti, line 246: not sure about this "if", is it possible we get a different status code ? ex: 201

frontend/src/components/jobs/result/utils/SaveAsPlaybooksForm.jsx
```diff
@@ -60,20 +60,22 @@ export function SaveAsPlaybookForm({ onFormSubmit }) {
...
                 </Label>
                 <Input
                   autoFocus
                   id="forminput-name"
                   type="text"
                   name="name"
+                  onBlur={formik.handleBlur}
                   onChange={formik.handleChange}
                 />
+                {formik.touched.name && <small>{formik.errors.name}</small>}
```
drosetti, line 74: I suggest to write this test message in red, also the description field is required

drosetti (changes requested):
* Two little changes on the frontend.
* I suggest to add some tests in the backend part about multi-tenancy:  I'd like to have some tests where we can check the users can do only permitted actions and in case they try to do something else they receive an error. Ex: create an org with an admin and a normal user and test normal user cannot delete a playbook created by the admin. Or an user out of the org doesn't view a playbook of an org and so on.
* I don't remember if this should be done or it's a future development, should the admins have a button in the frontend to make a playbook available to all the org's users ?

PR #1947, opened 2023-10-18
Timeout capesandbox analyzer

federicofantini:
# Description
added timeout on every request of capesandbox analyzer + start the poll for result after the analysis is ended

## Type of change

# Checklist

- [X] A new plugin (analyzer, connector, visualizer, playbook or ingestor) was added or changed, in which case:

### DIFF
--- a/api_app/analyzers_manager/file_analyzers/cape_sandbox.py
+++ b/api_app/analyzers_manager/file_analyzers/cape_sandbox.py
@@ -209,7 +209,7 @@ class CAPEsandbox(FileAnalyzer):
     ) -> dict:
         logger.info(
             f"Job: {self.job_id} ->"
-            "Sleeping for the entire duration of the analysis: {self.timeout}s"
+            f"Sleeping for the entire duration of the analysis: {self.timeout}s"
         )
         time.sleep(self.timeout)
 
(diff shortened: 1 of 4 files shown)


api_app/analyzers_manager/file_analyzers/cape_sandbox.py
```diff
@@ -204,17 +207,51 @@ def __poll_for_result(
...
+        #   empirical value of processing duration time
+        #   real polling with max tries
+        #   final attempts in case there is a bottleneck
+        timeout_attempts = (
+            [60] * (self.timeout // 60)
+            + [self.timeout % 60]
+            + [50]
+            + [self.poll_distance] * self.max_tries
+            + [60] * 3
+        )
```
0ssigeno, line 227: What the fuck is this ahahahaha

0ssigeno (approved):
lgtm, just fix the migration file (you just need to change the number and its dependency to the latest in develop)

api_app/analyzers_manager/file_analyzers/cape_sandbox.py
federicofantini: I'd like to have a point interpolation function but those are empirical magic numbers as to when you are most likely to get a result from the analysis.

federicofantini: I could handle it by doing many `if` inside the `for` but I preferred to enclose everything in this data structure.
I would greatly appreciate suggestions because I know the code isn't easily readable like that

federicofantini: I was thinking if there was a way to tell Intelowl that the analysis is finished instead of doing polling, or maybe a mix of the two

PR #1949, opened 2023-10-19
analyzer support for ipv6

### LINKED ISSUE #1885: Support for ipv6 addresses
We should add the chance to analyze ipv6 addresses like any other ipv4 address. We should fix it both in the frontend and the backend regex.

Plus, we should check out the integrated services to understand whether they need some internal change or not to support it (maybe some different functions in the libraries)

shivam-Purohit:
IPv6 closes #1885

# Description
Add ipv6 support for the analyzers 

## Type of change

# Checklist

- [] If the GUI has been modified:

### Important Rules


shivam-Purohit:
I am getting positive response for ipapi but most analyzers do not support ipv6. I am confused if I have to add some more support in the backend as of now no tests are failing.

one more thing
when I was installing the app I had to make some changes in the Dockerfile.  I had to change 
```
RUN npm install npm@latest --location=global \
    && npm install \
    && PUBLIC_URL=/static/reactapp/ npm run build
```

i removed the `npm install npm@latest --location=global`  as it was installing some version that was not compatible.
This was the error i was getting

```
1.879 npm ERR! engine Not compatible with your version of node/npm: npm@10.2.1

1.879 npm ERR! notsup Not compatible with your version of node/npm: npm@10.2.1

1.879 npm ERR! notsup Required: {"node":"^18.17.0 || >=20.5.0"}

1.879 npm ERR! notsup Actual:   {"npm":"8.11.0","node":"v16.15.1"}


```

i am running node 18.18.2 and npm 10.2.0. I don't know why its not showing the actual versions. I cannot build the app, I have to remove the npm update line.

mlodic:
yep the changes requested are more complicated than what you have done, you surely need to touch the backend and fix the regex there too.

Plus we would need to test the majority of analyzer which support IP addresses now (at least the ones in the "FREE_TO_USE_ANALYZERS" playbook) and see if they need some changes. Otherwise, this feature could be a double-edged sword if every analyzer fails. That would be a very bad user experience

shivam-Purohit:
> yep the changes requested are more complicated than what you have done, you surely need to touch the backend and fix the regex there too.

I have gone through the backend  We are defining the observable type in api_app/analyzer_manager/constants.py.

```
    def calculate(cls, value: str) -> str:
        """Returns observable classification for the given value.\n
        Only following types are supported:
        ip, domain, url, hash (md5, sha1, sha256), generic (if no match)

        Args:
            value (str):
                observable value
        Returns:
            str: one of `ip`, `url`, `domain`, `hash` or 'generic'.
        """
        try:
            ipaddress.ip_address(value)
        except ValueError:
            if re.match(
                r"^.+://[a-z\d-]{1,200}"
                r"(?:\.[a-zA-Z\d\u2044\u2215!#$&(-;=?-\[\]_~]{1,200})+"
                r"(?::\d{2,6})?"
                r"(?:/[a-zA-Z\d\u2044\u2215!#$&(-;=?-\[\]_~]{1,200})*"
                r"(?:\.\w+)?",
                value,
            ):
                classification = cls.URL
            elif re.match(
                r"^([\[\\]?\.[\]\\]?)?[a-z\d-]{1,63}"
                r"(([\[\\]?\.[\]\\]?)[a-z\d-]{1,63})+$",
                value,
                re.IGNORECASE,
            ):
                classification = cls.DOMAIN
            elif (
                re.match(r"^[a-f\d]{32}$", value, re.IGNORECASE)
                or re.match(r"^[a-f\d]{40}$", value, re.IGNORECASE)
                or re.match(r"^[a-f\d]{64}$", value, re.IGNORECASE)
            ):
                classification = cls.HASH
            else:
                classification = cls.GENERIC
                logger.info(
                    "Couldn't detect observable classification"
                    f" for {value}, setting as 'generic'"
                )
        else:
            # it's a simple IP
            classification = cls.IP

        return classification

```
we are using ipaddress library to check if its a valid IP address. As I can see [here](https://docs.python.org/3/library/ipaddress.html) it supports both ipv4 and ipv6 addresses. So there shouldn't be an issue checking for v6 as well. As we in this function checking its against ipaddress library and throwing error if it fails to  recognise.  I dont see any other files were we are checking it against regular expression.


Also I ran a test IP against the playbook free analyzers. All of them are showing status 200 with some showing response while others returning no results mostly as they dont support ipv6.
I dont think we would need to check for each analyzer as it would have at least thrown error as unsupported IP addresses. If its not then I think I would have to move to each analyzer documentation to check for ipv6 info.
ubuntuMachine  Running  - Oracle VM VirtualBox 21-10-2023 19_15_51

mlodic:
thank you for that analysis, really appreciated.

PR #1958, opened 2023-10-24
Pivot visualizer

0ssigeno:


drosetti (approved):

api_app/visualizers_manager/migrations/0030_visualizer_config_pivot.py
```diff
@@ -0,0 +1,80 @@
+from django.db import migrations
+from django.db.models.fields.related_descriptors import (
+    ForwardManyToOneDescriptor,
+    ForwardOneToOneDescriptor,
+    ManyToManyDescriptor,
+)
+
+plugin = {'name': 'Pivot', 'python_module': {'module': 'pivot.Pivot', 'base_path': 'api_app.visualizers_manager.visualizers'}, 'description': 'Visualizer for pivot', 'disabled': False, 'soft_time_limit': 60, 'routing_key': 'default', 'playbooks': ['Dns', 'FREE_TO_USE_ANALYZERS', 'Popular_IP_Reputation_Services', 'Popular_URL_Reputation_Services', 'Sample_Static_Analysis'], 'model': 'visualizers_manager.VisualizerConfig'}
```
mlodic, line 8: just a question: does the `playbook` field implicates that we should add to every new public playbook a reference to this Pivot plugin to have it executed, right? In that case, we should add a reference somewhere in the doc and in the pr template to avoid forgetting it

0ssigeno: It depends: if the playbook does not have any pivot, there are no reasons to add the visualizer because it will always display an empty page. This could confuse the users more than help them

mlodic: so I didn't get why these playbooks are referred here considering they don't have any pivot

mlodic (approved):

PR #1991, opened 2023-11-14
frontend - playbook runtime configuration

carellamartina:
# Description

- added runtime configuration icon in the scan page for analysis with playbooks
- moved the tlp section out of the "advanced settings" section

## Type of change

# Checklist

    - [ ] I have dumped the configuration from Django Admin using the `dumpplugin` command and added it in the project as a data migration. [Doc](["How to create a Plugin"](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-add-a-new-plugin))

### Important Rules

### DIFF
--- a/frontend/src/components/plugins/types/utils.jsx
+++ b/frontend/src/components/plugins/types/utils.jsx
@@ -242,7 +242,7 @@ export function PlaybooksCollapse({ value, pluginType_ }) {
   );
 }
 
-export function PlaybookPluginList({ pluginInfo, pluginType_ }) {
+function PlaybookPluginList({ pluginInfo, pluginType_ }) {
   const plugin = `${pluginType_}s`;
   return (
     <div>
--- a/frontend/src/components/plugins/types/utils.jsx
+++ b/frontend/src/components/plugins/types/utils.jsx
@@ -49,7 +49,7 @@ export function PluginInfoCard({ pluginInfo }) {
       <CardBody className="bg-darker border-top border-tertiary">
         <div>
           <h6 className="text-secondary">Description</h6>
-          <p>{markdownToHtml(pluginInfo?.description)}</p>
+          <span>{markdownToHtml(pluginInfo?.description)}</span>
         </div>
         <div>
           <div>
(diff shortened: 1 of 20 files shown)


drosetti (changes requested):
`frontend/src/components/plugins/types/utils.jsx`
* convert function PluginHealthSpinner (called only once) into it's return where needed
* Move the button components into another file

frontend/src/components/scan/utils/RuntimeConfigurationModal.jsx
```diff
@@ -7,123 +7,137 @@ import { ContentSection, CustomJsonInput } from "@certego/certego-ui";
...
+            ...singlePluginConfig,
+            [paramName]: paramValue,
+          }),
+          {},
+        ),
+        ...(formik.values.runtime_configuration[pluginType] || {}),
+      }),
+      {},
+    );
+  });
```
drosetti, line 119: it's correct but I think a reduce inside another reduce is quite difficult to read, I suggest to use forEach instead, I hope it will be more readable.

frontend/src/components/plugins/types/utils.jsx
```diff
@@ -388,90 +242,41 @@ export function PlaybooksCollapse({ value, pluginType_ }) {
...
-    try {
-      await deletePlaybook(playbookName);
-      setShowModal(false);
-      await retrievePlaybooksConfiguration();
-    } catch {
-      // handle error in deletePlaybook
-    }
-  };
-
+export function PlaybookPluginList({ pluginInfo, pluginType_ }) {
```
drosetti, line 245: This is used only in this file, we can remove the export

frontend/src/components/scan/utils/RuntimeConfigurationModal.jsx
```diff
@@ -7,123 +7,151 @@ import { ContentSection, CustomJsonInput } from "@certego/certego-ui";
...
+        Object.entries(pluginParams).forEach(
+          ([paramName, { value: paramValue }]) => {
+            editableConfig[pluginType][pluginName] = {
+              ...editableConfig[pluginType][pluginName],
+              [paramName]: paramValue,
+            };
+          },
+        );
+      },
+    );
```
drosetti, line 127: I didn't try, but I think this code could be simplified, ex:
```
editableConfig[pluginType] = {
          ...editableConfig[pluginType],
          [pluginName]: {},
        };
```
This shouldn't be required anymore: it creates a new dict each time and with unpack syntax (...) put in the new dict the old data in this way it doesn't lost the previous iterations results, but with for each I think we could move the initialization before the loop and inside the loop add only the new key.

drosetti (changes requested):
When i type an observable (ex: dns.google.com) auto loads playbook DNS but when i click on the runtime_configuration button i get an exception:
```
TypeError: Cannot convert undefined or null to object
```

drosetti (approved):

PR #2028, opened 2023-12-18
Dashboard improv: add button to filter for the user's org only

### LINKED ISSUE #1936: Dashboard improv: add button to filter for the user's org only
Right now the dashboard provides generic data from all instance's users. It would be useful to get statistics only from the logged user's organization only.

We could add a toggle button at the top right of the page that perform this change. This would require a little change in the backend too to filter the data by the selected org. It is enough to use a toggle button because an user can be part of a single organization only.

shivam-Purohit:
Button to switch between org only reports or all the reports 
closes #1936

# Description

## Type of change

# Checklist

    - [ ] I have dumped the configuration from Django Admin using the `dumpplugin` command and added it in the project as a data migration. [Doc](["How to create a Plugin"](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-add-a-new-plugin))

### Important Rules


shivam-Purohit:
@mlodic its almost done now. But the problem is the response is cached so if I change the users the previous response is shown not the new one.

shivam-Purohit:
this is how the UI looks like. I am not good with names. We can get the switch which has label inside the switch only. That looks way cooler though we will need to add one more library for that.
ubuntuMachine  Running  - Oracle VM VirtualBox 22-12-2023 13_05_45

mlodic:
hello @shivam-Purohit, thank you for your PR! :)

> @mlodic its almost done now. But the problem is the response is cached so if I change the users the previous response is shown not the new one.

when you say "if I change the users", you mean if you logout and then login with another user?

> We can get the switch which has label inside the switch only. That looks way cooler though we will need to add one more library for that.

Do you mean something like that? 
image

This how we switch the visualization in the "Job Result" page. It could make sense to leverage the same component also here. Maybe, I'd also put that component at the left of the time picker, not at the right.

Thoughts?

api_app/views.py
```diff
@@ -478,22 +479,238 @@ def aggregate_md5(self, request):
...
+    #################################################################
+    def __aggregation_response_static(
+        self, annotations: dict, users=None, user=None
+    ) -> Response:
         delta, basis = self.__parse_range(self.request)
+        filter_kwargs = {"received_request_time__gte": delta}
+        if user:
+            filter_kwargs["user"] = user
+        if users:
+            filter_kwargs["user__in"] = users
```
mlodic, line 691: I think you can optimize the code by choosing a single parameter "users" here and change the filter based on the number of items in the list

api_app/views.py
```diff
@@ -478,22 +479,238 @@ def aggregate_md5(self, request):
...
+        except AttributeError:
+            # Handle the case where the user has no membership
+            logger.info("User has no membership.")
+
+        if organization:
+            return self.__aggregation_response_dynamic(
+                "md5", False, users=users_of_organization
+            )
+        else:
+            return self.__aggregation_response_dynamic("md5", False, user=current_user)
```
mlodic, line 680: we should call the aggregation method a single time, we need to optimize this code

api_app/views.py
```diff
@@ -478,22 +479,238 @@ def aggregate_md5(self, request):
...
+            if current_user.has_membership():
+                organization = current_user.membership.organization
+                if organization:
+                    users_of_organization = [
+                        membership.user for membership in organization.members.all()
+                    ]
+
+        except AttributeError:
+            # Handle the case where the user has no membership
+            logger.info("User has no membership.")
```
mlodic, line 673: this code has been replicated lots of time. We need it just once

frontend/src/components/dashboard/utils/charts.jsx
```diff
@@ -20,19 +20,38 @@ import {
...
+  const getValue = porp.key;
+  console.debug("props is ", getValue);
+  /* eslint-enable */
+
+  if (getValue) {
+    /* eslint-disable */
+    sendURL = JOB_AGG_STATUS_URI_ORG;
+    console.debug("this is sending to my urls", sendURL);
+    /* eslint-enable */
+  }
```
mlodic, line 50: more generally, I don't think it is necessary to create new API endpoints for this case. We should use the URL parameters to customize the results. This would avoid a lot of additional code that would be really difficult to maintain.

Example:
aggregate/status -> like it is right now
aggregate/status?org=True -> with the toggle enabled

mlodic (changes requested):
there is some duplicated code that must be addressed

frontend/src/components/dashboard/utils/charts.jsx
shivam-Purohit: Yeah I was thinking the same. I just tried some sub-optimised code to check the core functionality. I will improve the code quality.

shivam-Purohit:
> when you say "if I change the users", you mean if you logout and then login with another user?

yes same. To check / test the feature I created two more users and added one to my org. Though when i change the users  i.e logout and login again I get the results from the previous user. But if we restart the container the new results do re-render. Either it caches the result or there is some bug in my code?

shivam-Purohit:
> This how we switch the visualization in the "Job Result" page. It could make sense to leverage the same component also here. Maybe you can call the 2 options: "Org / Global" and make it "toggable" only if the user is part of an org. Otherwise it could be hidden or disabled.
> Maybe, I'd also put that component at the left of the time picker, not at the right.

yes this look better. I will try that. But to disable the option if user is not part of any org we will need to send request to the backend during the rendering of dashboard.
if so then we wont need to check if user is part of any org in backend.

mlodic:
> > This how we switch the visualization in the "Job Result" page. It could make sense to leverage the same component also here. Maybe you can call the 2 options: "Org / Global" and make it "toggable" only if the user is part of an org. Otherwise it could be hidden or disabled.
> > Maybe, I'd also put that component at the left of the time picker, not at the right.
> 
> yes this look better. I will try that. But to disable the option if user is not part of any org we will need to send request to the backend during the rendering of dashboard. if so then we wont need to check if user is part of any org in backend.

yep I think there is already something similar in place for this case in the "plugin" section:
image
The button related to org's config is shown only if the user belongs to an org

shivam-Purohit:
ubuntuMachine  Running  - Oracle VM VirtualBox 28-12-2023 12_47_19
ubuntuMachine  Running  - Oracle VM VirtualBox 28-12-2023 11_49_46

shivam-Purohit:
hey @mlodic hope the Christmas went well! 
i re-adjusted some code
changed the component to only render if the organisation exist for the user using `useOrganisationStore `components.
renamed new variables
and shifted some repeated code to new function

shivam-Purohit:
> ubuntuMachine Running - Oracle VM VirtualBox 28-12-2023 12_47_19 ubuntuMachine Running - Oracle VM VirtualBox 28-12-2023 11_49_46

this is how it looks. I have to change some layout to fit in new structure. If I just move the `buttongroup` component above the `timepicker ` in like the original code it was breaking like the buttongroup was moving to the left so i separated both the title and this components (timepicker and buttongroup).

mlodic:
hey, I hope this message finds you well :)

Unluckily I don't see the images that you linked in the last message. Could you please re-share them?

mlodic:
hey, I hope this message finds you well :)

I saw the images, that feels good!

api_app/views.py
```diff
@@ -60,7 +60,7 @@
     TagSerializer,
 )
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("__name__")
```
mlodic, line 63: why this change?

api_app/views.py
```diff
@@ -416,11 +416,17 @@ def pivot(self, request, pk=None, pivot_config_pk=None):
     )
     @cache_action_response(timeout=60 * 5)
     def aggregate_status(self, request):
+        org_param = request.GET.get("org", "").lower() == "true"
+        users_of_organization = None
+        if org_param:
+            users_of_organization = self.get_org_members(request.user)
```
mlodic, line 422: this could be a method returning `users_of_organization`, right?

mlodic (commented):
cool to me, do you want to make a little review @drosetti too?

frontend/src/components/dashboard/Dashboard.jsx
```diff
@@ -57,20 +61,91 @@ export default function Dashboard() {
...
+        </ButtonGroup>
         <ElasticTimePicker
           className="ms-auto"
           size="sm"
           defaultSelected={range}
           onChange={onTimeIntervalChange}
           id="Dashboard_timepicker"
         />
+
+      </div> */}
```
drosetti, line 103: remove this comment

frontend/src/components/dashboard/Dashboard.jsx
```diff
@@ -57,20 +61,91 @@ export default function Dashboard() {
...
+          <h3 className="fw-bold" id="Dashboard_title">
+            Dashboard
+          </h3>
+        </div>
+        <div className="d-flex flex-wrap align-items-baseline ">
+          {organization?.name ? (
+            <ButtonGroup className="mb-3">
+              <Button
+                outline={orgstate}
+                color={!orgstate ? "primary" : "tertiary"}
```
drosetti, line 116: I suggest to reverse the check, i think it's easier to understand:
`orgstate ? "tertiary" : "primary"`

frontend/src/components/dashboard/Dashboard.jsx
```diff
@@ -39,8 +39,12 @@ const charts2 = [
 ];
 
 export default function Dashboard() {
+  // const isSelectedUI = JobResultSections.VISUALIZER;
   const { guideState, setGuideState } = useGuideContext();
 
+  const [orgstate, setorgState] = useState(() => false);
```
drosetti, line 45: It's a very silly suggestion but I'd to have orgState, it has the same style of guideState (it's strange the linter doesn't mark this)

frontend/src/components/dashboard/utils/charts.jsx
```diff
@@ -27,12 +27,23 @@ const colors = getRandomColorsArray(10, true);
...
   console.debug("JobStatusBarChart rendered!");
+  /* eslint-disable */
+  var sendURL = JOB_AGG_STATUS_URI;
+  const parameter = props.sendProp;
+  const getValue = parameter.key;
+  console.debug("props is ", getValue);
+  /* eslint-enable */
+
+  /* eslint-disable */
+  sendURL = `${JOB_AGG_STATUS_URI}?org=${getValue}`;
```
drosetti, line 40: Pls use a more specific name for this const, what do you think about ORG_JOB_AGG_STATUS_URI ?

drosetti (changes requested):
I think you did a good job, I requested little changes (most of them on the names of the const/var). Be careful with `eslint-disable`, I know sometimes it's annoying, but it's useful to have same code style with many developers, it should be disabled only in rare cases.

api_app/views.py
shivam-Purohit: I changed it while I was trying to add a console logger. I was having some problem with log file to see the recent changes. I will revert it back.

api_app/views.py
shivam-Purohit: yes returns all the members in the organisation

frontend/src/components/dashboard/utils/charts.jsx
shivam-Purohit: sure!

shivam-Purohit:
@drosetti I have made the changes requested. One more thing I noticed is in the` aggregate_observable_classification`.Its counts the new ipv6 addressing IPs in the generic and not in the IP classification. That also reflect in the chart named `Job: Frequent IPs, Hash & Domains` which also does not count ipv6 IPs.
I can create a new issue regarding this and if it is not on higher priority I can fix that.

api_app/views.py
mlodic: I meant that those 4 lines could be a method which returns `users_of_organization`.

I am telling you this because I saw the code repeated multiple times

drosetti (approved):

drosetti:
@shivam-Purohit I think this problem in the dashboard is related to the fact that the regex doesn't allow IPv6, then they are classified as generic and not show in the dashboard charts. Matteo told me you previously changed the regex to accept IPv6, I think I deleted it during a refactor, my apologies. However, I resolved and restored your work on the regex: in the develop branch you can find the refex that accept both IPv4 and IPv6, with a rebase everything should work, let me know if you have some problems. 

I also check your code and approved your pr from a frontend prospective, good work :muscle:

api_app/views.py
mlodic: @shivam-Purohit onceyou get some time, if you could just do this little refactor then we can merge this PR :) thank you

shivam-Purohit:
ubuntuMachine  Running  - Oracle VM VirtualBox 05-01-2024 18_09_39

@drosetti Thanks! I think we should be done here. Do let me know if further changes are required.

api_app/views.py
shivam-Purohit: Sorry i missed the comment above earlier. I think the new refactor should be good.

Issue #1863, opened 2023-08-29
Upgrade Python from 3.9 to 3.11


mlodic:
To verify required changes in the code

g4ze:
hello @mlodic . Is this an issue i can take up? If yes, please guide me to a starting point.

mlodic:
I would prefer to have a maintainer to solve this task cause there is the risk to break things if not done properly

PR #2081, opened 2024-01-19
Python postgres update

### LINKED ISSUE #1862: Upgrade Local Postgres Image from 12 to 16 (or latest)
Aside from changing the pointer to the Docker Image, this change can be introduced only with a new major release of IntelOwl.

In this way we can advise the users to leverage the  official PostgreSQL command to upgrade](https://www.postgresql.org/docs/current/pgupgrade.html) if:
* they are using the local PostgreSQL image
* they want to keep the old DB data intact

New users can just benefit from the new image.

To verify whether there are possible incompatibilties in the code.

0ssigeno:
Closes #1862 and #1863


docs/source/Installation.md
```diff
@@ -294,12 +294,32 @@ Maintainers strive to keep the upgrade between major version easy but it's not a
 Below you can find the additional process required to upgrade from each major versions.
 </div>
 
+### Updating to >v6.0.x from version v5.x.x
+
+You can safely update to >[v6.0.0](https://github.com/intelowlproject/IntelOwl/releases/tag/v6.0.0) from a 5.x.x version of IntelOwl.
+
+
```
mlodic, line 301: I didn't get why this part

docs/source/Installation.md
```diff
@@ -294,12 +294,32 @@ Maintainers strive to keep the upgrade between major version easy but it's not a
...
 #### Updating to >=6.0.0 from a 5.x.x version
 IntelOwl v6 introduced some major changes regarding how the project is started.
 Before upgrading, some important things should be checked by the administrator:
 * Docker Compose V1 support has been dropped project-wide. If you are still using a Compose version prior to v2.3.4, please [upgrade](https://docs.docker.com/compose/migrate/) to a newer version or install Docker Compose V2.
 * IntelOwl is now started with the new Bash `start` script that has the same options as the old Python `start.py` script but is more manageable and has decreased the overall project dependencies. The `start.py` script has now been removed.
 
+The migration procedure is as follows:
+- You have IntelOwl version 5.x.x up and running
+- Bring down the application (you can use the start.py script or manually)
```
mlodic, line 310: now the script is a bash script called `start`

docker/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN npm install npm@latest --location=global \
     && PUBLIC_URL=/static/reactapp/ npm run build
 
 # Stage 2: Backend
-FROM python:3.9.16 as backend-build
+FROM python:3.11 as backend-build
```
mlodic, line 16: it makes sense to explicitly add the patch version for better reproducibility

docs/source/Installation.md
0ssigeno: Forgot to delete it :(

PR #2173, opened 2024-03-06
Websocket

drosetti:
(Please add to the PR name the issue/s that this PR would close if merged by using a [Github](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) keyword. Example: `<feature name>. Closes #999`. If your PR is made by a single commit, please add that clause in the commit too. This is all required to automate the closure of related issues.)

# Description

## Type of change

# Checklist

    - [ ] I have dumped the configuration from Django Admin using the `dumpplugin` command and added it in the project as a data migration. (["How to share a plugin with the community"](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-share-your-plugin-with-the-community))
    - [ ] If you created a new analyzer and it is free (does not require API keys), please add it in the `FREE_TO_USE_ANALYZERS` playbook by following [this guide](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-modify-a-plugin).

### Important Rules


drosetti:
@0ssigeno I discussed with Matte about one thing to change. I'm going to tag you again in another message when i'll have done.

authentication/serializers.py
```diff
@@ -229,3 +230,24 @@ def validate(self, attrs):
...
+            "created",
+        ]
+
+    def create(self, validated_data):
+        user = self.context["request"].user
+        if Token.objects.filter(user=user).exists():
+            raise rfs.ValidationError("An API token was already issued to you.")
+
+        validated_data["user"] = user
+        return super().create(validated_data)
```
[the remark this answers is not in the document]
0ssigeno: `ValidationError` should be raised in the `validate` method.

api_app/websocket/consumer.py
```diff
@@ -0,0 +1,75 @@
...
+                f"send data for the job: {job.id} "
+                f"to the user: {channel.user.username} "
+                f"over the channel: {channel.channel_name}"
+            )
+            job_serializer = WsJobSerializer(job, context={"channel": channel})
+            job_data = job_serializer.data
+            async_to_sync(channel_layer.send)(
+                channel.channel_name,
+                {"type": "send.job", "job": job_data},
+            )
```
0ssigeno: Cant' we use celery for this? Just an idea, don't even know if it is doable

api_app/websocket/models.py
```diff
@@ -0,0 +1,20 @@
...
+
+class JobChannel(models.Model):
+    """Data stored about a job scan.
+
+    * job_id is used to send data to all the channels waiting for the job:
+        multiple users waiting for the same job
+    * user is used to get the permission to kill or delete the job
+    """
+
+    job_id = models.PositiveIntegerField()
```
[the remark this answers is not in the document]
drosetti: collection removed

intel_owl/settings/rest.py
```diff
@@ -12,7 +12,11 @@
 REST_FRAMEWORK = {
     "DEFAULT_RENDERER_CLASSES": ["rest_framework.renderers.JSONRenderer"],
     # Auth
-    "DEFAULT_AUTHENTICATION_CLASSES": ["durin.auth.CachedTokenAuthentication"],
+    "DEFAULT_AUTHENTICATION_CLASSES": [
+        "rest_framework.authentication.TokenAuthentication",
+        "rest_framework.authentication.SessionAuthentication",
+        "rest_framework.authentication.BasicAuthentication",
```
mlodic, line 18: basic auth is required for what?

intel_owl/settings/django.py
```diff
@@ -17,6 +17,10 @@
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "certego_saas.ext.middlewares.LogMiddleware",  # custom
 ]
+
+SESSION_COOKIE_HTTPONLY = True
+SESSION_COOKIE_AGE = 60 * 60 * 24 * 90  # seconds * minutes * hours * days
```
mlodic, line 22: 90 days for keeping the session alive? this means that, once authenticated, the session cookie is valid for 90 days and only after that time the user needs to login again? That is very bad on a security perspective. I'd like to keep this value customizable. In this way the users can decide whether to invest more in security rather than usability. It's a matter of risk analysis.
Then, I think that the default could be kept 2 weeks as the Django default. Or there are any other reasons that I am missing?

intel_owl/settings/celery.py
```diff
@@ -8,9 +8,12 @@
...
 RESULT_BACKEND = "django-db"
-BROKER_URL = get_secret(
-    "BROKER_URL", "sqs://" if AWS_SQS else "amqp://guest:guest@rabbitmq:5672"
-)
+BROKER_URL = get_secret("BROKER_URL", None)
+if not BROKER_URL and AWS_SQS:
+    BROKER_URL = "sqs://"
+else:
+    BROKER_URL = "redis://redis:6379/1"  # 0 is used by channels
+
```
mlodic, line 16:
```suggestion
if not BROKER_URL:
    if AWS_SQS:
        BROKER_URL = "sqs://"
    else:
        BROKER_URL = "redis://redis:6379/1"  # 0 is used by channels

```
I think you wanted to do that instead

intel_owl/settings/cache.py
```diff
@@ -35,6 +35,8 @@ def get_where(self, starts_with: str, version=None) -> Dict[str, Any]:
         return self.get_many([row[0] for row in rows], version=version)
 
 
+DJANGO_REDIS_IGNORE_EXCEPTIONS = True
```
mlodic, line 38: reading the [doc](https://github.com/jazzband/django-redis?tab=readme-ov-file#memcached-exceptions-behavior) about this, I am not sure that's a good idea because if redis is down everything goes down. We can't ignore exceptions

docker/default.yml
```diff
@@ -8,6 +8,7 @@ services:
   uwsgi:
     image: intelowlproject/intelowl:${REACT_APP_INTELOWL_VERSION}
     container_name: intelowl_uwsgi
+    hostname: uwsgi
```
mlodic, line 11: is this really needed?

docker/postgres.override.yml
```diff
@@ -5,6 +5,7 @@ services:
   postgres:
     image: library/postgres:16-alpine
     container_name: intelowl_postgres
+    hostname: postgres
```
mlodic, line 8: why you need to rename them all?

docs/source/Advanced-Configuration.md
```diff
@@ -174,15 +174,41 @@ First, you need to configure the environment variable `LOCAL_STORAGE` to `False`
...
 
-#### SQS
+#### Message Broker
 
-If you like, you could use AWS SQS instead of Rabbit-MQ to manage your queues.
+IntelOwl at the moment supports 3 different message brokers:
+- RabbitMQ
+- Aws SQS
+- Redis
+
```
mlodic, line 183: considering that the websockets do not work if we don't have redis, can we really say that now?
I think there are 2 options:
* easiest one is to drop support for other brokers and release this change with the major 
* find a way to have the websocket work without redis but I already feel the pain of managing these 2 cases.

mlodic (changes requested):

docker/postgres.override.yml
0ssigeno: We had some with the name, some without. Just to have the same syntax everywhere

docker/default.yml
0ssigeno: We had some with the name, some without. Just to have the same syntax everywhere

docs/source/Advanced-Configuration.md
0ssigeno: Yes we can. 
There is nothing that prevents an user to use SQS and have redis for the websockets. Same thing with rabbitmq

intel_owl/settings/cache.py
0ssigeno: Copy pasted from @sevdog, we can clearly change it

docs/source/Advanced-Configuration.md
mlodic: yes but in which case make sense to use another layer in the architecture when redis solve all of the problems at once? I'd remove the support for others, honestly

intel_owl/settings/django.py
drosetti: I agree with 2 weeks. I think it's possible to make it customizable, but could required some works: first of all this is a django settings, so we need to find a way to go around it. I'm not sure is possible and if it's possible do it without a strange workaround. Also we will need a UI or something like that (we need to ask to the user the time frame he likes).

intel_owl/settings/celery.py
drosetti: I think @0ssigeno  did this

api_app/websocket/consumer.py
drosetti: Honestly I don't know, I did some test, but it not seems to work. Also this function is called from the celery tasks so it's async.

docs/source/Advanced-Configuration.md
0ssigeno: I don't get why you want to remove the support since we have and it costs pretty much 0 to keep it. 
But again ,we can remove support (and add the breaking change in the next release)

docs/source/Advanced-Configuration.md
```diff
@@ -174,15 +174,41 @@ First, you need to configure the environment variable `LOCAL_STORAGE` to `False`
...
 
 
-#### SQS
+#### Message Broker
 
-If you like, you could use AWS SQS instead of Rabbit-MQ to manage your queues.
+IntelOwl at the moment supports 3 different message brokers:
+- RabbitMQ
+- Aws SQS
+- Redis
```
mlodic, line 182:
```suggestion
- Redis (default)
- RabbitMQ
- Aws SQS
```

I know that it is written below but that's to be more more clear at the very start

docs/source/Advanced-Configuration.md
mlodic: I checked again how the doc is written and the default configuration implemented cause I would like to do not create confusion on the user and yes, ok, I agree that by changing the doc a little we can maintain the support

intel_owl/settings/cache.py
mlodic: @drosetti after this check I think we can merge the PR

drosetti: @mlodic removed

api_app/websocket.py
```diff
@@ -0,0 +1,116 @@
...
+        @cached_property
+        def group_list(self) -> List[str]:
+            return [self.job_group_name, self.job_group_perm_name]
+
+        def get_group_for_user(self, user: User) -> str:
+            return (
+                self.job_group_perm_name
+                if (
+                    self._job.user == user
+                    or self._job.user.membership.organization.user_has_membership(user)
```
0ssigeno, line 42: If the user has no membership this will raise an exception

0ssigeno (changes requested):

PR #2209, opened 2024-03-19
Tweet feedsfixes#1770

### LINKED ISSUE #1770: [Analyzer] TweetFeeds
## Name
TweetFeeds

## Link
https://tweetfeed.live

## Type of analyzer
observables

## Why should we use it
Free from the community

## Possible implementation
https://tweetfeed.live/api.html Extract their IOCs periodically in a text file and perform a simple match there.

g4ze:
closes #1770

# Description

add tweet feeds
## Type of change

# Checklist

    - [ ] I have dumped the configuration from Django Admin using the `dumpplugin` command and added it in the project as a data migration. (["How to share a plugin with the community"](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-share-your-plugin-with-the-community))
    - [ ] If you created a new analyzer and it is free (does not require API keys), please add it in the `FREE_TO_USE_ANALYZERS` playbook by following [this guide](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-modify-a-plugin).

### Important Rules
Screenshot from 2024-03-20 01-20-24
{
        "date": "2024-03-19 00:31:36",
        "user": "Metemcyber",
        "type": "url",
        "value": "http://210.56.49.214",
        "tags": [
            "#phishing"
        ],
        "tweet": "https://twitter.com/Metemcyber/status/1769884392477077774"
    },

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
+++ b/api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
@@ -15,25 +15,21 @@ logger = logging.getLogger(__name__)
 
 class TweetFeeds(ObservableAnalyzer):
     """
-    wrapper for https://tweetfeed.live
+    Wrapper for https://tweetfeed.live api
     """
 
     url = "https://api.tweetfeed.live/v1/"
     filter1: str = ""
-    update_on_run: bool = True
     time: str
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.db_dir = os.path.join(settings.MEDIA_ROOT, "TweetFeedsDBs")
-        self.create_db_dir()
+    @classmethod
+    def location(cls) -> Tuple[str, str]:
+        db_name = "tweetfeed_month.json"
+        url = "https://api.tweetfeed.live/v1/month"
+        return f"{settings.MEDIA_ROOT}/{db_name}", url
 
-    def create_db_dir(self):
-        if not os.path.exists(self.db_dir):
-            os.makedirs(self.db_dir)
-
-    def location(self) -> Tuple[str, str]:
-        if self.filter1 != "":
+    def run_url(self) -> str:
+        if self.filter1:
             url = (
                 self.url
                 + self.time
@@ -44,32 +40,57 @@ class TweetFeeds(ObservableAnalyzer):
             )
         else:
             url = self.url + self.time + "/" + self.observable_classification
-
-        db_name = f"""
-        tweetfeed_{self.time}_{self.filter1}_{self.observable_classification}.json
-        """
-
-        return f"{settings.MEDIA_ROOT}/TweetFeedsDBs/{db_name}", url
+        return url
 
     def run(self):
-        db_location, url = self.location()
-        logger.info(f"Running TweetFeeds {url} at {db_location}")
-        if self.update_on_run or not os.path.exists(db_location):
-            if not self.update():
-                raise AnalyzerRunException("Unable to update database")
+        # update logic for first time run
+        default_db, default_url = self.location()
+        if not os.path.exists(default_db) and not self.update():
+            raise AnalyzerRunException(
+                f"Could not find or update db at {default_db} using {default_url}"
+            )
 
-        with open(db_location, "r", encoding="utf-8") as f:
-            try:
-                db = json.load(f)
-            except json.JSONDecodeError as e:
-                raise AnalyzerRunException(f"Decode JSON in run: {e}")
+        with open(default_db, "r", encoding="utf-8") as f:
+            logger.info(f"TweetFeeds running with {default_db}")
+            db = json.load(f)
+            for tweet in db:
+                if tweet["value"] == self.observable_name:
+                    if self.filter1 and (
+                        self.filter1 in tweet["tags"] or self.filter1 == tweet["user"]
+                    ):
+                        # this checks if our user has demanded for a
+                        # specific filter and return data based on the
+                        # filter in default db
+                        return tweet
+                    elif not self.filter1:
+                        return tweet
+
+        if self.time == "year":
+            # we already have the updated data for the month
+            # (covers week and today options) with us;
+            # year is the only extended version possible
+            run_url = self.run_url()
+            logger.info(f"TweetFeeds extending using {run_url}")
+
+            # simply make api call and search for observable
+            response = requests.get(run_url)
+            response.raise_for_status()
+            db = response.json()
             for tweet in db:
                 if tweet["value"] == self.observable_name:
                     return tweet
+
         return {"found": False}
 
-    def update(self) -> bool:
-        db_location, db_url = self.location()
+    @classmethod
+    def update(cls) -> bool:
+        """
+        Update TweetFeeds database:
+        Our default DB gets data with
+        no filter for the past month
+        """
+
+        db_location, db_url = cls.location()
         logger.info(f"Updating TweetFeeds {db_url} at {db_location}")
 
         try:
--- a/api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
+++ b/api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
@@ -1,0 +1,121 @@
+import json
+import logging
+import os
+from typing import Tuple
+
+import requests
+from django.conf import settings
+
+from api_app.analyzers_manager.classes import ObservableAnalyzer
+from api_app.analyzers_manager.exceptions import AnalyzerRunException
+from tests.mock_utils import MockUpResponse, if_mock_connections, patch
+
+logger = logging.getLogger(__name__)
+
+
+class TweetFeeds(ObservableAnalyzer):
+    """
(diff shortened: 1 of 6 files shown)


g4ze:
Hey! @mlodic does this implementation make sense? We can have health checks and update schedule hourly?

api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
```diff
@@ -0,0 +1,122 @@
...
+        super().__init__(*args, **kwargs)
+        self.db_dir = os.path.join(settings.MEDIA_ROOT, "TweetFeedsDBs")
+        self.create_db_dir()
+
+    def create_db_dir(self):
+        if not os.path.exists(self.db_dir):
+            os.makedirs(self.db_dir)
+
+    def location(self) -> Tuple[str, str]:
+        if self.filter1 != "":
```
mlodic, line 36: isn't enough to check whether this is populated or not instead of this comparison?

api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
```diff
@@ -0,0 +1,122 @@
...
+                + self.filter1
+                + "/"
+                + self.observable_classification
+            )
+        else:
+            url = self.url + self.time + "/" + self.observable_classification
+
+        db_name = f"""
+        tweetfeed_{self.time}_{self.filter1}_{self.observable_classification}.json
+        """
```
mlodic, line 50: please use normal strings with just one quote

mlodic (commented):
yep, health check and update could be done hourly

api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
```diff
@@ -0,0 +1,122 @@
...
+
+        return f"{settings.MEDIA_ROOT}/TweetFeedsDBs/{db_name}", url
+
+    def run(self):
+        db_location, url = self.location()
+        logger.info(f"Running TweetFeeds {url} at {db_location}")
+        if (
+            self.update_on_run or not os.path.exists(db_location)
+        ) and not self.update():
```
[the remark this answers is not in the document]
g4ze: won't we any how need to have `db_location` to check `os.path.exists(db_location)` and for that we would have to call location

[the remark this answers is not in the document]
g4ze: `os.path.exists(db_location)` che5ck is needed in case the analyzer is running for the first time for the url and `update on run` is set to false

api_app/analyzers_manager/migrations/0071_analyzer_config_tweetfeed.py
```diff
@@ -0,0 +1,186 @@
...
+        "required": True,
+    },
+    {
+        "python_module": {
+            "module": "tweetfeeds.TweetFeeds",
+            "base_path": "api_app.analyzers_manager.observable_analyzers",
+        },
+        "name": "filter1",
+        "type": "str",
+        "description": "filter1 for TweetFeed",
```
mlodic, line 63: reference to available options can help the user

api_app/analyzers_manager/migrations/0071_analyzer_config_tweetfeed.py
```diff
@@ -0,0 +1,186 @@
...
+
+params = [
+    {
+        "python_module": {
+            "module": "tweetfeeds.TweetFeeds",
+            "base_path": "api_app.analyzers_manager.observable_analyzers",
+        },
+        "name": "time",
+        "type": "str",
+        "description": "time for TweetFeed",
```
mlodic, line 52: hour, minutes, ???

api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
```diff
@@ -0,0 +1,123 @@
...
+class TweetFeeds(ObservableAnalyzer):
+    """
+    wrapper for https://tweetfeed.live
+    """
+
+    url = "https://api.tweetfeed.live/v1/"
+    filter1: str = ""
+    update_on_run: bool = True
+    time: str = "test_time"  # for tests
+    observable_classification: str = "test"  # for tests
```
mlodic, line 25: mmmm test things should remain in tests, cannot be the default value of these attributes.

Considering that the `update` works hourly by default, we need to put here decent values and we need also to add those as default parameters in the migration.

`observable_classification` should not exist while `time` should be a default, Idk, a week, a month?

api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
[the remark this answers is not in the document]
mlodic: I was trying to suggest to avoid to call 2 times `        db_location, db_url = self.location()` but that does not really matter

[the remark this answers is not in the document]
g4ze: Umm, I'm sorry but I don't quite understand. Location is being called two times which is unavoidable in my understanding. It's required to be called once to check for pre existence of files, and once inside update.
Can you please elaborate a bit more?

api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
g4ze: Okay so I have some doubts here, Example if I did a run with some parameters that generate a unique url to download, I would want my update function to update the db based on that generated url every hour or so. Here the update function wrt the framework in place requires me to hardcode the url for update which seems unusable/irrelevant in this use case, where the url can be unique/new/user based

api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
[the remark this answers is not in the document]
mlodic: yeah np, just ignore

api_app/analyzers_manager/observable_analyzers/tweetfeeds.py
g4ze: Hmm, okay, so you are saying to have one generic database url hardcoded. If the user wants they can extend the default database by using update on run parameter and providing other parameters at the same time. We would require other url parameters only and only if the update on run is true. Else in every run, we can simply look up the default db.
It does make sense and simplifies things a lot and can be implemented.

However, in the longer run, I do believe that the update function should have some mechanism to be able to cache or sustain the values of the parameters used in the last run of such analyzers, this could save us extra computation by downloading and updating the exact database required by the user instead of relying on having a larger superset of the the database that the user **might** use.
Example:
In our case as you suggest, we are downloading data for all the supported observables types for a month and updating it every hour. If I'm a user who is only looking for the data from a particular user with time set to just `today` with classification type `url` why would I want my software to download, update and search for data of past 30 days that has entries of all other observables including `url`.

mlodic: I understand but consider that the problem is a lot bigger than what you think because if you need to manage the data by the user you should also start to leverage and check all things related to permissions, organizations and access and things can become quite messy very fast.

You would need to create a different cache for every user based on each user custom parameters and manage application and file access.

Again, I'd stick for the simple one. It benefits maintenance and stability.

> In our case as you suggest, we are downloading data for all the supported observables types for a month and updating it every hour.

not every hour, it is too much, we can do that once in a day.

> I would want to cache and update only for the database that my generated url gives and not care about a superset of my use case db

yeah you could add another boolean option to manage the case where the user does use only the data downloaded at the time of request but again, simple is better. Consider that the normal user does not know all the details that we are talking about. He needs to have something simple to understand and that works. That's all

g4ze: Understood! Let's move ahead with what you proposed. I'm travelling right now, so I will probably commit the updates tomorrow.

mlodic: np man, take your time :)

g4ze:
@mlodic please tell me if this change looks fine. 
I'm also thinking that we can remove the `update_on_run`  completely. The new flow can be:

- look up for observable in default_db
  - check and **return** tweet if found based on the `filter1` if `filter1` exists [I'm doing this as per the latest commit]
  - if there's no `filter1` then simply **return** the tweet if found in `default_db` [I'm doing this as per the latest commit]
- check if `time` is entered and is `year` [`month covers `week` and `today` ; only supported filter for time left for api is `year`]
  - return tweet based on the extended search of past year with applied filters if any

This could make things even more simple for the user. Instead of dealing with another update variable, they can simply provide whatever parameters they want and we can internally apply logic based on our default db and decide if we want to make an api call or not.

mlodic:
yeah why not :)

mlodic:
please pull from `develop`, the problem should not be related to your pr

g4ze:
@mlodic I think we are gtg, kindly review :)

mlodic (approved):

PR #2234, opened 2024-03-29
Malcore file analyzer #2168

### LINKED ISSUE #2168: [Analyzer] Malcore
## Name
Malcore

## Link
https://malcore.io

## Type of analyzer
file analyzer, similar to the other malware sandbox services already implemented in intelowl

## Why should we use it
free api quota: https://malcore.io/pricing

## Possible implementation
https://support.malcore.io/hc/en-au/articles/17711707070617-Using-the-Malcore-API
https://github.com/Internet-2-0/Malcore-SDK

skkra0:
Closes #2168 

# Description

Created Malcore file analyzer using the executable analysis endpoint

## Type of change

# Checklist

    - [x] I have dumped the configuration from Django Admin using the `dumpplugin` command and added it in the project as a data migration. (["How to share a plugin with the community"](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-share-your-plugin-with-the-community))
    - [x] If a File analyzer was added and it supports a mimetype which is not already supported, you added a sample of that type inside the archive `test_files.zip` and you added the default tests for that mimetype in [test_classes.py](https://github.com/intelowlproject/IntelOwl/blob/master/tests/analyzers_manager/test_classes.py).
    - [x] If you created a new analyzer and it is free (does not require API keys), please add it in the `FREE_TO_USE_ANALYZERS` playbook by following [this guide](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-modify-a-plugin).
    - [x] Check if it could make sense to add that analyzer/connector to other [freely available playbooks](https://intelowl.readthedocs.io/en/develop/Usage.html#list-of-pre-built-playbooks).
- [x] If changes were made to an existing model/serializer/view, the docs were updated and regenerated (check [CONTRIBUTE.md](https://github.com/intelowlproject/IntelOwl/blob/master/docs/source/Contribute.md)).

[report.json](https://github.com/intelowlproject/IntelOwl/files/14891840/report.json)


skkra0:
Hey @mlodic, what tests would you recommend I add for this analyzer?
Also, the Malcore API has another 21 (free to use) endpoints that take files, should I include those as well?

api_app/analyzers_manager/file_analyzers/malcore_scan.py
```diff
@@ -0,0 +1,60 @@
...
+        files = {"filename1": binary}
+
+        try:
+            response = requests.post(self.url, headers=headers, files=files)
+            response.raise_for_status()
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
+
+        result = response.json()
+        return result
```
mlodic, line 28: great but actually, as specified in the issue, this should be implemented like other similar sandbox services integrations we have in IntelOwl.
We expect this analyzer to:
* first check whether the file md5/sha256 has already been analyzed recently. In that case, just retrieve the report of the already existing analysis
* if that file is new, upload the file to the service like you just did here
* then, poll an api that checks the status of the analysis, until the analysis is ended
* once the analysis is ended, retrieve the final result

Please check other similar analyzers like [triage](https://github.com/intelowlproject/IntelOwl/blob/master/api_app/analyzers_manager/file_analyzers/triage_scan.py)

api_app/analyzers_manager/migrations/0075_analyzer_config_malcore_scan.py
```diff
@@ -0,0 +1,121 @@
...
+    "python_module": {
+        "health_check_schedule": None,
+        "update_schedule": None,
+        "module": "malcore_scan.MalcoreScan",
+        "base_path": "api_app.analyzers_manager.file_analyzers",
+    },
+    "name": "Malcore_Scan",
+    "description": "Malcore Scan",
+    "disabled": False,
+    "soft_time_limit": 60,
```
mlodic, line 18: this should be higher and should be higher than the maximum polling time

api_app/analyzers_manager/migrations/0075_analyzer_config_malcore_scan.py
```diff
@@ -0,0 +1,121 @@
...
+    "health_check_status": True,
+    "type": "file",
+    "docker_based": False,
+    "maximum_tlp": "CLEAR",
+    "observable_supported": [],
+    "supported_filetypes": [
+        "application/x-sharedlib",
+        "application/vnd.microsoft.portable-executable",
+        "application/x-elf",
+    ],
```
mlodic, line 29: aren't they supporting more than these type of files?

mlodic (changes requested):

mlodic:
> Hey @mlodic, what tests would you recommend I add for this analyzer?
Also, the Malcore API has another 21 (free to use) endpoints that take files, should I include those as well?

it's enough to populate the monkeypatch like you did.
No need for all the other endpoints. Just implement the ones required for the flow that I have explained in the review

skkra0:
Hey @mlodic,
I rewrote the analyzer, added more file types to the migration file, and raised the time limit.
Can you clarify what you mean by checking whether the file has already been analyzed recently? It seems like IntelOwl already prevents users from repeating scans on a file. I tried renaming a file and scanning it again, but the app still took me back to the previous scan.

api_app/analyzers_manager/file_analyzers/malcore_scan.py
```diff
@@ -0,0 +1,333 @@
...
+        binary = self.read_file_bytes()
+        files = {"filename1": binary}
+        try:
+            logger.info(f"Sending {self.md5} to Malcore")
+            response = requests.post(
+                self.base_url + "upload", headers=self.headers, files=files
+            )
+            response.raise_for_status()
+        except requests.RequestException as e:
+            raise AnalyzerRunException(e)
```
mlodic, line 33: if you raise an `AnalyzerRunException` without changing the log name, that would be already handled automatically by the framework. Uncaught errors always trigger AnalyzerRunException

api_app/analyzers_manager/file_analyzers/malcore_scan.py
```diff
@@ -0,0 +1,333 @@
...
+        except TypeError:
+            raise AnalyzerRunException(f"Failed analysis for {self.md5}")
+
+        for _try in range(self.max_tries):
+            logger.info(f"polling malcore try #{_try + 1}")
+            self.result = self._get_status(uuid)["data"]
+            if "status" not in self.result and "msg" not in self.result:
+                break
+
+            time.sleep(self.poll_distance)
```
mlodic, line 46: max_tries and poll_distance must be defined both as a class attribute and in the migration file as params.

api_app/analyzers_manager/file_analyzers/malcore_scan.py
```diff
@@ -0,0 +1,333 @@
...
+        try:
+            uuid = response.json()["data"]["data"]["uuid"]
+        except TypeError:
+            raise AnalyzerRunException(f"Failed analysis for {self.md5}")
+
+        for _try in range(self.max_tries):
+            logger.info(f"polling malcore try #{_try + 1}")
+            self.result = self._get_status(uuid)["data"]
+            if "status" not in self.result and "msg" not in self.result:
+                break
```
mlodic, line 44: is this an error? In case, we need to log something about

mlodic: we also needs to log when the final result has been correctly retrieved

mlodic (changes requested):

api_app/analyzers_manager/migrations/0075_analyzer_config_malcore_scan.py
```diff
@@ -0,0 +1,169 @@
...
+    "docker_based": False,
+    "maximum_tlp": "CLEAR",
+    "observable_supported": [],
+    "supported_filetypes": [
+        "application/x-sharedlib",
+        "application/vnd.microsoft.portable-executable",
+        "application/x-elf",
+        "application/octet-stream",
+        "application/x-binary",
+    ],
```
mlodic, line 31: I would just remove this and keep it empty.

mlodic:
> Can you clarify what you mean by checking whether the file has already been analyzed recently? It seems like IntelOwl already prevents users from repeating scans on a file. I tried renaming a file and scanning it again, but the app still took me back to the previous scan

Yep, the default is that the IntelOwl analysis won't be repeated if done in the last 24 hours. see:
image

This cache can be overwritten or forced. In that case, you would re-analyze the file again.

However here my point is different.
In case the external service, which is Malcore in this case, already has a public analysis available of that sample, we should avoid to force a new analysis on the same service. So I would check whether the md5 of the file that we upload has already been analyzed in that service. If yes, just retrieve the result of the already available analysis on the external service.
(as already said, in other similar analyzers the same logic has been already implemented...)

skkra0:
> So I would check whether the md5 of the file that we upload has already been analyzed in that service. If yes, just retrieve the result of the already available analysis on the external service.

Hello, sorry for being away for a while. 
Checking if the uploaded file has already been analyzed is a premium feature. Should I still implement it?

mlodic:
> Hello, sorry for being away for a while.
Checking if the uploaded file has already been analyzed is a premium feature. Should I still implement it?

no need so

skkra0:
> no need so
no need to check for a public analysis?

mlodic:
closing cause not working as expected

PR #2395, opened 2024-06-22
docker-compose not docker compose

vncloudsco:
Command typo error does not contain spaces


ashcorpdev:
This is not a typo - `docker-compose` refers to v1 of Compose. The correct (and supported) syntax for v2 is `docker compose`.

See [here](https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2).

PR #2401, opened 2024-06-27
Goresym analyzer, fixes#1451 and fixes executable file support

### LINKED ISSUE #1451: [Analyzer] GoReSym
## Name
GoReSym

## Link
https://github.com/mandiant/GoReSym

## Type of analyzer
docker analyzer, to insert in the malware_analysis_tools image

## Why should we use it
This allows to extract useful info when analyzing go binaries. This should be executed only after having detected that the file is go-compiled

## Possible implementation
Follow the Usage tips in the official repo: https://github.com/mandiant/GoReSym

g4ze:
closes #1451

# Description

## Type of change

# Checklist

    - [x] I have dumped the configuration from Django Admin using the `dumpplugin` command and added it in the project as a data migration. (["How to share a plugin with the community"](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-share-your-plugin-with-the-community))
    - [ ] If you created a new analyzer and it is free (does not require any API key), please add it in the `FREE_TO_USE_ANALYZERS` playbook by following [this guide](https://intelowl.readthedocs.io/en/latest/Contribute.html#how-to-modify-a-plugin).
    - [ ] If the plugin interacts with an external service, I have created an attribute called precisely `url` that contains this information. This is required for Health Checks. 

### Important Rules


g4ze:
The analyzer works fine when i configure it to support all files. when i configure it to only support binaries and exe , it says incompatible mimetype. Moreover, the FE shows that the file type is `application/x-executable` , which doesn't seem to be supported anywhere in models.

g4ze:
@mlodic  any suggestions on what files should the analyzer support?

api_app/analyzers_manager/file_analyzers/goresym.py
```diff
@@ -0,0 +1,33 @@
...
+    # http request polling max number of tries
+    max_tries: int = 5
+
+    def update(self) -> bool:
+        pass
+
+    def run(self):
+        binary = self.read_file_bytes()
+        fname = str(self.filename).replace("/", "_").replace(" ", "_")
+        args = ["-t", "-d", "-p", f"@{fname}"]
```
mlodic, line 23: i would add parameters for all the options, these 3 ones but also the other available ones:
```
-m <virtual address> ("manual", optional) flag will dump the RTYPE structure recursively at the given virtual address
-v <version string> ("version", optional) flag will override automated version detection and use the provided version. This is needed for some stripped binaries. Type parsing will fail if the version is not accurate.
```

g4ze:
[main.zip](https://github.com/user-attachments/files/16028076/main.zip)
the file i used is zipped here...

api_app/analyzers_manager/file_analyzers/goresym.py
g4ze: you mean I should add all the available flags as parameter options in the analyzer settings?

mlodic: yep, that allows the user to customize the execution of the tool like he was running the tool standalone

mlodic:
ok I tried that tool and I think you should use `application/vnd.microsoft.portable-executable` for that. Can you try with that and lmk?

g4ze:
I used application/vnd.microsoft.portable-executable only initially but frontend said incompatible mimetype. Does it work done in your local with accepted mime type as application/vnd.microsoft.portable-executable?
If yes then I'll push a commit with that setting only.

g4ze:
here's a recording of what is exactly happening, in the first run as it can be seen, analyzer is set to accept all files. In the second run it is set to `exe` only. Moreover, after the first run, file type is shown as `application/x-executable`.
[2024-06-28 21-18-10.zip](https://github.com/user-attachments/files/16033113/2024-06-28.21-18-10.zip)

mlodic:
I think you found a bug. When we did [this change](https://github.com/intelowlproject/IntelOwl/blob/master/api_app/analyzers_manager/migrations/0063_alter_analyzerconfig_not_supported_filetypes_and_more.py), we wanted to use the correct mimetype value cause the `x-` based one are not definitive and the new python library from python 3.11 started to use `application/vnd.microsoft.portable-executable` instead of `application/x-dosexec`. But it seems it is not the same for ELF executables and we need to revert this as soon as possible for ELF files, otherwise they are not supported.

Would you like to try to help us in solving this too?
If you check the old migrations, you can find all the analyzers where the `x-executable` has been removed. We should re-add it.
Plus, we should re-add it as supported mimetype (there are 2 sections in the code where this is done, 1 backend and 1 frontend).
Then, you can use this mimetype for this analyzer too.

g4ze:
sure, i'll take a look. I did have a look at it before implementing the analyzer, when i was thinking for a way to check if the file was go-compiled as soon as we upload. It seems like the tool we are using is `from magic import from_buffer as magic_from_buffer` and it stills gives `application/x-executable` as an output for the go binary.
Here's a code snippet whcich lets us have an insight on how the tool reacts:
'''
from magic import from_buffer as magic_from_buffer

files = ['/home/nilay/Documents/GitHub/byoc/main', "/home/nilay/Downloads/file.exe", "/home/nilay/Downloads/ping.elf"]

for file in files:
    try:
        with open(file, 'rb') as f:
            file_type = magic_from_buffer(f.read(2048), mime=True)
        print(f"{file}: {file_type}")
    except IOError:
        print(f"{file}: File not found or unreadable")
'''
o/p:
'''
/home/nilay/Documents/GitHub/byoc/main: application/x-executable
/home/nilay/Downloads/file.exe: application/vnd.microsoft.portable-executable
/home/nilay/Downloads/ping.elf: application/x-sharedlib
'''
We have made implementations for elf and exe but not anything like a go-compiled binary.

g4ze:
>Plus, we should re-add it as supported mimetype (there are 2 sections in the code where this is done, 1 backend and 1 frontend).
you mean somethings like the recent [commit](https://github.com/intelowlproject/IntelOwl/pull/2401/commits/1983b606abeaa8b61b7b38d4d7b5865673fa880c) i made.

g4ze:
This works as expected on my machine. 
[x] changed the migration [("application/vnd.tcpdump.pcap", "Pcap"),](https://github.com/intelowlproject/IntelOwl/blob/27d89b36d6fbb5dfae5be17b4655716f06af0d35/api_app/analyzers_manager/migrations/0063_alter_analyzerconfig_not_supported_filetypes_and_more.py) 
 - here I removed the part where `x-executable` was being filtered and removed 
 - appended new file type X-Exe in choices for supported and non supported files
 [x] appended the file type in another [("application/vnd.tcpdump.pcap", "Pcap"),](https://github.com/intelowlproject/IntelOwl/blob/27d89b36d6fbb5dfae5be17b4655716f06af0d35/api_app/analyzers_manager/migrations/0063_alter_analyzerconfig_not_supported_filetypes_and_more.py) migration involving file choices
 [x]updated model and job const [fe and be updates for reflection of file selctions]

g4ze:
>This works as expected on my machine.

 changed the [migration](https://github.com/intelowlproject/IntelOwl/blob/27d89b36d6fbb5dfae5be17b4655716f06af0d35/api_app/analyzers_manager/migrations/0063_alter_analyzerconfig_not_supported_filetypes_and_more.py)
here I removed the part where x-executable was being filtered and removed
appended new file type X-Exe in choices for supported and non supported files
 appended the file type in another [migration](https://github.com/intelowlproject/IntelOwl/blob/27d89b36d6fbb5dfae5be17b4655716f06af0d35/api_app/analyzers_manager/migrations/0084_alter_analyzerconfig_not_supported_filetypes_and_more.py) involving file choices
 updated model and job const [fe and be updates for reflection of file selctions]
 
 
 i just realised that this fix wont help the current users of intelowl who have already migrated their databases in the recent updates.
in the migration where we have removed x-executable, we have replaced two mimetypes with a single one.
`application/x-dosexec` and `application/x-executable` are replaced with newer `application/vnd.microsoft.portable-executable` now the problem is that we want to bring back `application/x-executable` as a supported type and add it to the old analyzers which supported it. The choke point is that we need to decide which all were the analyzers which supported `application/x-executable` as one of their mimetypes. Currently, if an analyzer supports `application/vnd.microsoft.portable-executable` , it is hard to decipher what exact mimetype it used to support previously in our two possible choices.
In my understanding, it leaves us two options:
- Filter all the analyzers that support `application/vnd.microsoft.portable-executable` and blindly add `application/x-executable` as one of the supported mimetype.
OR
- Migrate back, run a query to find all the analyzers that supported `application/x-executable` in that time, use the resultant list to migrate and update all the specific analyzers.

Do you have any suggestions or add-ons for this? Please correct me if I'm wrong with how I understood the problem.

g4ze:
I implemented the second option for the time being, here's what the quries returned:
```
intel_owl_db=# SELECT name
intel_owl_db-# FROM analyzers_manager_analyzerconfig
intel_owl_db-# WHERE 'application/x-executable' = ANY(supported_filetypes);
     name      
---------------
 Capa_Info
 ELF_Info
 Malpedia_Scan
 Intezer_Scan
 Qiling_Linux
(5 rows)
```
```
intel_owl_db=# SELECT name
FROM analyzers_manager_analyzerconfig
WHERE 'application/x-executable' = ANY(not_supported_filetypes);
 name 
------
(0 rows)
```
Migrated the database based on the above o/p

mlodic:
yep you got everything right. and yes, those are the analyzers that were using that mimetype. You could also have found them just by looking for the string ` application/x-executable` in the old migrations

api_app/analyzers_manager/file_analyzers/goresym.py
```diff
@@ -0,0 +1,37 @@
...
+        args.append(f"@{fname}")
+        req_data = {"args": args}
+        req_files = {fname: binary}
+        logger.info(
+            f"Running {self.analyzer_name} on {self.filename} with args: {args}"
+        )
+        result = self._docker_run(req_data, req_files, analyzer_name=self.analyzer_name)
+        if "error" in result:
+            raise AnalyzerRunException(result["error"])
+
```
mlodic, line 36: can you provide a sample result with a file that is not go-compiled?

api_app/analyzers_manager/migrations/0101_analyzer_config_goresym.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class Migration(migrations.Migration):
+    atomic = False
+    dependencies = [
+        ("api_app", "0062_alter_parameter_python_module"),
+        ("analyzers_manager", "0100_add_x_executable"),
+    ]
+
+    operations = [migrations.RunPython(migrate, reverse_migrate)]
```
mlodic, line 152: if the execution time is fast enough (can you make some tests?) we could add this in the default playbook of Static File Analyzers

mlodic (changes requested):

api_app/analyzers_manager/file_analyzers/goresym.py
```diff
@@ -0,0 +1,37 @@
...
+
+class GoReSym(FileAnalyzer, DockerBasedAnalyzer):
+    name: str = "executable_analyzer"
+    url: str = "http://malware_tools_analyzers:4002/goresym"
+    # interval between http request polling
+    poll_distance: int = 5
+    # http request polling max number of tries
+    max_tries: int = 5
+    flags: str = "-t -d -p"
```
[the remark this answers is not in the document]
g4ze: yes, makes sense

api_app/analyzers_manager/file_analyzers/goresym.py
g4ze:
```
GoReSym /home/nilay/Downloads/file.exe
{"error": "Failed to parse file: failed to read pclntab: failed to locate pclntab"}
```

g4ze: {'error': 'Failed to parse file: failed to read pclntab: failed to locate pclntab'}
for 
[file.zip](https://github.com/user-attachments/files/16076748/file.zip)

api_app/analyzers_manager/migrations/0101_analyzer_config_goresym.py
g4ze: done

g4ze:
the test fails as `application/vnd.microsoft.portable-executable` isn't necessarily compatible with the analyzer and our test file is surely not.

api_app/analyzers_manager/migrations/0101_analyzer_config_goresym.py
g4ze: and yes, execution is pretty fast.

mlodic:
ah ok so I think we should modify the analyzer to avoid that the error that you mentioned `{'error': 'Failed to parse file: failed to read pclntab: failed to locate pclntab'}` does make the analyzer fail. The result would be "success" and the report would be "this is not a Go-compiled file. Full error <error>". And we could add a warning log about that.

For all of the other errors triggered by the tool, it would fail the analyzer normally.
In that way, the analyzer should never fail for any portable executables.
Actually there's no better way to handle this and to avoid that this analyzer fails everytime a PE is sent to the platform, which is very common use case

g4ze:
image

Issue #2554, opened 2024-10-25
Celery not working in develop branch


arakakiv:
## What happened
The celery components aren't working as expected; no job can be successfully executed by the GUI.

## Environment
Latest changes in the `develop` branch.

##  What did you expect to happen
No error being thrown just like the `prod` environment.

## How to reproduce your issue
Just run the start script in `test` mode.

## Error messages and logs
For both `celery_beat` and `celery_default`:
```
Error: Invalid value for '-A' / '--app': 
Unable to load celery application.
```

mlodic:
🤔  you right let me check

mlodic:
no nevermind kidding I am not replicating the same issue. Things work like expected. Have you tried again?

Did you change anything regarding name queues for instance in the `env_file_app` file? that could have break this

Sharing your personal configuration could help identify the issue if you didn't manage to solve it

arakakiv:
I haven’t changed anything except the variables related to AWS, but only those for secrets and IAM. I am running on an Ubuntu machine in the AWS infrastructure; the web application runs normally, but due to Celery errors, the jobs never finish. The secrets could be obtained as they should, and that’s the only problem. I am starting the application with the command `./start test up -- --build`. Finally, I'm cloning the `develop` branch.

arakakiv:
It looks like the `aws_sqs` is not on the settings object too.

mlodic:
can you share the exact same config that you have for the `env_file_app` (obviously redacting secrets)?

ah one thing: which kind of SQS queues have you configured? Cause fifo ones are required. Standard ones won't work

arakakiv:
I didn't configured any SQS queues... Probably this is the problem? SQS queues are required when running on the AWS infrastructure? See below my `env_file_app`:

```
# Required Secrets
DJANGO_SECRET=
DB_HOST=postgres
DB_PORT=5432
DB_USER=
DB_PASSWORD=
DB_SSL=False
DB_NAME=intel_owl_db

# Additional Config variables
# jobs older than this would be flushed from the database periodically. Default: 14 days
OLD_JOBS_RETENTION_DAYS=14
# used for generating links to web client e.g. job results page; Default: localhost
INTELOWL_WEB_CLIENT_DOMAIN=localhost
# used for automated correspondence from the site manager
DEFAULT_FROM_EMAIL=
# used for correspondence with users
DEFAULT_EMAIL=
# Storage
LOCAL_STORAGE=True

# OAuth2
[...]

PR #2567, opened 2024-11-05
fix for correctly handling AWS Secrets

mlodic:
for @drosetti : I added a boolean environment variable for the integration with Elastic to avoid to have empty ("None") environment variables. This would have created problems in case the _AWS_SECRETS_ variable was enabled.

More generally, it would also make more sense to have that boolean flag to enable/disable the feature instead of leveraging an URL-based variable. In this way you don't have to configure the URL multiple times in case of changes but you have to configure a more easily understandable boolean flag.

I updated the doc too: https://github.com/intelowlproject/docs/commit/ecb7c721158eb26a22ad7d1a50329cc4b154c49d

### DIFF
--- a/docker/env_file_app_template
+++ b/docker/env_file_app_template
@@ -53,6 +53,7 @@ SLACK_TOKEN=
 DEFAULT_SLACK_CHANNEL=
 
 # Elastic Search Configuration
+ELASTIC_DSL_ENABLED=False
 ELASTIC_HOST=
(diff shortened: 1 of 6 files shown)


mlodic:
I am early merging it to enable the user [here](https://github.com/intelowlproject/IntelOwl/issues/2554) to solve their problem. In case, please feel free to suggest any changes even afterwards

PR #2584, opened 2024-11-29
added MobSF_Service analyzer

### LINKED ISSUE #2496: [Analyzer] MobSF_Service
## Name
MobSF_Service

## Link
https://github.com/MobSF/mobsfscan

## Type of analyzer
file analysis for APK.

## Why should we use it
We already have a MobSF analyzer for static analysis only of source code elements of an APK.
We want to add support for the Mob_SF APIs, for ones that have the engine available as a service and wants to launch a static or dynamic analysis programmatically at scale.

Optional: it could make sense to add a MOBSF framework as an optional docker container and let it to be used by IntelOwl

## Possible implementation
API docs can be found in the GUI of the project once you install it. http://localhost:8000/api_docs

spoiicy:
Closes #2496 

# Description

Added MobSF_Service analyzer

## Type of change

# Checklist

    - [ ] [Usage](https://github.com/intelowlproject/docs/blob/main/docs/IntelOwl/usage.md) file was updated.
    - [ ] [Advanced-Usage](https://github.com/intelowlproject/docs/blob/main/docs/IntelOwl/advanced_usage.md) was updated (in case the plugin provides additional optional configuration).
    - [ ] If the plugin interacts with an external service, I have created an attribute called precisely `url` that contains this information. This is required for Health Checks. 

### Important Rules


spoiicy:
Screenshot 2024-11-29 at 3 41 34 PM
Attached is the screenshot for the MobSF_Service report and below is the full JSON report containing results for static & dynamic analysis. 
[job#38_report.json](https://github.com/user-attachments/files/17957663/job.38_report.json)

spoiicy:
@mlodic Could you please re-run the workflow since the backend-test CI has failed for some reason. It was passing in the previous commit [ffc968b](https://github.com/intelowlproject/IntelOwl/pull/2584/commits/ffc968b1849661101c98613057d84b673a999865)

api_app/analyzers_manager/migrations/0139_analyzer_config_mobsf_service.py
```diff
@@ -0,0 +1,150 @@
...
+    "routing_key": "default",
+    "health_check_status": True,
+    "type": "file",
+    "docker_based": False,
+    "maximum_tlp": "RED",
+    "observable_supported": [],
+    "supported_filetypes": [
+        "application/vnd.android.package-archive",
+        "application/zip",
+    ],
```
mlodic, line 30: you can use this list for android:

application/java-archive
application/vnd.android.package-archive
application/x-dex
application/zip

api_app/analyzers_manager/migrations/0139_analyzer_config_mobsf_service.py
```diff
@@ -0,0 +1,150 @@
...
+    },
+    "name": "MobSF_Service",
+    "description": "[MobSF_Service](https://github.com/MobSF/Mobile-Security-Framework-MobSF) can be used for a variety of use cases such as mobile application security, penetration testing, malware analysis, and privacy analysis.",
+    "disabled": False,
+    "soft_time_limit": 400,
+    "routing_key": "default",
+    "health_check_status": True,
+    "type": "file",
+    "docker_based": False,
+    "maximum_tlp": "RED",
```
mlodic, line 25: Even if you should have control of the MOBSF Instance, it is still an external service so by definition this should be AMBER

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        binary = self.read_file_bytes()
+
+        upload_url = self.mobsf_host + "/api/v1/upload"
+        upload_response = requests.post(
+            url=upload_url,
+            files={"file": (self.filename, binary, "application/octet-stream")},
+            headers=headers,
+        )
+        upload_response.raise_for_status()
+        scan_hash = upload_response.json()["hash"]
```
mlodic, line 99: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+            data={"hash": scan_hash},
+        )
+        dynamic_analysis_report.raise_for_status()
+
+        return dynamic_analysis_report.json()
+
+    def run(self):
+        headers = {"X-Mobsf-Api-Key": self._mobsf_api_key}
+        binary = self.read_file_bytes()
+
```
mlodic, line 91: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+class MobSF_Service(FileAnalyzer):
+    mobsf_host: str
+    _mobsf_api_key: str
+
+    def update(self) -> bool:
+        pass
+
+    def static_analysis(self, scan_hash, headers):
+
```
mlodic, line 19: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        scan_url = self.mobsf_host + "/api/v1/scan"
+        data = {"hash": scan_hash}
+        scan_response = requests.post(url=scan_url, data=data, headers=headers)
+        scan_response.raise_for_status()
+
```
mlodic, line 24: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        report_url = self.mobsf_host + "/api/v1/report_json"
+        report_response = requests.post(url=report_url, data=data, headers=headers)
+        report_response.raise_for_status()
+        return report_response.json()
+
```
mlodic, line 29: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+    def dynamic_analysis(self, scan_hash, headers):
+
+        mobsfy_runtime_response = requests.post(
+            url=self.mobsf_host + "/api/v1/android/mobsfy",
+            headers=headers,
+            data={"identifier": "127.0.0.1:6555"},
```
mlodic, line 35: this should be a parameter cause it can change between environments

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+from api_app.analyzers_manager.classes import FileAnalyzer
+from tests.mock_utils import MockUpResponse, if_mock_connections, patch
+
+logger = logging.getLogger(__name__)
+
+
+class MobSF_Service(FileAnalyzer):
+    mobsf_host: str
+    _mobsf_api_key: str
+
```
mlodic, line 14: I would add a parameter to enable or not dynamic analysis. Static is the default. Dynamic could be opted out to save time and computation.

api_app/analyzers_manager/file_analyzers/mobsf_service.py
mlodic, line 29: about the static analysis I am worried about execution time. By default, `requests` does not set any timeout. This is potentially dangerous in IntelOwl because it could lead to jobs that never end, creating bottlenecks in the overall application.
IntelOwl cannot rely on external tools. This is the reason why we need to explicity set timeouts to all the `requests` calls that you made in this analyzer. Which value? What your experience with this tool tells you? I think that static analysis could last several seconds. 30 seconds as default could make sense?
Plus, this value should be a parameter of the analyzer and customized so the users are not bound to it.

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        scan_response = requests.post(url=scan_url, data=data, headers=headers)
+        scan_response.raise_for_status()
+
+        report_url = self.mobsf_host + "/api/v1/report_json"
+        report_response = requests.post(url=report_url, data=data, headers=headers)
+        report_response.raise_for_status()
+        return report_response.json()
+
+    def dynamic_analysis(self, scan_hash, headers):
+
```
mlodic, line 31: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        mobsfy_runtime_response = requests.post(
+            url=self.mobsf_host + "/api/v1/android/mobsfy",
+            headers=headers,
+            data={"identifier": "127.0.0.1:6555"},
+        )
+        mobsfy_runtime_response.raise_for_status()
+
```
mlodic, line 38: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        start_dynamic_analysis_response = requests.post(
+            url=self.mobsf_host + "/api/v1/dynamic/start_analysis",
+            headers=headers,
+            data={"hash": scan_hash},
+        )
+        start_dynamic_analysis_response.raise_for_status()
+
```
mlodic, line 45: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        tls_tests_response = requests.post(
+            url=self.mobsf_host + "/api/v1/android/tls_tests",
+            headers=headers,
+            data={"hash": scan_hash},
+        )
+        tls_tests_response.raise_for_status()
+
```
mlodic, line 52: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+            headers=headers,
+            data={
+                "hash": scan_hash,
+                "default_hooks": "api_monitor,ssl_pinning_bypass,root_bypass,debugger_check_bypass",
+                "auxiliary_hooks": "",
+                "frida_code": "",
+            },
+        )
+        start_frida_instrumentation_response.raise_for_status()
+
```
mlodic, line 64: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        get_runtime_dependency_response = requests.post(
+            url=self.mobsf_host + "/api/v1/frida/get_dependencies",
+            headers=headers,
+            data={"hash": scan_hash},
+        )
+        get_runtime_dependency_response.raise_for_status()
+
```
mlodic, line 71: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        stop_dynamic_analysis = requests.post(
+            url=self.mobsf_host + "/api/v1/dynamic/stop_analysis",
+            headers=headers,
+            data={"hash": scan_hash},
+        )
+        stop_dynamic_analysis.raise_for_status()
+
```
mlodic, line 78: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        dynamic_analysis_report = requests.post(
+            url=self.mobsf_host + "/api/v1/dynamic/report_json",
+            headers=headers,
+            data={"hash": scan_hash},
+        )
+        dynamic_analysis_report.raise_for_status()
+
```
mlodic, line 85: add a log here please

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        get_runtime_dependency_response = requests.post(
+            url=self.mobsf_host + "/api/v1/frida/get_dependencies",
+            headers=headers,
+            data={"hash": scan_hash},
+        )
+        get_runtime_dependency_response.raise_for_status()
+
+        stop_dynamic_analysis = requests.post(
+            url=self.mobsf_host + "/api/v1/dynamic/stop_analysis",
```
mlodic, line 73: as you suggested, dynamic analysis time is debatable. This is the reason why I think we should add here a sleep and that this should be a parameter of the analyzer. Which default? 20-30 seconds? What do you think about?

api_app/analyzers_manager/file_analyzers/mobsf_service.py
```diff
@@ -0,0 +1,128 @@
...
+        tls_tests_response.raise_for_status()
+
+        start_frida_instrumentation_response = requests.post(
+            url=self.mobsf_host + "/api/v1/frida/instrument",
+            headers=headers,
+            data={
+                "hash": scan_hash,
+                "default_hooks": "api_monitor,ssl_pinning_bypass,root_bypass,debugger_check_bypass",
+                "auxiliary_hooks": "",
+                "frida_code": "",
```
mlodic, line 60: these 3 values should be parameters of the analyzer. In that way the users can choose how to leverage dynamic analysis as they wish

mlodic (changes requested):
thanks for the PR!

api_app/analyzers_manager/file_analyzers/mobsf_service.py
spoiicy: Though i've not used mobsf extensively but the apps i've tested did not take more than 30 seconds for static analysis. I understand it will be better if we set a parameter for this so that the end-user can modify it according to their needs.

But I wan't to understand, do we need to set this timeout because when I was implementing this analyzer, it was failing at the "dynamic analysis" stage with an error "SoftTimeLimitExceeded". So I had increased the Soft-time limit in the analyzer config.

I do understand that setting a default timeout would be better for error handling but wouldn't just configuring the soft-time limit take care of everything. Let me know what you think!

api_app/analyzers_manager/file_analyzers/mobsf_service.py
spoiicy: As I commented earlier, in static analysis review comment, wouldn't just configuring the soft-time limit do the trick. Because if I keep it at default value, "SoftTimeLimitExceeded" error is thrown and my job fails. I had to set it to 400 in order for my job to be successful. Let me know what you think!

api_app/analyzers_manager/file_analyzers/mobsf_service.py
spoiicy: Sure I'll add them as parameters for users to pass a desired value.

api_app/analyzers_manager/file_analyzers/mobsf_service.py
mlodic: > Even in the worst-case scenario, where our mobsf doesn't respond to the request and the request keeps hanging, wouldn't celery's Soft time limit take care of this and would end the job after specific period of time?

Yes but that would be the _last_ resort and should be treated as an unmanaged exception. It's like creating a generic Exception in python code. Ok, you get when something bad happens but you did not manage it properly cause you don't know what exactly was happening. Pretty much the same thing here. Soft time limit should be treated as the emergency exception.

About adding the timeout, we already did that for other similar services if you check other analyzers like [this](https://github.com/intelowlproject/IntelOwl/blob/master/api_app/analyzers_manager/observable_analyzers/circl_pdns.py#L39). High soft time limits and customizable timeouts, with low timeout values by default.

Regarding the values that you proposed, I think that even more than 400 could be set, maybe 1000 to handle the worst cases scenarios.

api_app/analyzers_manager/file_analyzers/mobsf_service.py
mlodic: see previous comment

api_app/analyzers_manager/file_analyzers/mobsf_service.py
spoiicy: Completely get your point about the _unmanaged exception_. I'll provide an option of request timeout parameter with a reasonable default value which can be modified as per user's need. 

I just want to understand one last thing, let's say a mobsf job has been triggered where default soft-time limit is 60 seconds but now the user has set the request timeout value as 400 seconds. Would this 400 seconds timeout be respected or the job will fail with error "SoftTimeLimitExceeded". 

Because from my limited understanding of the Intelowl, if the alloted time for a job to run is set to 60 seconds, the job will end even if we set the request timeout higher than the soft-time limit. So along with setting the request timeout, we'll have to configure the soft-time limit to a decent time period so that the job succeeds.

Let me know if my understanding is correct/incorrect?

mlodic: > I just want to understand one last thing, let's say a mobsf job has been triggered where default soft-time limit is 60 seconds but now the user has set the request timeout value as 400 seconds. Would this 400 seconds timeout be respected or the job will fail with error "SoftTimeLimitExceeded".

the job would fail with SoftTimeLimitExceeded.

> we'll have to configure the soft-time limit to a decent time period so that the job succeeds.

Exactly. In this way soft time limits exception are limited to unwanted exceptions. I understand that the timeout can be changed by the user in a way that the soft time limit could always be reached but that's something we can ignore to avoid over-engineering this

spoiicy:
Hi @mlodic I've made the changes as per your comments. I hope we are good to go, let me know otherwise. :)

mlodic (changes requested):

api_app/analyzers_manager/migrations/0141_analyzer_config_mobsf_service.py
```diff
@@ -0,0 +1,219 @@
...
+    "python_module": {
+        "health_check_schedule": None,
+        "update_schedule": None,
+        "module": "mobsf_service.MobSF_Service",
+        "base_path": "api_app.analyzers_manager.file_analyzers",
+    },
+    "name": "MobSF_Service",
+    "description": "[MobSF_service](https://github.com/MobSF/Mobile-Security-Framework-MobSF) can be used for a variety of use cases such as mobile application security, penetration testing, malware analysis, and privacy analysis.",
+    "disabled": False,
+    "soft_time_limit": 400,
```
mlodic, line 20: as mentioned in a previous comment, you can extend it to 1000 to allow for broader use cases

api_app/analyzers_manager/file_analyzers/mobsf_service.py
mlodic: I noticed that no sleep was added so basically the analysis has a really brief duration time, just the time needed to execute all the steps between the calls "start" and "stop". As mentioned, I would add a sleep between the "start" and the later calls, to allow time for the stuff to start properly and the code loaded and maybe to allow the malware to do some nasty things too.

mlodic (changes requested):
I have to ask you other stuff to have this PR accepted. Some standards needs to be followed:

* a PR against the documentation repository has been added? Please do that following the checklist and add a link to that PR here.
* the static_analysis and dynamic_analysis functions are really difficult to read and they contain a lot or redundant code. I ask you to refactor that code in a way that there is an helper method that executes all of the requests stuff

spoiicy:
@mlodic I've added the usage for MobSF_Service. [Here](https://github.com/intelowlproject/docs/pull/18) is the pull request.
I've improved the readability of the code by adding helper method to handle request-response and refactored some of the code. 

The PR is ready to be reviewed.

mlodic:
thank you, we are ready to merge it!

PR #2691, opened 2025-01-21
Added timeout parameter malware bazaar end threatfox

federicofantini:


drosetti (approved):

drosetti:
@fgibertoni once the test pass, you can proceed with merge

api_app/ingestors_manager/ingestors/malware_bazaar.py
```diff
@@ -53,20 +55,19 @@ def get_recent_samples(self):
         current_time = timezone.now()
         for signature in self.signatures:
             data = self.get_signature_information(signature)
+            hours_str = "hour" if self.hours == 1 else "hours"
+            if len(data) > self.limit:
+                logger.info(
```
fgibertoni, line 60: `warning` maybe ? So we can monitor how frequently it happens

drosetti: I checked and there are more or less 330k warning per day, if we  want to see it casually (without searching) we should put it to error. Or change the warning policy for other logs (also viable, but maybe not now)

PR #2722, opened 2025-02-10
remove drf_spectacular and not used add_docs

mlodic:


mlodic:
@drosetti @0ssigeno FYI I removed the leftovers of the `drf_spectacular` package that is actually not used. Docstrings should be used as of now.

PR #2885, opened 2025-06-04
HuntingAbuseAPI analyzer

### LINKED ISSUE #2778: [Analyzer] Hunting Abuse.ch
## Name
Hunting_Abuse_Ch

## Link
https://hunting.abuse.ch/api/

## Type of analyzer
ip address, domain, url, hash

## Why should we use it
central point to query data from all abuse.ch services

## Possible implementation
https://hunting.abuse.ch/api/

spoiicy:
Closes #2778 

# Description

This PR aims to add an analyzer to identify if the provided observable is present in the false positive list or not via Hunting Abuse API. 

## Type of change

# Checklist

    - [x] If the plugin interacts with an external service, I have created an attribute called precisely `url` that contains this information. This is required for Health Checks (HEAD HTTP requests). 
    - [x] I have created the corresponding `DataModel` for the new analyzer following the [documentation](https://intelowlproject.github.io/docs/IntelOwl/contribute/#how-to-create-a-datamodel)

### Important Rules


spoiicy:
Hi @fgibertoni, hope you are doing well. The PR is ready to be reviewed.

spoiicy:
**FP found Response**: 

image

**FP not found Response**:
image

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,58 @@
...
+
+import requests
+
+from api_app.analyzers_manager.classes import ObservableAnalyzer
+from tests.mock_utils import MockUpResponse, if_mock_connections, patch
+
+logger = logging.getLogger(__name__)
+
+
+class HuntingAbuseAPI(ObservableAnalyzer):
```
fgibertoni: For abuse.ch services we introduced a mixin called `AbuseCHMixin` that handles the required API key, you just need to add the `authentication_header` property to the other headers and add the Mixin in the superclasses list. You can find some examples about this in ThreatFox or URLHaus analyzers.

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,58 @@
...
+
+        response = requests.post(self.url, json=data, headers=headers)
+        response.raise_for_status()
+        fp_list = response.json()
+
+        for _key, value_dict in fp_list.items():
+            logger.info(f"Fetching fp_status for {self.observable_name}")
+            if value_dict["entry_value"] == self.observable_name:
+                return {"fp_status": "true", "details": value_dict}
+        return {"fp_status": "False"}
```
fgibertoni:
```suggestion
        return {"fp_status": False}
```
So it gets automatically converted in JSON compatible format

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,58 @@
...
+        data = {"query": "get_fplist", "format": "json"}
+
+        response = requests.post(self.url, json=data, headers=headers)
+        response.raise_for_status()
+        fp_list = response.json()
+
+        for _key, value_dict in fp_list.items():
+            logger.info(f"Fetching fp_status for {self.observable_name}")
+            if value_dict["entry_value"] == self.observable_name:
+                return {"fp_status": "true", "details": value_dict}
```
fgibertoni:
```suggestion
                return {"fp_status": True, "details": value_dict}
```
So it gets automatically converted in JSON compatible format

fgibertoni (changes requested):
I think it would be useful to add a Data Model too.
You can check out the doc for guidelines on how to write one (evaluation, reliability and maybe additional details can be mapped) or just check other analyzers like GreyNoise Intel or Crowdsec.
Since it's a relatively new feature feel free to ask for additional guidance if you're having trouble :smile:

AnshSinghal:
Hi @spoiicy I had previously explored this issue a bit and just wanted to share a quick suggestion. Would it make sense to use a TTL-based cache for the false positives list? That way, we can store it once and refresh it periodically, instead of fetching the full list on every search. This could help with performance and reduce API load.
Totally appreciate your work on this — feel free to ignore if you’ve already considered it or have a better approach!

spoiicy:
@fgibertoni As per suggestion from @AnshSinghal, it actually makes sense to store the results for some amount of time, update them after, let's say 1 day, so that the server doesn't have to query on every call. 

And I am aware that we used a file-based approach previously to perform such tasks and now @cristinaascari is working on solution, making it a DB-based approach and refactoring old code. 

So how should I proceed with this? Let me know what you think.

fgibertoni:
Yes I do agree with you and @AnshSinghal. At the moment @cristinaascari is blocked by other work on that task so it may take longer than expected.
I think we can proceed with the "standard" approach and then we will refactor slighlty the code to adapt to her changes.

spoiicy:
## Data Model Result
```
{
  "report": {
    "details": {
      "platform": "MalwareBazaar",
      "entry_type": "md5_hash",
      "removed_by": "user",
      "time_stamp": "2025-06-22 09:45:33 UTC",
      "entry_value": "099137899ece96f311ac5ab554ea6fec",
      "removal_notes": null
    },
    "fp_status": true
  },
  "data_model": {
    "id": 387,
    "analyzers_report": [
      519
    ],
    "signatures": [],
    "evaluation": "trusted",
    "reliability": 9,
    "kill_chain_phase": null,
    "external_references": [],
    "related_threats": [],
    "tags": null,
    "malware_family": null,
    "additional_info": {
      "platform": "MalwareBazaar",
      "entry_type": "md5_hash",
      "removed_by": "user",
      "time_stamp": "2025-06-22 09:45:33 UTC",
      "entry_value": "099137899ece96f311ac5ab554ea6fec",
      "removal_notes": null
    },
    "date": "2025-06-28T19:39:09.202500Z",
    "comments": [],
    "file_information": {},
    "stats": {}
  },
  "errors": [],
  "parameters": {}
}
```

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,103 @@
...
+from api_app.models import PluginConfig
+from tests.mock_utils import MockUpResponse, if_mock_connections, patch
+
+logger = logging.getLogger(__name__)
+
+db_name = "hunting_abuse_fplist.json"
+database_location = f"{settings.MEDIA_ROOT}/{db_name}"
+
+
+class HuntingAbuseAPI(ObservableAnalyzer):
```
fgibertoni: Can you adapt the analyzer to use the `AbuseCHMixin` to avoid using different names and duplicate code per all the services of abuse.ch platform?

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,103 @@
...
+    url: str = "https://hunting-api.abuse.ch/api/v1/"
+    _auth_key: str
+
+    def _do_create_data_model(self) -> bool:
+        return super()._do_create_data_model() and self.report.report.get(
+            "fp_status", False
+        )
+
+    @classmethod
+    def get_auth_key(cls) -> str | None:
```
fgibertoni: This method should also be adapted to use the common Mixin. There shouldn't be problems in setting it up as it remains a parameter like the others

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,103 @@
...
+            parameter__name="auth_key",
+        ):
+            if plugin.value:
+                return plugin.value
+        return None
+
+    @classmethod
+    def update(cls) -> bool:
+        auth_key = cls.get_auth_key()
+        headers = {"Content-Type": "application/json", "Auth-Key": auth_key}
```
fgibertoni: Here the header can be integrated with the default provided by the Mixin. If you check other abuse.ch analyzers you should see an example using the `|` operator for dictionaries

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,103 @@
...
+
+        try:
+            response = requests.post(cls.url, json=data, headers=headers)
+            response.raise_for_status()
+
+            with open(database_location, "w", encoding="utf-8") as f:
+                json.dump(response.json(), f)
+
+            if not os.path.exists(database_location):
+                raise Exception(f"database location {database_location} does not exist")
```
fgibertoni: Here the correct exception should be raised. Maybe an `OSError` can fit the case

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,103 @@
...
+            return True
+        except Exception as e:
```
fgibertoni, line 66: Also here you can use `except Exception` as a last chance. Before this, the appropriate exceptions should be catched and handled

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,103 @@
...
+            logger.error(f"Failed to update Hunting Abuse database: {e}")
+            return False
+
+    def run(self):
+        if not os.path.isfile(database_location):
```
fgibertoni, line 72: On each run if the DB is found it is never updated, right ? It gets updated when the `update` method is run "externally", is this correct ?

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
```diff
@@ -0,0 +1,103 @@
...
+            logger.info("Hunting Abuse database not found, updating...")
+            if not self.update():
+                raise AnalyzerRunException("Failed extraction of Hunting Abuse db")
+
+        with open(database_location, "r", encoding="utf-8") as f:
+            fp_list = json.load(f)
+
+        for _key, value_dict in fp_list.items():
+            is_match = self.observable_name == value_dict["entry_value"] or (
```
fgibertoni, line 83: Some doubt about the logic of this for loop and variable inside:
- Maybe the logical expression can be simplified: is it necessary to check if the observable is an IP and perform the same check ?
- If at iteration `i` out of `n` you set the variable `is_match` to `True` and then in all other iterations it gets set to `False` you lose your result. Maybe you should use a `break` statement or return the result as soon as you find it if you're okay in not checking the entire db. Am I missing something ?

fgibertoni (changes requested):
Nice work! Just some consideration 😁

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
spoiicy: Yes, you are absolutely correct. It just checks if the DB is present or not. If it's not present, it will create a DB file, otherwise it will run with existing DB file.

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
spoiicy: 1. I think this was necessary because, some of the results that we receive from the API are `ip:port` form, so if the user inputs `IP` observable then I'm just checking if the observable is **\**in\**** the **ip:port** string or not.
E.g 
```
2025-06-25 21:35:34 | ThreatFox | ip:port | 147.185.221.29:4449
-- | -- | -- | --
2025-06-25 21:35:26 | ThreatFox | ip:port | 147.185.221.29:46866
```

2. This one will return True, for the first match, it gets. Since the analyzer basically tells if provided observable is a false positive or not, we are just concerned about the first `True` condition and we are returning from there. if the loop doesn't return a `True`, the default result `False`, that is outside the loop, will be returned

Please let me know what you think about this and if you have any queries :)

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
spoiicy: I'm wondering, since we are using **`@classmethod`**, would `authentication_header` be accessible since it would be instance method.

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
spoiicy: @fgibertoni, I was wondering if this would make any sense apart from unifying all the abuse.ch analyzers, because this mixin only provides `authentication_header` to retrieve the API key but we won't be able to leverage this since we have to implement @classmethod update() to cache the API results. 

Feodo Tracker deals with this in similar way, where `service_api_key` has been added in the DB later on with this migration file `0156_alter_analyzer_config_required_api_key_abuse_ch.py`

I can definitely add AbuseCHMixin and keep the auth key name to `service_api_key` just like others to avoid different name but I am unsure if we would be able to avoid some duplicate code. 

Let me know what you think.

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
spoiicy: As I mentioned in this [comment](https://github.com/intelowlproject/IntelOwl/pull/2885#discussion_r2175687903), we might have to implement this method just like FeodoTracker.

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
fgibertoni: I think it would favor uniformity between services. If it's not much pain I'd prefer to do it

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
fgibertoni: Maybe you're right, fixing this would make things too complicated for such a simple thing

api_app/analyzers_manager/observable_analyzers/hunting_abuse.py
fgibertoni: 1. Oh ok now I get the difference. Let's just add a little comment to explain the use case for this. Thanks!
2. Wops my bad. I missed an indentation reading the code on my phone 😬 Completely fine then!

fgibertoni (approved):
Great work, thank you!

PR #2886, opened 2025-06-05
Gsoc25 refactor analyzer tests

pranjalg1331:
(Please add to the PR name the issue/s that this PR would close if merged by using a [Github](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) keyword. Example: `<feature name>. Closes #999`. If your PR is made by a single commit, please add that clause in the commit too. This is all required to automate the closure of related issues.)

# Description

## Type of change

- [ x] New feature (non-breaking change which adds functionality).

# Checklist

- [ x] I have read and understood the rules about [how to Contribute](https://intelowlproject.github.io/docs/IntelOwl/contribute/) to this project
- [ x] The pull request is for the branch `develop`
- [ ] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowlproject.github.io/docs/IntelOwl/contribute/#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
- [ x] I have added tests for the feature/bug I solved (see `tests` folder). All the tests (new and old ones) gave 0 errors.

### Important Rules


pranjalg1331:
Hello mentors, I hope you’re all doing well.
In this PR, as discussed in the proposal, I have implemented a parent test class that runs unittest subtests for all supported analyzer observable types. I also updated the test_nvd_cve to work with this new base test class. Could you please review it before I extend this approach to the other analyzers?

image

pranjalg1331:
@fgibertoni

fgibertoni (commented):
Thanks for your work!
I like the general approach you're proposing here. My only consideration is about the `analyzer_mocks.py` file: I like having a centralized place to hold the configurations but I think that each mock should be related to one or two classes at most. So I don't see the benefits of having a separate file and separate dictionary to hold them and having to update it every time a new analyzer/plugin is created. This also imply that for every plugin there will be a similar file, right ?

I would like also to hear what you think about this, and asking for @mlodic and @drosetti for an opinion

mlodic:
Thank you for asking early review. I agree with Federico. General approach is really good and I love it. I think that the mocks should stay inside the related analyzer files. An idea is to make the BaseAnalyzerTest an ABC class and add an abstract property that must be declared by each Analyzer Test that would contain the patch.
In this way, in case there are some different analyzers that use the same patch, you could use another level of inheritance to make it re-usable between difference instances.

AnshSinghal:
Hi @pranjalg1331, I’ve been following your work on the analyzer test refactoring in #2886, and I’m really impressed with the direction you’re taking! Your approach to organizing the tests and centralizing mocks (like in analyzer_mocks.py) is super clean and promising.

I had played around with a similar test structure a while back when contributing, and I thought it might give you some ideas for your reusable base test classes—especially for handling observable analyzers like NVDDetails, CloudFlareMaliciousDetector, and DNS0EUMaliciousDetector. No pressure at all to use this—I’m just sharing in case it sparks something useful for your design. Feel free to adapt it, ignore it, or tweak it however you see fit!

One thought is to lean more into inheritance to organize both the test logic and mocks in a way that feels natural and maintainable. You could start with a base ObservableAnalyzerTest class that handles the common test structure—like testing all supported observable types. Then, for analyzers that share similar mock needs, you could add intermediate base classes. For example, since CloudFlareMaliciousDetector and DNS0EUMaliciousDetector both rely on DNS-related mocks, an intermediate BaseDNSAnalyzerTest could define that shared mock setup. Each specific analyzer’s test class would inherit from this, keeping the mocks tightly linked to their analyzers while cutting down on repetition.

This could strike a nice balance—keeping mocks close to their relevant classes, as the maintainers emphasized, without needing a separate centralized file that might get unwieldy. It also makes it easier to extend later when new analyzers come along, since you’d just plug them into the right inheritance chain.

Another quick idea: I noticed in BaseAnalyzerTest that you’re skipping tests for observable_type == "generic". Maybe consider adding a brief comment there to explain why—it’d help anyone reading the code understand the reasoning, especially since “generic” observables can be a bit of a wildcard.


Your refactoring is looking solid, and these are just some extra options to play with as you go. Can’t wait to see the final version!

pranjalg1331:
Thanks for the feedback! I tried using an ABC with abstract properties and shared test methods, but Django’s test discovery tries to instantiate all TestCase subclasses—even abstract ones. This causes a TypeError when abstract properties aren’t implemented (So I cannot define a base test inside an abstract class).

fgibertoni:
Can you post a little example of how you did the implementation? So we can help you better :smile:

pranjalg1331:
image
Currently, I have implemented the Base Class without using the ABC class. The analyzer's mocked data is in the same file as the analyzer test class. The base class has a `get_mocked_response()` function which needs to be overridden by every test class that inherits it.

drosetti:
> Thanks for the feedback! I tried using an ABC with abstract properties and shared test methods, but Django’s test discovery tries to instantiate all TestCase subclasses—even abstract ones. This causes a TypeError when abstract properties aren’t implemented (So I cannot define a base test inside an abstract class).

The auto discovery should look for the files called `test_*.py`, if you put the abstract classes into files with different names shouldn't run.

pranjalg1331:
Yes, @drosetti. The base test case is not directly discovered in base_test_class.py, but when another class inherits it, the base test case is also run for the base class.
Thus, with ABC class we get an error like this
`TypeError: Can't instantiate abstract class BaseAnalyzerTest with abstract methods analyzer_class, mock_patch_key`

pranjalg1331:
Hello mentors,

Over the past few days, I’ve been exploring the possibility of defining a common base structure for unit testing file-based analyzers. However, I’ve noticed that these analyzers vary significantly — some are Docker-based, while others like Androguard don’t rely on mock data at all.

Given this diversity, I’m finding it challenging to apply a uniform strategy across all of them. I’d appreciate your thoughts on whether we should still aim for a shared base test structure (similar to what we have for observable analyzers), or if it would be more practical to focus on writing well-structured, analyzer-specific unit tests for file_analyzers.

Looking forward to your guidance. :smile:

AnshSinghal:
Hi @pranjalg1331 ,

I had previously explored this task a bit and just wanted to share a quick suggestion if that can help you. 
How about a base test class, say `BaseFileAnalyzerTest`, that handles common setup like creating test jobs and files. Then, each analyzer can have its own test class inheriting from this base, adding specific mocks for Docker or skipping mocks for analyzers like Androguard. Since analyzers support different file types, the base class could parameterize the file type or have a list of supported types per analyzer, testing each type with `_analyze_sample`. 
Totally appreciate your work on this — feel free to ignore if you’ve already considered it or have a better approach!

fgibertoni:
Hello @pranjalg1331,
I think that the approach suggested by @AnshSinghal should be the way to go. A common class for all file analyzers and then the specific classes for each analyzer. 
You can also extend the base `BaseFileAnalyzerTest` with different subclasses, e.g. `XXXFileAnalyzerTest` for each "classic" file analyzer, `XXXDockerFileAnalyzerTest` that contains specific code for Docker based analyzers. The analyzers that don't rely on mocks should be also treated specifically to use generic testing even without mocks.
What do you guys think ? @mlodic @drosetti 

Also, let me know if you have any other doubts on this :smiley:

pranjalg1331:
Hello @fgibertoni 
I’m a bit hesitant to introduce jobs into our unit tests, because that would move them away from “pure” unit testing and would largely replicate what we already cover elsewhere. 
I’ve also noticed that some analyzers rely on mocked responses, while others—such as APKID and BoxJS—don’t appear to use any. Given that the testing logic varies so much across analyzers with no mocked responses, I’m not sure a common base test class would offer many advantages.

Could you shed some light on why APKID and BoxJS don’t include mocked data? Is that a deliberate choice, or simply an area we haven’t addressed yet?

mlodic:
> Could you shed some light on why APKID and BoxJS don’t include mocked data? Is that a deliberate choice, or simply an area we haven’t addressed yet?

mocked data has been introduced at a later point and those are old analyzer, that's the reason. Ideally, all analyzers should have a decent mock.

mlodic:
> I’m a bit hesitant to introduce jobs into our unit tests, because that would move them away from “pure” unit testing and would largely replicate what we already cover elsewhere. 

It makes sense to me

> Given this diversity, I’m finding it challenging to apply a uniform strategy across all of them. I’d appreciate your thoughts on whether we should still aim for a shared base test structure (similar to what we have for observable analyzers), or if it would be more practical to focus on writing well-structured, analyzer-specific unit tests for file_analyzers.

To me, it makes sense to use common strategies for only the analyzers that are similar to each other. Yes, there are differences, but they are not many. I think that you can try starting low and easy, implementing the tests for each analyzer and when you find two of them that are different to all the others but similar to each other, create a structure for them.

What Federico suggested "_e.g. XXXFileAnalyzerTest for each "classic" file analyzer, XXXDockerFileAnalyzerTest that contains specific code for Docker based analyzers_.", it could be an idea. The important thing is to avoid repetition and keep the code clean. When we/you see repetitive patterns, then create an additional structure to handle it

pranjalg1331:
Hello @fgibertoni,

I've implemented a base class for file analyzers that supports both Docker-based and non-Docker analyzers. It works by loading sample files based on mimetypes and mocking any external dependencies. I've also written unit tests for a few analyzers for you to review.

I'd appreciate it if you could take a look at the implementation and share your thoughts—especially on the maintainability of the current structure—before I begin scaling it to all of the analyzers.

Thank you!

tests/api_app/analyzers_manager/unit_tests/file_analyzers/base_test_class.py
```diff
@@ -0,0 +1,155 @@
...
+            for patch_obj in patches:
+                stack.enter_context(patch_obj)
+            return stack
+
+        # Single patch object
+        return patches
+
+    def test_analyzer_on_supported_filetypes(self):
+        if self.analyzer_class is None:
+            self.skipTest("analyzer_class is not set")
```
mlodic: print test name so that we can track where is the problem when this trigger

tests/api_app/analyzers_manager/unit_tests/file_analyzers/base_test_class.py
```diff
@@ -0,0 +1,155 @@
...
+        config = AnalyzerConfig.objects.get(
+            python_module=self.analyzer_class.python_module
+        )
+        print(config)
```
mlodic: leftover / or use logging.debug

tests/api_app/analyzers_manager/unit_tests/file_analyzers/base_test_class.py
```diff
@@ -0,0 +1,155 @@
...
+            supported_types = self.get_all_supported_mimetypes()
+
+        for mimetype in supported_types:
+            with self.subTest(mimetype=mimetype):
+
+                try:
+                    file_bytes = self.get_sample_file_bytes(mimetype)
+                except (ValueError, OSError):
+                    print(f"SKIPPING {mimetype}")
+                    continue
```
mlodic, line 143: specify test case in the log and use logging

pranjalg1331:
Hello Mentor,
Just a quick update on the PR and its progress:
    Test cases for almost all the analyzers have been successfully written.
    I’ve also cleaned up the existing monkeypatching.
    Currently, I’m working on configuring the CI pipelines.

Documentation- https://github.com/intelowlproject/docs/pull/38
Please let me know if there’s anything specific you'd like me to focus on next. (I am working on this pr's errors) :smile:

pranjalg1331:
Hello @fgibertoni,

I’ve been trying to run my unittests in the PR automation pipeline, but when I run them alongside other tests, I keep getting a "Python module not found" error.

Screenshot from 2025-08-08 16-03-41

My hunch is that there might be a pre-existing test that’s removing Python modules during execution, but I haven’t been able to track it down. Could you help point me in the right direction?

drosetti:
I think you assumption is correct: i run locally `tests.api_app.analyzers_manager.unit_tests.observable_analyzers.test_zoomeye.ZoomEyeTestCase.test_analyzer_on_supported_observables` on your branch and it pass, so it's not a test problem.

I don't have a "real solution" for the problem, my suggestions are:
* Try to run your test in subset each time bigger than before like: `tests.api_app.analyzers_manager.unit_tests.observable_analyzers.test_zoomeye` then `tests.api_app.analyzers_manager.unit_tests.observable_analyzers` and again `tests.api_app.analyzers_manager.unit_tests` and so on, in this way youcan find where the tests start to fail and which part of the code base contains the problem.
* You can search in the code a deletion, like `.delete()` and discover which unittest delete the plugin config.

tests/api_app/analyzers_manager/unit_tests/file_analyzers/base_test_class.py
mlodic: can you please address this comment?

tests/api_app/analyzers_manager/unit_tests/file_analyzers/base_test_class.py
```diff
@@ -0,0 +1,198 @@
...
+        if self.analyzer_class:
+            analyzer_module = self.analyzer_class.__module__
+            logging.getLogger(analyzer_module).setLevel(logging.NOTSET)
+            logging.getLogger("api_app.analyzers_manager").setLevel(logging.NOTSET)
+
+    def test_analyzer_on_supported_filetypes(self):
+        if self.analyzer_class is None:
+            self.skipTest("analyzer_class is not set")
+
+        logger.info("Starting file analyzer test for: %s", self.analyzer_class.__name__)
```
mlodic: anything against using more modern string formatting with {self.analyzer_class.__name__}?

I see you used it also in other lines. It would be great if you could use that most recent one also there because it has better performance under the hood

tests/api_app/analyzers_manager/unit_tests/observable_analyzers/base_test_class.py
```diff
@@ -0,0 +1,161 @@
...
+logger = logging.getLogger(__name__)
+
+
+class BaseAnalyzerTest(TestCase):
+    analyzer_class = None
+    suppress_analyzer_logs = True
+
+    def setUp(self):
+        super().setUp()
+        logger.info("Setting up test environment")
```
mlodic: can you log analyzer name too here?

tests/api_app/analyzers_manager/unit_tests/observable_analyzers/base_test_class.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if self.suppress_analyzer_logs and self.analyzer_class:
+            analyzer_module = self.analyzer_class.__module__
+            logging.getLogger(analyzer_module).setLevel(logging.CRITICAL)
+            logging.getLogger("api_app.analyzers_manager").setLevel(logging.WARNING)
+
+    def tearDown(self):
+        super().tearDown()
+        logger.info("Tearing down test environment")
```
mlodic: can you log analyzer name too here?

tests/api_app/analyzers_manager/unit_tests/observable_analyzers/base_test_class.py
```diff
@@ -0,0 +1,161 @@
...
+            setattr(analyzer, key, value)
+
+        return analyzer
+
+    def _validate_response(self, response, observable_type):
+        if isinstance(response, str):
+            try:
+                response = json.loads(response)
+            except json.JSONDecodeError:
+                logger.error("Invalid JSON response for %s", observable_type)
```
mlodic: same question as before regarding use of old string formatting in this file

tests/api_app/analyzers_manager/unit_tests/file_analyzers/test_hfinger.py
```diff
@@ -0,0 +1,67 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+from hfinger.analysis import hfinger_analyze
+
+from api_app.analyzers_manager.classes import FileAnalyzer
+from tests.mock_utils import if_mock_connections, patch
+
+
+class Hfinger(FileAnalyzer):
```
fgibertoni: Is the parent class a typo in this case ?

tests/api_app/analyzers_manager/unit_tests/file_analyzers/base_test_class.py
pranjalg1331: updated

tests/api_app/analyzers_manager/unit_tests/observable_analyzers/base_test_class.py
pranjalg1331: updated

tests/api_app/analyzers_manager/unit_tests/file_analyzers/base_test_class.py
pranjalg1331: updated

tests/api_app/analyzers_manager/unit_tests/file_analyzers/test_hfinger.py
pranjalg1331: Yes, it was a typo. I have updated it.

pranjalg1331:
@fgibertoni The tests inside  _tests/api_app/analyzers_manager/file_analyzers/_ are more integration-focused (running real jobs), while the unit tests mock everything. Because of that, we can’t merge them into the unit_tests folder, but we could organize them better by moving them into a dedicated integration_tests directory.
I agree with your second point — that approach works well and can be done directly in the analyzer’s unit test class. Example: test_phishing_form_compiler.py → If you want to test a single function in isolation, add a unit test that mocks all external dependencies.

fgibertoni:
I agree with moving them to keep the project folder tree cleaner. Since those tests are more specific and rare to be written than yours, I would move them into a subfolder, so we can keep the "default" tests easier to find.

As for the second point: great. And those custom tests will be ran together with the pipeline automatically, right?

pranjalg1331:
Yes, they will run automatically with the updated pipeline.

mlodic:
could you please update the tests for the newly added analyzer Guarddog? https://github.com/intelowlproject/IntelOwl/pull/2930

There is only another one left that Akshit is working on, otherwise no other analyzer will be added during this phase . Reference: https://github.com/intelowlproject/IntelOwl/pull/2933

EDITED: Nevermind, the GuardDog analyzer needs rework so it is not needed as of now FYI @spoiicy

PR #2933, opened 2025-07-24
Floss Capa Refactor

### LINKED ISSUE #2738: Move Flare Capa and Floss from binary download to pip installation
Right now those two tools are installed from the malware_tools_analzyers Docker Integration by downloading the binary.
This solution is not optimal because it makes difficult to track updates and the downloaded binary is made for a single arch.
We should try to rewrite the analyzers to leverage the officialy supported python libraries.
See: https://pypi.org/project/flare-capa/ and https://pypi.org/project/flare-floss/

spoiicy:
Closes #2738 

# Description

This PR aims to update the Flare Capa and Flare Floss analyzers, by refactoring the exisiting implementation and utilizing the pip packages for better support. 

## Type of change

# Checklist

### Important Rules


spoiicy:
## Capa Results

image

## Floss Results

```
{
  "report": {
    "strings": {
      "stack_strings": [],
      "tight_strings": [],
      "static_strings": [],
      "decoded_strings": [],
      "language_strings": [],
      "language_strings_missed": []
    },
    "analysis": {
      "functions": {
        "library": 23,
        "discovered": 153,
        "analyzed_stack_strings": 104,
        "analyzed_tight_strings": 0,
        "analyzed_decoded_strings": 20,
        "decoding_function_scores": {
          "4198416": {
            "score": 0.6,
            "xrefs_to": 1
          },
          "4198640": {
            "score": 0.6,
            "xrefs_to": 1
          },
          "4198704": {
            "score": 0.6,
            "xrefs_to": 1
          },
          "4198848": {
            "score": 0.6,
            "xrefs_to": 1
          },
          "4203136": {
            "score": 0.607,
            "xrefs_to": 2
          },
          "4203440": {
            "score": 0.682,
            "xrefs_to": 10
          },
          "4203984": {
            "score": 0.571,
            "xrefs_to": 4
          },
          "4204592": {
            "score": 0.901,
            "xrefs_to": 72
          },
          "4204960": {
            "score": 0.714,
            "xrefs_to": 22
          },
          "4205664": {
            "score": 0.771,
            "xrefs_to": 2
          },
          "4206107": {
            "score": 0.693,
            "xrefs_to": 2
          },
          "4206162": {
            "score": 0.796,
            "xrefs_to": 25
          },
          "4206224": {
            "score": 0.7,
            "xrefs_to": 84
          },
          "4206869": {
            "score": 0.693,
            "xrefs_to": 2
          },
          "4207819": {
            "score": 0.732,
            "xrefs_to": 2
          },
          "4208097": {
            "score": 0.579,
            "xrefs_to": 4
          },
          "4208772": {
            "score": 0.85,
            "xrefs_to": 2
          },
          "4209782": {
            "score": 0.656,
            "xrefs_to": 2
          },
          "4210331": {
            "score": 0.745,
            "xrefs_to": 1
          },
          "4210406": {
            "score": 0.745,
            "xrefs_to": 1
          }
        }
      },
      "enable_stack_strings": true,
      "enable_tight_strings": true,
      "enable_static_strings": false,
      "enable_decoded_strings": true
    },
    "metadata": {
      "runtime": {
        "total": 6.5962,
        "vivisect": 3.2142,
        "start_date": "2025-07-22T04:05:41.936423Z",
        "find_features": 0.2296,
        "stack_strings": 0.2385,
        "tight_strings": 0.0003,
        "static_strings": 0,
        "decoded_strings": 2.6583,
        "language_strings": 0
      },
      "version": "3.1.1",
      "language": "unknown",
      "file_path": "/opt/deploy/files_required/06ebf06587b38784e2af42dd5fbe56e5",
      "imagebase": 4194304,
      "min_length": 4,
      "language_version": "version unknown",
      "language_selected": ""
    },
    "exceeded_max_number_of_strings": {}
  },
  "data_model": null,
  "errors": [],
  "parameters": {
    "rank_strings": {
      "stack_strings": true,
      "static_strings": false,
      "decoded_strings": false
    },
    "max_no_of_strings": {
      "stack_strings": 1000,
      "static_strings": 1000,
      "decoded_strings": 1000
    }
  }
}
```

spoiicy:
### Couple of points to note with this PR. 
1. `flare-capa` pip package doesn't come with `capa-rules` and `signatures` by default unlike the binary file, so the update method dynamically pulls the latest versions for both. 
2. `flare-capa` usually takes couple of seconds to analyze the given executable though when specifically tested on `main.out` test file from tests.zip, it takes enormous time to analyze (around 1000 seconds). 
As a result, `soft_time_limit` parameter has been updated to 1800 seconds and a new `timeout` parameter has been added which can be set according to user's requirement.

drosetti:
The code looks good to me, however I tried to replicate your analysis (main.out sample) but I had some problems with the updates.

Trying from the plugin pages I got an error message. In the logs there are these:
```
2025-08-01 16:07:22,183 - api_app.analyzers_manager.file_analyzers.capa_info - _unzip - INFO - Rules have been succesfully extracted
2025-08-01 16:07:22,184 - api_app.analyzers_manager.file_analyzers.capa_info - _download_signatures - INFO - Downloading signatures at /opt/deploy/files_required/capa/sigs now
2025-08-01 16:07:22,526 - api_app.analyzers_manager.file_analyzers.capa_info - _download_signatures - ERROR - Failed to download signature: Command '['/usr/bin/wget', '-O', '/opt/deploy/files_required/capa/sigs', 'https://raw.githubusercontent.com/mandiant/capa/master/sigs/1_flare_msvc_rtf_32_64.sig']' returned non-zero exit status 1.
2025-08-01 16:07:22,526 - api_app.analyzers_manager.file_analyzers.capa_info - update - ERROR - Failed to update capa rules with error: Failed to update signatures due to error: None
```
I tried to replicate the wget command from my command line and seems to work, am I missing something ?

spoiicy:
Hi @drosetti I've made some fixes. It should work now.

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,182 @@
...
+        if not os.path.exists(SIGNATURE_LOCATION):
+            os.makedirs(SIGNATURE_LOCATION)
+
+        signatures_url = "https://api.github.com/repos/mandiant/capa/contents/sigs"
+        response = requests.get(signatures_url)
+        signatures_list = response.json()
+
+        for signature in signatures_list:
+            try:
+                subprocess.run(
```
fgibertoni: Just to be extra sure, I'd like to make `shlex` handle splitting and quoting of parameters before passing them into `run()`

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,182 @@
...
+            # Setting default signatures location
+            command.append("-s")
+            command.append(SIGNATURE_LOCATION)
+
+            command.append(self.filepath)
+
+            logger.info(f"Starting CAPA analysis for {self.filename}")
+
+            process: subprocess.CompletedProcess = subprocess.run(
+                command,
```
fgibertoni, line 141: Same here

api_app/analyzers_manager/file_analyzers/floss.py
```diff
@@ -29,30 +32,39 @@ def update(cls) -> bool:
...
         # retrieve them with more simple tools
-        args = [f"@{fname}", "--json", "--no", "static"]
-        req_data = {"args": args, "timeout": self.timeout}
-        req_files = {fname: binary}
-        result = self._docker_run(req_data, req_files)
-        if not isinstance(result, dict):
+        try:
+            process: subprocess.CompletedProcess = subprocess.run(
+                [
+                    "/usr/local/bin/floss",
```
fgibertoni, line 43: Same here

fgibertoni (changes requested):
Nice work!

drosetti:
@spoiicy now I can download correctly the rules, but when I run the analysis I get an error:
`Analyzer for main.out failed with error: ERROR    capa: [Errno 13] Permission denied:                         main.py:678\n         '/opt/deploy/intel_owl/.cache/capa`

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,182 @@
...
+                        "/usr/bin/wget",
+                        "-P",
+                        SIGNATURE_LOCATION,
+                        signature["download_url"],
+                    ],
+                    check=True,
+                    capture_output=True,
+                )
+
+            except subprocess.CalledProcessError as e:
```
mlodic: If possible I would like to prefer avoid using subprocess for a tool (wget) that performs HTTPS connections that can be handled with classic HTTP libraries. Can you change this?

api_app/analyzers_manager/file_analyzers/capa_info.py
fgibertoni: What about quoting directly `command` before passing it into `run()` ?

fgibertoni (commented):
Please also update migration number since we merged the other PR. Then we should be good :smiley:

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: @fgibertoni  Though this command doesn't have involve any variable that directly comes from user input but still I've quoted `self.filepath` for our peace of mind. Also since the subprocess.run() in it's default implementation doesn't invoke the "sophisticated" shell, I think this would be sufficient. 

Also as mentioned [here](https://semgrep.dev/docs/cheat-sheets/python-command-injection#:~:text=a%20single%20string.-,Use%20shlex.split%20to%20correctly%20parse%20a%20command%20string%20into%20an%20array%20and%20shlex.quote%20to%20correctly%20sanitize%20input%20as%20a%20command%2Dline%20parameter.,-Do%20not%20include), we would first need to use the use `shlex.split` to parse the command into an array, which has already been done and is redundant, and then use `shlex.quote` to escape any possible dangerous user input.

This was my reasoning for this particular implementation. Let me know what you think :)

fgibertoni: Okay I see your point, I was being too cautious 😬

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+            command.append("-r")
+            command.append(RULES_LOCATION)
+
+            # Setting default signatures location
+            command.append("-s")
+            command.append(SIGNATURE_LOCATION)
+
+            command.append(quote(self.filepath))
+
+            logger.info(f"Starting CAPA analysis for {self.filename}")
```
mlodic: I would use md5 here too so we can universally differentiate these logs from one another

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+            process: subprocess.CompletedProcess = subprocess.run(
+                command,
+                capture_output=True,
+                text=True,
+                timeout=self.timeout,
+                check=True,
+            )
+
+            result = json.loads(process.stdout)
+            logger.info("CAPA analysis successfully completed")
```
mlodic: I would use md5 here too so we can universally differentiate these logs from one another

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+
+        except subprocess.CalledProcessError as e:
+            stderr = e.stderr
+            logger.info(f"Capa Info failed to run for {self.filename} with command {e}")
```
mlodic: I would use md5 here too so we can universally differentiate these logs from one another

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+            raise AnalyzerRunException(
+                f" Analyzer for {self.filename} failed with error: {stderr}"
```
mlodic: I would use md5 here too so we can universally differentiate these logs from one another

api_app/analyzers_manager/migrations/0164_update_capa.py
```diff
@@ -0,0 +1,37 @@
...
+from django.db import migrations
+
+
+def migrate(apps, schema_editor):
+    PythonModule = apps.get_model("api_app", "PythonModule")
+    Parameter = apps.get_model("api_app", "Parameter")
+    AnalyzerConfig = apps.get_model("analyzers_manager", "AnalyzerConfig")
+
+    pm = PythonModule.objects.get(
+        module="capa_info.CapaInfo",
```
mlodic: if I remember correctly there is also a param called "docker_based" that should be set to False in the migration to both the analyzers

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
-        if self.shellcode:
-            self.args.append("-f")
-            self.args.append("sc" + self.arch)
+        except Exception as e:
+            logger.error(f"Failed to download signature: {e}")
+            raise AnalyzerRunException("Failed to update signatures")
+        logger.info("Successfully updated singatures")
+
+    @classmethod
+    def update(cls) -> bool:
```
mlodic: we need also to set up a cronjob to automatically update these rules cause otherwise they will be obsolete soon because they are updated only once when those files are not available while on the repositories they update them continously.

If I remember correctly, you should update the "update_schedule" value in the migration for both of the analyzers. See how "Maxmind" analyzer if configured for an example of an already working one

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+                    os.path.isdir(RULES_LOCATION) and os.path.isdir(SIGNATURE_LOCATION)
+                )
+                and not self.update()
+            ):
+
+                raise AnalyzerRunException(
+                    "Couldn't update capa rules or signatures successfully"
+                )
+
+            command: list[str] = ["/usr/local/bin/capa", "--quiet", "--json"]
```
mlodic, line 120: can you please add a key in the final JSON result with the full command that has been executed? that can be extremely useful for debugging purposes plus it would be a way to transparently say to the user what happened under the hood

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
-        req_data = {"args": args, "timeout": self.timeout}
-        req_files = {fname: binary}
-
-        return self._docker_run(req_data, req_files)
+        try:
+            if (
+                not (
+                    os.path.isdir(RULES_LOCATION) and os.path.isdir(SIGNATURE_LOCATION)
+                )
+                and not self.update()
```
mlodic: can you please add an additional parameter that is disabled by default but, when enabled, it would force the capa signatures/rules download at the time of execution of this analyzer? This is important because in this way the user can force the execution with the most recent rules. Otherwise this could bring the user in situations where they expect a result but it's not there and they can't explain why.

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+        logger.info("Successfully updated singatures")
+
+    @classmethod
+    def update(cls) -> bool:
+        try:
+            logger.info("Updating capa rules and signatures")
+            response = requests.get(
+                "https://api.github.com/repos/mandiant/capa-rules/releases/latest"
+            )
+            latest_version = response.json()["tag_name"]
```
mlodic, line 75: can you please log the version?

plus, an ideal solution would be to save the latest downloaded version in a new separate model. Then, when a new update is requested, compare the new version with the latest downloaded one and abort in case the version is the same. This would optimize the caching mechanism

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+
+            process: subprocess.CompletedProcess = subprocess.run(
+                command,
+                capture_output=True,
+                text=True,
+                timeout=self.timeout,
+                check=True,
+            )
+
+            result = json.loads(process.stdout)
```
mlodic, line 148: when we save the capa rules version, please log them in the final JSON so that this information is available to the user.

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+
+        if not os.path.exists(RULES_LOCATION):
+            os.makedirs(RULES_LOCATION)
+
+        file_to_download = latest_version + ".zip"
+        file_url = RULES_URL + file_to_download
+        try:
+
+            response = requests.get(file_url, stream=True)
+            logger.info(f"Started downloading rules from {file_url}")
```
mlodic: log the version

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+            with open(RULES_FILE, mode="wb+") as file:
+                for chunk in response.iter_content(chunk_size=10 * 1024):
+                    file.write(chunk)
+
+        except Exception as e:
+            logger.error(f"Failed to download rules with error: {e}")
+            raise AnalyzerRunException("Failed to download rules")
+
+        logger.info(f"Rules have been successfully downloaded at {RULES_LOCATION}")
```
mlodic: log the version

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+
+        file_to_download = latest_version + ".zip"
+        file_url = RULES_URL + file_to_download
+        try:
+
+            response = requests.get(file_url, stream=True)
+            logger.info(f"Started downloading rules from {file_url}")
+            with open(RULES_FILE, mode="wb+") as file:
+                for chunk in response.iter_content(chunk_size=10 * 1024):
+                    file.write(chunk)
```
mlodic: how does this signature/rule update works when we want to update the existing rules? I would expect a full replace of the old content.

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,184 @@
...
+
+        except Exception as e:
+            logger.error(f"Failed to download rules with error: {e}")
+            raise AnalyzerRunException("Failed to download rules")
+
+        logger.info(f"Rules have been successfully downloaded at {RULES_LOCATION}")
+
+    @classmethod
+    def _download_signatures(cls) -> None:
+        logger.info(f"Downloading signatures at {SIGNATURE_LOCATION} now")
```
mlodic, line 37: the "signatures" are less important than the rules ones. These are almost never updated while the "rules" are are updated often. Plus, most of the time, we don't want these signatures to execute either cause it would slow the Capa execution. The rules are always necessary because they are the core part of the tool while these one could not be necessary. Because of that, I would not re-update them once they are here, like you already do. But we need another additional parameter for the user to enable them explicitly otherwise it would be better if these signatures would be disabled by default.

mlodic: reference: https://github.com/mandiant/capa/tree/master/sigs

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: yes it compeletly overwrites the old rule file.

mlodic: would this remove old files that are not available in the rules anymore? this is also why I am suggesting to delete and re-write

mlodic: so we have a cleanup situation

spoiicy: So basically what this piece of code does is, it takes `capa_rules.zip` file and everytime `_download_rules` function is called it re-writes this same file with new bytes which will contain the latest rules from the repository.

As a result, we don't need to delete and re-write, everything is handled automatically. 

Let me know if this answer was helpful.

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Sure I'll add a parameter `pull_latest`

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Sure this can be good optimisation, I'll make the appropriate changes.

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: > Plus, most of the time, we don't want these signatures to execute either cause it would slow the Capa execution.

Regarding your point, Actually I've tried executing the flare-capa without the signatures but it threw an error, when only executed with rules. So, I feel the signatures are necessary for it's execution.

Though, I can definitely make changes in the code that the signatures are only downloaded once or updated on-demand by the user.

api_app/analyzers_manager/file_analyzers/capa_info.py
fgibertoni: I think Matteo was suggesting to add a simple delete operation before rewriting the entire file in filesystem. So it's a bit "cleaner" than re-writing on the same file over and over again. I agree with Matteo to add an explicit delete operation tbh.
Let me know what you think :smile:

spoiicy: Definitely I can add an explicit delete operation. Initially I was trying to find a reason to add delete operation since it's automatically handled by the file operations itself internally. 

But no issues, I'll add it ☺️

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: done

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: done

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Done

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Done

api_app/analyzers_manager/migrations/0164_update_capa.py
spoiicy: Done

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: made the changes such that signatures are only downloaded the first or whenever `force_pull_signatures` is set to True.

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Made the changes such that on every update old dir is deleted and fresh new directory with new rules file is created.

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Done

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Done

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Done

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: created new model `AnalyzerRulesFileVersion` to track the last downloaded version

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: I've not added as this would be redundant as the code is anyways checking if the latest rules are being used or not. If the not it updates the rules file with latest one.

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Added

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: I've update the migration file with appropriate cron values.

api_app/analyzers_manager/migrations/0168_update_capa.py
```diff
@@ -0,0 +1,64 @@
...
+        pm.full_clean()
+        pm.save()
+
+    AnalyzerConfig.objects.filter(python_module=pm).update(soft_time_limit=1800)
+    AnalyzerConfig.objects.filter(python_module=pm).update(docker_based=False)
+
+    p1 = Parameter(
+        name="timeout",
+        type="float",
+        description="Duration in seconds for which intelowl waits for capa to return results. Default is set to 15 seconds.",
```
fgibertoni, line 37: Maybe I am missing something, but have you set the default of the new parameter?

spoiicy: You are right, I've not set the default for this one. No worries I'll update and add it. :)

fgibertoni: Perfect, thank you!

requirements/project-requirements.txt
```diff
@@ -91,8 +91,10 @@ wad==0.4.6
 debloat==1.6.4
 phonenumbers==9.0.3
 die-python==0.4.0
-guarddog==2.1.0 # version greater than 2.1.0 raises dependency conflicts
+# guarddog==2.1.0 # version greater than 2.1.0 raises dependency conflicts. Commenting this out due to dependency conflicts.
```
mlodic, line 94: I am fine with this but I think we should disable this Guarddog analyzer by default until there is a valid solution to avoid users running into it and getting fails.

api_app/analyzers_manager/models.py
```diff
@@ -350,3 +350,13 @@ def plugin_type(cls) -> str:
     @property
     def config_exception(cls):
         return AnalyzerConfigurationException
+
+
+class AnalyzerRulesFileVersion(models.Model):
+    last_downloaded_version = models.CharField(max_length=50, blank=True, null=True)
```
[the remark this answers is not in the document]
mlodic: I think this should be solved before merging

api_app/analyzers_manager/models.py
```diff
@@ -350,3 +350,13 @@
...
+    download_url = models.URLField(max_length=200, blank=True, null=True)
```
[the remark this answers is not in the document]
mlodic: I think this should be solved before merging

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,251 @@
...
+            stdout='{"meta": {}, "rules": {"contain obfuscated stackstrings": {}, "enumerate PE sections":{}}}',
+            stderr="",
+        )
+        patches = [
+            if_mock_connections(
+                patch.object(CapaInfo, "update", return_value=True),
+                patch("subprocess.run", return_value=response_from_command),
+            )
+        ]
+        return super()._monkeypatch(patches)
```
mlodic: this can be removed

api_app/analyzers_manager/file_analyzers/capa_info.py
```diff
@@ -1,40 +1,251 @@
...
+    def _check_if_latest_version(self, latest_version: str) -> bool:
+
+        analyzer_rules_file_version = AnalyzerRulesFileVersion.objects.filter(
+            python_module=self.python_module
+        ).first()
+
+        if analyzer_rules_file_version is None:
+            return False
+
+        return latest_version == analyzer_rules_file_version.last_downloaded_version
```
mlodic: I would add a log file here before this comparison. In particular it would great to write there the new and the old version number.

mlodic (changes requested):
great job! I added only very few things that could help but then the work is done!

api_app/analyzers_manager/file_analyzers/capa_info.py
spoiicy: Sure, makes sense.

requirements/project-requirements.txt
spoiicy: So what do you suggest here. Do you want me update the migration to set "disabled" to true for this analyzer as well. Let me know your thoughts on this.

mlodic: yep!

mlodic (approved):

api_app/analyzers_manager/migrations/0168_update_capa.py
fgibertoni: Has this been updated?

spoiicy: Yes I've updated this in `update_capa.py` migration. Below is the snippet. 
```
    plugin_config_to_create = [
        PluginConfig(analyzer_config=config, parameter=p1, value=15)
        for config in analyzer_configs
    ]

    PluginConfig.objects.bulk_create(plugin_config_to_create)
```

PR #2935, opened 2025-07-29
fix: switch Hybrid Analysis /search/hash from POST to GET (API v2.35)

ali90h:
# Description
The **Hybrid Analysis** analyzer (`ha_get.py`) was still querying  
`/search/hash` with **POST**. Starting from API v2.35 this endpoint  
accepts **only GET** requests.

This patch  
* replaces `requests.post` with `requests.get`, and  
* moves the payload from `data=` to `params=` (`{"hash": <observable>}`).

No breaking changes are expected; the analyzer now fully complies with  
the current Hybrid Analysis API and avoids failure once POST is retired.

## Type of change
- [x] **Bug fix** (non‑breaking change which fixes an issue)  
- [ ] New feature  
- [ ] Breaking change  

# Checklist
- [x] I have read and understood the rules about **how to Contribute** to this project.  
- [x] The pull request is for the branch **master**.  
- [ ] A new plugin (analyzer, connector, …) was added or changed *(N/A)*.  
- [ ] GUI was modified *(N/A)*.  
- [ ] External libraries with restrictive licences were added *(N/A)*.  
- [x] Linters (`Black`, `Flake`, `Isort`) report **0 errors**.  
- [ ] Tests added/updated where relevant *(N/A – existing logic unchanged)*.


ali90h:
This should fix the deprecation. Let me know if any change is needed. 🙂

fgibertoni (changes requested):
The change impacts also the return type of the endpoint, so the code needs to be re-adapted to return the final results as did before. https://hybrid-analysis.com/docs/api/v2
Plus, open PR only for the `develop` branch and do not delete the full PR checklist please.

ali90h:
Hi @fgibertoni,
I’ve re-adapted the return type of the endpoint to match the previous behavior as requested.
Please let me know if any further changes are needed.
Thank you!

fgibertoni:
For the endpoint, I mean that the old API returned directly a sample as result as `SampleSummary`, this new endpoint is returning a different value for the same search. We probably need another API call to return the same result as the previous endpoint.

PR #2953, opened 2025-08-08
Fix websocket building on non standard ports - closes #2952

### LINKED ISSUE #2952: Job tracking web sockets broken for non standard ports
## What happened
When running on a nonstandard port the websockset connection for tracking job progress fails. 

## Environment
1. OS: Linux Fedora  - Firefox 140.0
2. IntelOwl version: 6.1.0 - current I think

Websockets are built with the same port that the site is hosted on

Start the webservice on a non port 80,433 servive by updating the docker-compose definition

Image

## Error messages and logs

Image

faux-eccles:
Fix connection issue when intelowl is on nonstandard port

image

# Description
Change the way the websocket address are built to ensure the hosts ports are being referenced.  https://github.com/intelowlproject/IntelOwl/issues/2952.

image

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue).

# Checklist

- [X ] I have read and understood the rules about [how to Contribute](https://intelowlproject.github.io/docs/IntelOwl/contribute/) to this project
- [X ] The pull request is for the branch `develop`
- [ ] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowlproject.github.io/docs/IntelOwl/contribute/#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.

### Important Rules

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -55,6 +55,7 @@
         "eslint-plugin-jsx-a11y": "^6.10.0",
         "eslint-plugin-react": "^7.37.1",
         "eslint-plugin-react-hooks": "^4.6.2",
+        "http-proxy-middleware": "^2.0.6",
         "jest": "^29.7.0",
         "jest-environment-jsdom": "^29.7.0",
         "prettier": "^3.3.3",


drosetti:
Hi @faux-eccles, thanks for your contribution! Next time request a review to me, in this way i'll be notified about your pr.

I tried it on my pc, but I can see in the network tab the ws request is stucked in pending status with you change

faux-eccles:
@drosetti  Ah my bad, the screenshot was of the issue not the fix.    Should have put it only on the issue ticket, and only had the fix in the PR.
screenshot-2025-08-12-21:23:42

drosetti:
Don't worry about the screenshot, it's not a big deal. Even if it's correct the problem remains: in my env the websocket cannot connect from the frontend to the backend. Maybe I have to change ports in the container ?

faux-eccles:
My change should be connecting to the same port that the page is serving from so if you can load the page it should connect.  That being said in my environment the ports are changed in the container, I've shared a screenshot of my setup in the linked issue

drosetti:
@faux-eccles Ok I changed the port on docker and it worked on standard port (80), but there is another problem: the UI environment is on port 3001, if you want to work on frontend this changes makes the websocket fail. I'd like to discuss with you a possible solution for this problem. I prefer to avoid to edit the docker conf because some contributors work quite often on the frontend and change it every time is quite annoying imho.

fgibertoni:
@faux-eccles any updates on this?

faux-eccles:
Changing the port in docker was a local config to run the service in my environment as port 80 is already in use and shouldn't be necessary for the fix itself.   I don't have experience with hosting the UI separately to the backend service for developemnt which is what it sounds like @drosetti  is doing.  

I think it would make sense to then dynamically build/set the port based on launch parameters.  I can take a look to see what looks like.

(also apologies for the slow responses, have just moved countries and in the middle of getting resituated)

faux-eccles:
> @faux-eccles Ok I changed the port on docker and it worked on standard port (80), but there is another problem: the UI environment is on port 3001, if you want to work on frontend this changes makes the websocket fail. I'd like to discuss with you a possible solution for this problem. I prefer to avoid to edit the docker conf because some contributors work quite often on the frontend and change it every time is quite annoying imho.

Looking into this a bit further this seems strange, my assumption is the npm proxy directive in package.json should be forwarding the requests along as with everything else.   So in your environment hosting the react server on 3001, my changes should be connecting to localhost:3000/ws and NPM should forward the requests.  I'll see if I can recreate locally

faux-eccles:
Launching the npm dev server with the default configuration on this branch using the method outlined in [the docs](https://intelowlproject.github.io/docs/IntelOwl/contribute/#frontend) appears to be working as expected (besides some unrelated CORS errors that prevents me from actually submitting jobs to validate messages are coming through as the NPM server appears to just open a websocket connection regardless of the other end being available) 

Do you have the dev react app calling the api server directly and not via the npm proxy?

image

--- 
edit: Fixed the 403 errors by setting the environment variable `stage=local` in `/docker/.env`

faux-eccles:
I've updated the package.json and added a [middleware](https://stackoverflow.com/questions/58088218/websockets-in-create-react-app-with-webpack-proxy) configurations for the react-scripts so that the webpack proxy correctly handles websockets.   This might fix the need for having websockets on different ports while developing.

image

drosetti (changes requested):
Hi @faux-eccles thanks for your work and apologizes for my late response.
Everything work fine, thansk for your changes.

I noticed you use `http-proxy-middleware`. This library is available becase it's installed in the `package-lock.json`, but it's not listed in the dependencies in the `package.json`.

I ask you to add it as dependency in the second file I mentioned or, even better, if you could do the same thing but with the config of Axios in the `initAxios` file. It would be better because we don't have to add another dependency, but I'm not sure if it's possibile, you should check the Axios's doc and do this change in case it's possibile. In case it's too work or it's not supported from Axios the first option is fine 

Edit: please use the pr review feature of github (on the top right), in this way I get a notification via email to review your work or it could be possible I forget to review your code

faux-eccles:
Not sure what you mean about moving the changes into Axios.  This looks likes purely a frontend request client which is behaving normally.  The  proxyConfig I'm introducing should only be loaded by and effect the server side of the npm dev server instructing it how to forward requests

frontend/src/setupProxy.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = function(app) {
+       app.use(
+        createProxyMiddleware('/ws', {
+            target: 'ws://localhost/',
+            ws: true,
+        })
+    );
+   app.use(
+        createProxyMiddleware('/',{
+            target: 'http://localhost',
```
drosetti, line 12: We cannot put "localhost" in the code, if you run IntelOwl in a production environment you will get and error, we should use the real domain

faux-eccles: This will never be used in production as it's only applied during local development via the NPM run command launching the webpack dev server.  It doesn't affect the deployed artifacts

faux-eccles: It's the same as using localhost as the proxy directive in the package.json for testing.

PR #3073, opened 2025-11-17
Fix: HybridAnalysis hash search: switch POST to GET + add overview fallback (#2934)

### LINKED ISSUE #2934: Deprecation of endpoint `POST /search/hash` in Hybrid Analysis API
As of [API v2.35.0](https://hybrid-analysis.com/docs/api/v2-changelog#v2-35-0) Hybrid Analysis started deprecating the POST version of the endpoint `/search/hash` in favor of it's GET version.

We should update our code before the full deprecation of the endpoint in order to avoid having a broken analyzer for hash-type observables.

PranavShukla7:
# Description

Fixes #2934

This PR updates the HybridAnalysis analyzer to support API changes introduced in
API v2.35.0, where the POST /search/hash endpoint was deprecated and replaced by a GET version.

The new GET endpoint returns a minimal response instead of a full SampleSummary.
To maintain backward compatibility, this PR adds a fallback request to
/overview/<sha256> to restore the previous output structure.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue).

# Checklist

- [X] I have read and understood the rules about [how to Contribute](https://intelowlproject.github.io/docs/IntelOwl/contribute/) to this project
    - [X] I strictly followed the documentation ["How to create a Plugin"](https://intelowlproject.github.io/docs/IntelOwl/contribute/#how-to-add-a-new-plugin)
- [X] I have inserted the copyright banner at the start of the file: ```# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl # See the file 'LICENSE' for copying permission.```
- [X] Please avoid adding new libraries as requirements whenever it is possible. Use new libraries only if strictly needed to solve the issue you are working for. In case of doubt, ask a maintainer permission to use a specific library.
- [X] Linters (`Black`, `Flake`, `Isort`) gave 0 errors. If you have correctly installed [pre-commit](https://intelowlproject.github.io/docs/IntelOwl/contribute/#how-to-start-setup-project-and-development-instance), it does these checks and adjustments on your behalf.
- [X] After you had submitted the PR, if `DeepSource`, `Django Doctors` or other third-party linters have triggered any alerts during the CI checks, I have solved those alerts.

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/ha_get.py
+++ b/api_app/analyzers_manager/observable_analyzers/ha_get.py
@@ -21,7 +21,7 @@ class HybridAnalysisGet(ObservableAnalyzer):
 
     @classmethod
     def update(cls) -> bool:
-        return True
+        pass
 
     def _fetch_sample_summary(
         self, sha256: str, headers: Dict[str, str]
@@ -111,16 +111,12 @@ class HybridAnalysisGet(ObservableAnalyzer):
 
         if obs_cls == Classification.DOMAIN:
             response = self._search_terms("domain", value, headers)
-
         elif obs_cls == Classification.IP:
             response = self._search_terms("host", value, headers)
-
         elif obs_cls == Classification.URL:
             response = self._search_terms("url", value, headers)
-
         elif obs_cls == Classification.HASH:
             response = self._search_hash(value, headers)
-
         else:
             raise AnalyzerRunException(
                 f"not supported observable type {obs_cls}. "


PranavShukla7:
Hey @mlodic ,
I’ve updated the analyzer to use the GET version of /search/hash as required by the recent Hybrid Analysis changes, and added a fallback to the overview/{sha256} endpoint to preserve full results when only minimal hashes are returned.

I’ve also refactored the logic into smaller helper functions to make the flow cleaner and easier to maintain.
Could you please review the changes and let me know if anything needs adjustment?

mlodic:
can you please show us the output from this analyzer from the GUI?

api_app/analyzers_manager/observable_analyzers/ha_get.py
```diff
@@ -17,46 +22,148 @@ class HybridAnalysisGet(ObservableAnalyzer):
...
-                sha256 = job.get("sha256", "")
-                job_id = job.get("job_id", "")
-                if sha256:
-                    job["permalink"] = f"{self.sample_url}/{sha256}"
-                    if job_id:
-                        job["permalink"] += f"/{job_id}"
+            self._add_permalink_list(result)
 
         return result
+
```
mlodic: we are not using monkeypatch anymore, you can change the related test here: https://github.com/intelowlproject/IntelOwl/blob/develop/tests/api_app/analyzers_manager/unit_tests/observable_analyzers/test_ha_get.py

api_app/analyzers_manager/observable_analyzers/ha_get.py
```diff
@@ -17,46 +22,148 @@ class HybridAnalysisGet(ObservableAnalyzer):
 
     @classmethod
     def update(cls) -> bool:
-        pass
+        return True
```
fgibertoni: I think this can be left as `pass`. Have you changed it for any particular reason?

fgibertoni (commented):
I do like the general approach of using functions.
If you're performing more than one call to APIs per analyzer run I think it would be better to use a [Session](https://docs.python-requests.org/en/latest/user/advanced/) object, just to make things more clear. Let me know what you think :)

fgibertoni:
Hey @PranavShukla7 any updates?

PranavShukla7:
@fgibertoni yes i need some time please...will do it by the end of the week

fgibertoni:
Sure, no hurries. Just to be sure you're still working on it :smile:

api_app/analyzers_manager/observable_analyzers/ha_get.py
PranavShukla7: I’ve updated the update() method to return True instead of pass.
This matches the expected analyzer interface, where update() should explicitly signal a successful metadata update (even though it’s a no-op for this analyzer).
Let me know if you’d prefer a different behavior here.

api_app/analyzers_manager/observable_analyzers/ha_get.py
PranavShukla7: Thanks for the clarification 
Since we’re no longer using monkeypatch for this analyzer, I can update the related unit test to use standard request mocking instead.

Before I proceed, just wanted to confirm with you: should I completely remove the monkeypatch logic and adapt the test to validate the analyzer behavior using regular mocked HTTP responses?

PranavShukla7:
@fgibertoni that makes sense
I thought about using a Session, but since the analyzer only runs a small, short-lived set of requests per execution, I kept it simple with plain requests for now.

That said, I’m totally open to switching to a Session if you feel it improves clarity or consistency with the rest of the codebase

api_app/analyzers_manager/observable_analyzers/ha_get.py
fgibertoni: I'd prefer to leave it as `pass` as every other analyzer if that's not a problem :smiley:

api_app/analyzers_manager/observable_analyzers/ha_get.py
fgibertoni: Yes that's correct. You can take a look at other existing tests for other analyzers as example.

fgibertoni:
Ok I understand your point, no problem in leaving it as single requests.

api_app/analyzers_manager/observable_analyzers/ha_get.py
PranavShukla7: ok

api_app/analyzers_manager/observable_analyzers/ha_get.py
PranavShukla7: Thanks for the clarification, I will proceed to do it

PR #3127, opened 2025-12-20
Add update checker management command and documentation

### LINKED ISSUE #2876: Add update checker
Add a way to check for updates and notify the users in case.
This should be documented and optionally removable.

lvb05:
This PR adds a basic update check mechanism for IntelOwl.  
It allows to verify whether the running instance is behind the latest published GitHub release.

### Included changes
- Added `api_app/core/update_checker.py` for version fetch + comparison logic
- Added Django management command: `check_updates`
- Added settings keys (`UPDATE_CHECK_URL`, version reuse via existing VERSION)
- Updated README with usage + configuration notes

### How to use

```bash
python manage.py check_updates
```

The command will:
1. fetch the latest GitHub release tag
2. normalize version strings
3. compare against the locally configured version
4. log whether a new version is available

No version upgrade happens automatically.
Fixes #2876
### Notes

- `UPDATE_CHECK_URL` is now environment configurable
- local version comes from existing `VERSION` secret setting
- no new dependencies added
- cron / job scheduler usage referenced in docs if periodic checking is desired.

### Checklist

- [x] I have read and understood the rules about how to contribute to this project.
- [x] The pull request targets the correct branch.
- [x] This PR is linked to an approved issue (#2876).
- [ ] A new plugin (analyzer, connector, visualizer, playbook, pivot, or ingestor) was added or modified.  
  **N/A – this PR does not introduce or modify any plugins.**
- [x] No new external libraries were added.
- [x] Linters (Black, Flake8, Isort) report no errors.
- [ ] Tests were added or updated.  
  **N/A – this PR adds a management command and configuration logic only; existing functionality and tests are unaffected.**
- [x] Documentation was updated where appropriate (README updated with usage and configuration).
- [ ] GUI changes were made.  
  **N/A – backend-only change.**
- [x] I have addressed any CI or linter warnings triggered after submitting the PR.

### DIFF
--- a/frontend/src/components/home/Home.scss
+++ b/frontend/src/components/home/Home.scss
@@ -38,9 +38,20 @@ html {
 }
 
 .system-update-compact {
-  padding: 8px 12px;
+  margin: 12px 0 16px 0;
+  padding: 10px 14px;
+  border-left: 4px solid #ff9800;
+  background: rgba(255, 153, 0, 0.093);
+  color: #f5f5f5;
+  font-size: 0.95rem;
   border-radius: 6px;
-  background-color: rgba(245, 158, 11, 0.15);
-  font-size: 0.9rem;
-  line-height: 1.4;
+  display: flex;
+  flex-wrap: wrap;
+  gap: 6px;
+  align-items: center;
+
+  strong {
+    color: #e7953f;
+    font-weight: 600;
+  }
 }


lvb05:
> Hey @lvb05 thanks for your contribution. Can you please compile the checklist in PR template ? It contains many important step to follow before asking for a review.

I’ve added and completed the PR checklist in the description.
Please let me know if anything needs adjustment. Thanks!

fgibertoni:
I can tell you've not read it carefully since your PR is pointing to the master branch.
Please read it carefully and adjust accordingly.

lvb05:
Hi @fgibertoni,  
I’ve updated the PR to target the develop branch and fixed the formatting issues.  
All checks are green for now.  
Thanks for your time!

api_app/management/commands/check_updates.py
```diff
@@ -0,0 +1,9 @@
+from django.core.management.base import BaseCommand
+from api_app.core.update_checker import check_for_update
+
+class Command(BaseCommand):
+    help = "Check for newer IntelOwl releases"
+
+    def handle(self, *args, **options):
+        check_for_update()
+        self.stdout.write(self.style.SUCCESS("Update check completed"))
```
fgibertoni: This is printed even if the updated was not successful. I think you should check the outcome of the function.

api_app/core/update_checker.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # remove optional 'v' prefix
+        return tag.lstrip("v")
+
+    except requests.RequestException as exc:
+        logger.error(f"update check failed: {exc}")
+        return None
+    except ValueError:
+        logger.error("invalid JSON in update response")
+        return None
```
fgibertoni: As general approach i think it would be better to have separate try...catch block for each error in this particular case.

api_app/core/update_checker.py
```diff
@@ -0,0 +1,96 @@
...
+        logger.warning(
+            f"New IntelOwl version available: {latest_str} "
+            f"(current: {current_version_str})"
+        )
+    elif latest < current:
+        logger.info(
+            f"Local version ahead of release: " f"{current_version_str} > {latest_str}"
+        )
+    else:
+        logger.info(f"IntelOwl version up to date ({current_version_str})")
```
fgibertoni: These outputs aren't printed to the user that is using the command. I think they should be.

fgibertoni (changes requested):
Hey, thanks for your PR! Looking good overall :smiley: 
Please address the pipeline for errors.

README.md
```diff
@@ -47,6 +47,26 @@ It provides:
 We try hard to keep our documentation well written, easy to understand and always updated.
 All info about installation, usage, configuration and contribution can be found [here](https://intelowlproject.github.io/docs/)
 
+
+### Update Checker
```
mlodic: we cannot pollute the ReadMe with this content. Please remove this line and add this instead with a new PR in the Installation section: https://github.com/intelowlproject/docs/blob/main/docs/IntelOwl/installation.md

lvb05:
Hi, I’ve tested the check_updates management command in a local IntelOwl Docker setup. It was executed inside the intelowl_uwsgi container.

Screenshot 2026-01-10 163335

README.md
lvb05: Thanks for highlighting this
I’ve restored the README to its original state. I’ll add the Update Checker documentation under the Installation section in the docs repository (as suggested) via a separate PR.

lvb05:
For this PR, I kept the update checker as a management command only, mainly to introduce and validate the core logic first.
This shouldn’t be something users run manually. The next step would be to run it automatically (via cron or Celery beat) and send a notification through the existing GUI notification system.
Since that involves scheduler decisions and tighter integration with notifications, I think its better with a follow-up PR. Happy to implement that next if thats fine?

mlodic:
We won't merge this PR with these changes only because they are actually incomplete work, by themselves they don't help solving the issue. 

So we can merge to the project only an implementation that contain all the necessary changes

lvb05:
Before implementing the automatic update check, I want to confirm some things.
Should the “update available” notification be treated as a global/admin-level state or something user/org-specific, and should it be persistent until upgrade or just informational when detected? From an integration point of view, should this reuse the existing user_events_manager, or be handled as a separate system-level notification? Finally, for scheduling and delivery, is Celery beat the preferred mechanism(IntelOwl already use this), and is a GUI notification expected, or is a log/admin-level message sufficient?

mlodic:
* the “update available” notification should be treated as a global/admin-level and persistent until upgrade
* It would be great to reuse the user_events_manager to simplify the integration. If this makes impossible to treat the notification ad global (so contradicts the previous point), I don't think it is a problem, it is better to re-use the actual flow than creating complete new integrations for just this case.
* Celery beat is good - A GUI notification would be great. However when you say log/admin level message, what do you mean? Cause that could be an option in case you have a good idea

lvb05:
By log/admin-level message, I mean a complementary signal, not a replacement for the GUI.
The idea would be to always surface the update via a persistent GUI notification, but also log it at INFO/WARNING level when detected. This is useful for headless/API-only deployments and for operational observability (logs/alerts), and also acts as a safe fallback if the GUI event creation fails.
I was thinking to implement the GUI notification as the primary mechanism and adding a structured log message alongside it.

mlodic:
> I was thinking to implement the GUI notification as the primary mechanism and adding a structured log message alongside it.

Fine with this, thanks!

lvb05:
Summary of current status

- This PR adds a persistent backend state for update checks and refactors the logic to be idempotent and admin/global scoped.

- A Celery task is implemented to allow running the update check asynchronously, but it is not scheduled yet.

- GUI notification rendering and Celery Beat scheduling are intentionally deferred, pending confirmation on UX and scheduling expectations.

Once aligned, I’ll add GUI integration, periodic scheduling, and documentation in a follow-up PR.
The full backend logic is implemented.
What remains is how and when it should be executed automatically.

api_app/models.py
```diff
@@ -67,6 +67,44 @@
...
+        blank=True,
+        help_text="Last time the update check was executed",
+    )
+
+    created_at = models.DateTimeField(default=now, editable=False)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        verbose_name = "Update check status"
+        verbose_name_plural = "Update check status"
```
mlodic: these are identical

api_app/models.py
```diff
@@ -67,6 +67,44 @@
 logger = logging.getLogger(__name__)
 
 
+class UpdateCheckStatus(models.Model):
```
mlodic, line 73: there should also be a migration

api_app/models.py
```diff
@@ -67,6 +67,44 @@
...
+    """
+    Stores global state for IntelOwl update checks.
+    This model is intended to be used as a singleton (accessed via get_or_create(pk=1)).
+    Ensures that update notifications are emitted only once per version.
+    """
+
+    latest_version = models.CharField(
+        max_length=50,
```
mlodic: maybe too much?

api_app/core/update_checker.py
```diff
@@ -0,0 +1,129 @@
...
+
+    return tag.lstrip("v"), None
+
+
+def check_for_update():
+    """
+    Compare the running IntelOwl version with the latest available one.
+
+    Returns:
+        (success: bool, message: str)
```
mlodic: use type hints too please

api_app/core/update_checker.py
```diff
@@ -0,0 +1,129 @@
...
+            state.notified = True
+
+        state.save(update_fields=["latest_version", "notified", "last_checked_at"])
+        return (
+            True,
+            f"New IntelOwl version available: {latest_str} "
+            f"(current: {current_version_str})",
+        )
+
+    state.save(update_fields=["last_checked_at"])
```
mlodic: this is a duplicated save if the logic goes inside the previous if.

api_app/core/update_checker.py
```diff
@@ -0,0 +1,129 @@
...
+        if latest_str != current_version_str:
+            return (
+                True,
+                f"Update available: {latest_str} (current: {current_version_str})",
+            )
+        return True, f"IntelOwl version up to date ({current_version_str})"
+
+    if latest > current:
+        if state.latest_version != latest_str or not state.notified:
+            logger.warning(
```
mlodic: info is better

api_app/core/update_checker.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+
+            UserEventQuerySet.notify_admins(
+                title="New IntelOwl version available",
+                message=(
+                    f"Version {latest_str} is available "
+                    f"(current: {current_version_str})"
+                ),
+                persistent=True,
+                severity="warning",
```
mlodic: info

api_app/core/update_checker.py
```diff
@@ -0,0 +1,129 @@
...
+        return None, "Invalid response from update server"
+
+    tag = data.get("tag_name")
+    if not tag:
+        return None, "Release response missing tag_name"
+
+    return tag.lstrip("v"), None
+
+
+def check_for_update():
```
mlodic: before putting this into a cron, we need to add also unittests for all the possible cases here.

api_app/core/update_checker.py
```diff
@@ -0,0 +1,131 @@
...
+
+            Notification.objects.create(
+                title="New IntelOwl Version Available",
+                description=(
+                    f"Version {latest_str} is available "
+                    f"(current: {current_version_str})"
+                ),
+                level="info",
+                for_admins=True,
+            )
```
fgibertoni, line 78: Can this create duplicate notifications if the command in ran twice ? You can check if the version has already been notified.

intel_owl/tasks.py
fgibertoni, line 1: What happened on this file? Unclean branch maybe?

mlodic:
I'll review as soon as I can

mlodic:
there are still conflicts and, as mentioned, this requires GUI changes otherwise it won't make a lot of sense adding this as a feature

lvb05:
> there are still conflicts and, as mentioned, this requires GUI changes otherwise it won't make a lot of sense adding this as a feature

Yep I'll add a frontend system update panel and an API endpoint  '/system/update-check/ ' to expose update status. Let me know if you’d like any specific fields or UI adjustments.
Thanks

lvb05:
Hi!
All CI checks are passing now. I’ve addressed the requested changes.
I will open a separate PR for documentation updates.
Let me know if any further adjustments are needed.

api_app/models.py
```diff
@@ -70,6 +70,42 @@
 logger = logging.getLogger(__name__)
 
 
+# UpdateCheckStatus (replace the existing class in api_app/models.py)
```
mlodic: I don't understand this comment, can you please explain?

api_app/models.py
mlodic: this has not been addressed

frontend/src/components/common/SystemUpdatePanel.jsx
```diff
@@ -0,0 +1,82 @@
...
+        <span>{data.latest_version || "Unknown"}</span>
+      </div>
+
+      <div style={styles.row}>
+        <strong>Last Checked:</strong>
+        <span>{lastChecked}</span>
+      </div>
+
+      {data.update_available ? (
+        <div style={styles.updateBox}>A new system update is available!</div>
```
mlodic: please show GUI screenshots/video fpr both cases

mlodic (commented):
we need a showcase about how this actually works to be able to merge this, otherwise we don't know the outcome

lvb05:
Hi!
I’ve attached
- Authenticated API responses for both cases (update available / system up to date)
- Dashboard screenshots showing the UI behavior in both scenarios

The update checker persists its state via the UpdateCheckStatus model (added in migration 0072_update_check_system.py).
Plus, I already added the migrations for that in 0072 in that pr. 

### Proof-

Screenshot 2026-02-06 201116
Screenshot 2026-02-06 194838
Screenshot 2026-02-06 201516
Screenshot 2026-02-06 201500
Screenshot 2026-02-06 201333
Screenshot 2026-02-06 201321

mlodic:
thanks for showcasing it, it really helps. The change is fine, just one thing: I think this should be shown in the Home page, just before the "IntelOWl News" section and, in case we have the most recent version, it should not render anything because it would uselessly clutter the page.
In this way the user would get the message in the Home without cluttering the Dashboards. Thoughts?

lvb05:
> thanks for showcasing it, it really helps. The change is fine, just one thing: I think this should be shown in the Home page, just before the "IntelOWl News" section and, in case we have the most recent version, it should not render anything because it would uselessly clutter the page. In this way the user would get the message in the Home without cluttering the Dashboards. Thoughts?

That makes sense. To keep things simple and avoid duplication, I can move the entire System Updates component to the Home page and render it only when an update is available.
This way, the Home page stays clean when the system is up to date, and we don’t need separate “summary” and “detail” components. I can implement this.

lvb05:
image

Hi! Done with all the refinements. Check once @mlodic @fgibertoni

mlodic:
that's nice, thanks

lvb05:
Thanks for the merge! @mlodic @fgibertoni 
I also wanted to check if you’re okay with me opening a separate PR to update the documentation related to this change. Please let me know if there’s anything specific you’d like covered there.

mlodic:
feel free to open a PR about it

PR #3141, opened 2025-12-31
 add APIVoid v2 API support with v1 fallback

Rohankaf:
### Description

This PR updates the existing APIVoid analyzer to support the new APIVoid v2 API, while keeping backward compatibility with the v1 API.

The analyzer automatically falls back to v1 if v2 is not configured, so existing setups will continue to work without any changes.

While working on this, I also addressed a CodeQL security warning by sanitizing analyzer logging to avoid logging externally derived or potentially sensitive data.

No new dependencies were added, and the overall behavior of the analyzer remains unchanged for users already using APIVoid v1.

Closes #3083


mlodic:
you need to follow the official guidelines

PR #3224, opened 2026-01-20
Fix phishstats url

### LINKED ISSUE #3152: Update phishstats base URL
The new URL is https://api.phishstats.info/api/phishing. Now we are using "https://phishstats.info:2096/api" which fails.

For more info see: https://phishstats.info/api-docs. 

A new migration is required for this.

nehamanoj1105:
Fixes #3152

- Updated PhishStats analyzer to use the new official API endpoint
- Added a data migration to update existing analyzer configurations

Note: migration not executed locally due to environment constraints.


mlodic:
follow the contribution gudelines please

PR #3242, opened 2026-01-27
Fix job stuck in RUNNING state when pipeline setup fails

Yashika0724:
### Summary

When job.execute() raises an exception during pipeline setup, related reports are marked as FAILED, but the Job itself remains in RUNNING state, finished_analysis_time is not set, and no WebSocket notification is sent. This causes silent stuck jobs in production when failures occur during task dispatch (e.g., broker issues, invalid queues, analyzer setup errors).
This PR updates the Job status to FAILED, records the completion time, and notifies connected clients in the exception handler, ensuring the job lifecycle is consistent even when pipeline dispatch fails.

---

### Impact
	•	Prevents jobs from remaining indefinitely in RUNNING state
	•	Ensures completion timestamps are always set on failure
	•	UI and API clients receive immediate failure notification
	•	Keeps existing success-path behavior unchanged

### DIFF
--- a/intel_owl/tasks.py
+++ b/intel_owl/tasks.py
@@ -256,6 +256,7 @@ def job_pipeline(
     job_id: int,
 ):
     from api_app.models import Job
+    from api_app.websocket import JobConsumer
 
     job = Job.objects.get(pk=job_id)
     try:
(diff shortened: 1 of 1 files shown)


Yashika0724:
Hi @mlodic , this PR updates the Job status when pipeline setup fails, to avoid jobs remaining in RUNNING state when no tasks are dispatched. Happy to make any changes if needed, thanks!

mlodic:
read and follow the guidelines, the PR template is also meant to be compiled, not deleted

PR #3270, opened 2026-02-03
Implemented Data Atomicity

### LINKED ISSUE #3225: Job.data_model may reference deleted BaseDataModel on save failure
## Location

**File:** `api_app/engines_manager/models.py`  
**Method:** `EngineConfig.run()`

## Description

In `EngineConfig.run()`, when a job already has a data model, the existing data model is deleted and a new one is assigned. These operations are not atomic. If `job.save()` fails after the deletion, the job's GenericForeignKey fields (`data_model_content_type`, `data_model_object_id`) can end up referencing a deleted `BaseDataModel`.

This leaves the job in a partially updated and inconsistent state.

## Execution Flow

Current order of operations:

1. Existing `job.data_model` is deleted from the database
2. New `BaseDataModel` is assigned to `job.data_model` in memory
3. `job.save()` is called to persist the change

If step 3 fails, the previous deletion is not rolled back.

## Trigger Scenarios

`job.save()` can fail due to normal conditions such as:

- Validation errors
- Signal exceptions
- Database constraint errors
- Connection or transaction failures

## Observed Invalid State

After a failed save:

- The old `BaseDataModel` is already deleted
- The database still contains FK values pointing to the deleted object
- The new `BaseDataModel` exists but is not linked to the job
- Accessing `job.data_model` raises `DoesNotExist` or returns `None`

This results in an inconsistent GenericForeignKey relationship.

## Impact

If this occurs, it can lead to:

- Runtime errors when accessing `job.data_model`
- Engines, pivots, or connectors operating on inconsistent job state
- Queries involving `job.data_model` returning incorrect results
- Orphaned `BaseDataModel` records accumulating over time
- Difficult debugging due to silent data corruption

## Expected Safe Behavior

The update should be atomic, so that either:

- The old data model remains until the new one is safely saved, or
- Partial state is prevented if the operation fails

This ensures `job.data_model` always references a valid object.

## Notes

This issue is about atomicity and data integrity, not about changing intended behavior. The exact solution approach is open for discussion. Scope is limited to `EngineConfig.run()`.

AnshSinghal:
Closes #3225 

# Description

Implemened data atomicity This ensures that if job.save() fails for any reason (validation errors, signal exceptions, database constraints, etc.), the entire transaction is rolled back

## Type of change

### DIFF
--- a/api_app/engines_manager/models.py
+++ b/api_app/engines_manager/models.py
@@ -45,10 +45,11 @@ class EngineConfig(SingletonModel):
 
         with transaction.atomic():
             data_model_result: BaseDataModel = job.get_analyzers_data_models().merge(append=True)
-            if job.data_model:
-                job.data_model.delete()
+            previous_data_model = job.data_model
             job.data_model = data_model_result
             job.save()
+            if previous_data_model:
+                previous_data_model.delete()
 
         runner = group(list(self.get_modules_signatures(job)))
         runner.apply_async(


AnshSinghal:
@mlodic can we merge?

api_app/engines_manager/models.py
```diff
@@ -35,16 +35,20 @@ def get_modules_signatures(self, job) -> Generator[Signature, None, None]:
...
         if job.analyzable.classification == Classification.GENERIC:
             # at the moment, since there are no datamodels for the generic, we are completely skipping an evaluation
             return
-        data_model_result: BaseDataModel = job.get_analyzers_data_models().merge(append=True)
-        if job.data_model:
-            job.data_model.delete()
-        job.data_model = data_model_result
-        job.save()
+
+        with transaction.atomic():
```
mlodic, line 46: please add a test which actually replicates the reported issue, otherwise there's no point about adding this

AnshSinghal:
@mlodic please review

mlodic (approved):

PR #3285, opened 2026-02-07
fix: add HTTP timeout to Yeti connector (Fixes #3256)

### LINKED ISSUE #3256: Yeti connector missing HTTP timeout can block workers
# Yeti connector missing HTTP timeout can block workers

## Summary

The Yeti connector makes an HTTP request without setting a timeout.
If the Yeti API is slow or not responding, this request can block a Celery worker for a long time. This reduces worker availability, creates queue backlog, and affects overall system throughput.

This is a reliability issue, not a feature change.

## Affected Component

**File:** `api_app/connectors_manager/connectors/yeti.py`  
**Method:** `run()`  
**Lines:** 56–61

## Problem Description

In the Yeti connector, the HTTP request is made like this:

```python
resp = requests.post(
    url=url,
    headers=headers,
    json=payload,
    verify=self.verify_ssl,
)
```

No timeout is provided.

By default, `requests.post()` can wait forever if the remote API does not respond. In this case:

- The connector runs inside a Celery worker
- A blocked HTTP call blocks the worker
- The task stays in `RUNNING` state
- The worker cannot process other jobs

Celery has global time limits, but they are very long (30+ minutes), which is not suitable for a simple API call.

### Worker Blocking

If the Yeti API is slow, down, or the network is unstable:

- The worker stays blocked waiting for response
- No useful work is done during this time

### Resource Exhaustion

If multiple jobs hit the same condition:

- Worker pool capacity drops
- Other jobs queue up
- System throughput degrades

### Poor Visibility

From an operator point of view:

- Jobs look "stuck" instead of clearly failed
- Debugging becomes harder
- Manual intervention may be needed

## Impact

**Severity:** Medium to High

Possible effects in production:

- Celery workers blocked for long periods
- Backlog of jobs
- Slow or stalled analysis pipelines
- External API instability affecting IntelOwl runtime

This type of failure usually appears only during external outages, making it harder to detect and more expensive to handle operationally.

## Evidence from the Codebase

Other parts of the project already handle this correctly:

### MISP Connector

**File:** `api_app/connectors_manager/connectors/misp.py`  
**Lines:** 102–108

```python
misp_instance = pymisp.PyMISP(
    url=self._url_key_name,
    key=self._api_key_name,
    ssl=ssl_param,
    debug=self.debug,
    timeout=5,
)
```

### HIBP Analyzer

**File:** `api_app/analyzers_manager/observable_analyzers/hibp_utils.py`  
**Lines:** 24–46

```python
def make_hibp_request(
    url: str,
    params: dict = None,
    api_key: str = None,
    timeout: int = 10,
) -> dict | list | str:
    """
    Make a GET request to HIBP API and handle common responses.
    :param timeout: Request timeout in seconds.
    """
    try:
        response = requests.get(
            url,
            params=params or {},
            headers=get_headers(api_key),
            timeout=timeout,
            verify=True,
        )
```

### OTX Analyzer

**File:** `api_app/analyzers_manager/observable_analyzers/otx.py`  
**Lines:** 19–48

```python
class OTXv2Extended(OTXv2.OTXv2):
    """
    This is to add "timeout" feature without having to do a fork
    Once this PR is merged: https://github.com/AlienVault-OTX/OTX-Python-SDK/pull/66
    we can remove this and use the upstream
    """
    
    def __init__(self, *args, timeout=None, **kwargs):
        super().__init__(*args, **kwargs)
        self.timeout = timeout
    
    def get(self, url, **kwargs):
        try:
            response = self.session().get(
                self.create_url(url, **kwargs),
                headers=self.headers,
                proxies=self.proxies,
                verify=self.verify,
                cert=self.cert,
                timeout=self.timeout,
            )
```

The OTX analyzer even includes comments explaining why timeouts are necessary: *"this was needed because, otherwise, the analyzer could last too much time and become the bottleneck of all the application"* (line 32).

The Yeti connector is currently the only connector doing an HTTP request without a timeout.

## Expected Behavior

All external HTTP calls should have a bounded execution time.

If the Yeti API does not respond within a reasonable time:

- The request should fail
- The connector should report failure
- The worker should be released

Failing fast is safer than blocking indefinitely.

## Proposed Fix

Add a timeout to the HTTP request:

```python
resp = requests.post(
    url=url,
    headers=headers,
    json=payload,
    verify=self.verify_ssl,
    timeout=60,
)
```

### Why 60 seconds

- Enough for normal API responses
- Fails clearly if the API is unhealthy
- Consistent with other parts of the codebase
- Prevents long worker blocking
- Does not prevent future adjustment if operational data suggests a different value

## Scope and Risk

**Change size:** Very small (1 line)

**Risk:** Low

- **Behavior change:** Only on failure paths
- **Success path:** No change

This improves reliability without affecting normal operations.

## Priority Justification

Even though Celery has global time limits, waiting 30+ minutes for a single HTTP request is not practical in production.

This is a small fix with high operational value and aligns with existing project patterns.

## Proposed Implementation

I can open a small PR to add an explicit timeout to the HTTP request.

For consistency, I was thinking of using `timeout=60` seconds:

- MISP connector already uses `timeout=5`
- HIBP analyzer uses `timeout=10`
- OTX analyzer has explicit timeout handling due to blocking issues
- Yeti payloads and responses can be heavier and slower in practice

Does `timeout=60` sound reasonable here? If not, I'm happy to adjust before opening the PR.

---

gks281263:
Adds `timeout=60` to the `requests.post()` call in the Yeti connector so that slow or unresponsive Yeti API does not block Celery workers indefinitely.

Fixes #3256

### DIFF
--- a/api_app/connectors_manager/connectors/yeti.py
+++ b/api_app/connectors_manager/connectors/yeti.py
@@ -58,6 +58,7 @@ class YETI(classes.Connector):
                 headers=headers,
                 json=payload,
                 verify=self.verify_ssl,
+                timeout=60,
             )
             resp.raise_for_status()
         except requests.RequestException as e:


gks281263:
Hi @mlodic , this PR is still relevant and ready from my side. Please let me know if any changes are needed. Happy to adjust.

PR #3331, opened 2026-02-19
Optimized Decay() Method

### LINKED ISSUE #3248: Optimize UserEventQuerySet.decay() method

AnshSinghal:
Closes #3248 

# Description

Before: Broken code with incomplete if-block (SyntaxError) and old N+1 implementation using individual .save() calls (2N+2 queries for N events, no transaction atomicity).

After: Clean single-pass implementation with:

defaultdict(list) to group data_models by class for efficient bulk updates
Single loop that mutates objects in memory (increment decay_times, decrement reliability, calculate next_decay)
transaction.atomic() wrapping all database writes for atomicity
bulk_update() for both data models (grouped by class) and events (2-4 queries total vs 2N+2)
len(events) return instead of objects.count() (avoids extra query)

## Type of change

### DIFF
--- a/api_app/user_events_manager/queryset.py
+++ b/api_app/user_events_manager/queryset.py
@@ -22,16 +22,20 @@ class UserEventQuerySet(QuerySet):
                 next_decay__lte=now(),
             )
         )
+        # ForeignKey appears in _meta.fields (wildcard models) -> use JOIN.
+        # GenericForeignKey does not (UserAnalyzableEvent) -> use prefetch.
         model_fields = {field.name for field in self.model._meta.fields}
         if "data_model" in model_fields:
             objects = objects.select_related("data_model")
         else:
             objects = objects.prefetch_related("data_model")
 
+        # Load into memory so we can mutate fields and bulk-write back.
         events = list(objects)
         if not events:
             return 0
 
+        # Group by concrete class since bulk_update works per-table.
         data_models_by_class = defaultdict(list)
 
         for obj in events:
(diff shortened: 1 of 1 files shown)


AnshSinghal:
@mlodic please review. Screenshot is not available as its a performance optimization. I have included relevant tests.

api_app/user_events_manager/queryset.py
```diff
@@ -20,12 +22,26 @@ def decay(self):
...
             )
         )
-        # TODO we can probably translate all of this in sql query
-        for obj in objects:
-            obj: UserEvent
+        model_fields = {field.name for field in self.model._meta.fields}
+        if "data_model" in model_fields:
+            objects = objects.select_related("data_model")
+        else:
+            objects = objects.prefetch_related("data_model")
```
mlodic, line 31: could you please  explain these 5 lines?

mlodic: also add comments in the code about please

mlodic (commented):
broadly speaking, this seems fine but please add comments in all the changed parts in the code that explain the reasoning behind this change. 
you know, this may optimize the performance but it add complexity and this is not always helpful

PR #3339, opened 2026-02-20
fix: optimize decay method to eliminate N+1 query. Closes #3316

### LINKED ISSUE #3316: Optimization: Avoid N+1 queries in UserEventQuerySet.decay
## Description
The `decay` method in `api_app/user_events_manager/queryset.py` performs an N+1 query operation by iterating over a queryset and calling `.save()` inside the loop. This can be inefficient as the number of user events increases.

## Current Implementation

The current implementation iterates over every user event that needs decay, performs arithmetic calculations in Python, and saves each object individually:

```python
# api_app/user_events_manager/queryset.py

    def decay(self):
        # ...
        objects = (
            self.exclude(decay_progression=DecayProgressionEnum.FIXED.value)
            .exclude(next_decay__isnull=True)
            .filter(
                next_decay__lte=now(),
            )
        )
        # TODO we can probably translate all of this in sql query
        for obj in objects:
            obj: UserEvent
            obj.decay_times += 1
            obj.data_model.reliability -= 1
            if obj.data_model.reliability == 0:
                obj.next_decay = None
            else:
                if obj.decay_progression == DecayProgressionEnum.LINEAR.value:
                    obj.next_decay += datetime.timedelta(days=obj.decay_timedelta_days)
                elif obj.decay_progression == DecayProgressionEnum.INVERSE_EXPONENTIAL.value:
                    obj.next_decay += datetime.timedelta(
                        days=obj.decay_timedelta_days ** (obj.decay_times + 1)
                    )
            obj.data_model.save()
            obj.save()
        return objects.count()
```

## Proposed Solution
Refactor this method to use Django's `update()` method combined with `F()` expressions and `Case`/`When` statements to perform the update in a single database query.

This will involve:
1. Using `F('decay_times') + 1` for incrementing `decay_times`.
2. Using `Case`/`When` to handle the conditional logic for `next_decay` calculation based on `decay_progression` (LINEAR vs INVERSE_EXPONENTIAL) and `reliability == 0`.
3. The `data_model` field is a `GenericForeignKey`, so its `reliability` cannot be updated via `F()` expressions on the `UserEvent` queryset. A **separate** bulk update on the related data model table(s) will be needed.
4. The exponentiation in `INVERSE_EXPONENTIAL` mode (`decay_timedelta_days ** (decay_times + 1)`) may require `django.db.models.functions.Power()` or a database function to express in SQL.

Please let me know if this approach works fine.

AdityaGupta716:
Refactored the decay method in UserEventQuerySet to eliminate N+1 database queries.
Previously, the method iterated over each UserEvent and called .save() individually, resulting in O(N) queries. This PR replaces that with bulk database operations:

Used bulk_update() to write decay_times and next_decay in a single query
Grouped data_model objects by concrete class and bulk-updated reliability per table
Used select_related / prefetch_related based on field type to efficiently load related data
Wrapped all writes in transaction.atomic() for consistency

Related issue: #3316
Type of change: Bug fix


AdityaGupta716:
Hi @mlodic, I noticed the bulk_update() block in the merged code is unreachable due to return objects.count() appearing before it — the N+1 issue is still present. I've updated my PR #3339 with a working fix that also cleanly separates FK and GFK handling

AnshSinghal:
> Hi @mlodic, I noticed the bulk_update() block in the merged code is unreachable due to return objects.count() appearing before it — the N+1 issue is still present. I've updated my PR #3339 with a working fix that also cleanly separates FK and GFK handling

Thanks for pointing this out. You’re right, I missed the early return and the per-object save() calls make the bulk update path unreachable. I just realized this and will fix it in my PR.

Also, a gentle request to please wait until an issue is assigned before working on it, to avoid duplicate effort and to respect maintainers and ongoing contributions in the open-source workflow.

mlodic:
this is not needed and already addressed

mlodic:
@AnshSinghal wait a second, which PR are you mentioning? You are creating some chaos here, I am not understanding what is actually happening. I'll close this PR in any case because it contains too many unrelated commits but I'd like to continue the conversation in the issue

PR #3378, opened 2026-02-26
fix: handle None rate_limit_timeout in disable_for_rate_limit(). Clos…

### LINKED ISSUE #3366: Bug: disable_for_rate_limit() crashes with TypeError when rate_limit_timeout is not set
## What happened
While going through api_app/models.py, I found that OrganizationPluginConfiguration.disable_for_rate_limit() crashes with a TypeError whenever rate_limit_timeout is not configured for a plugin.

The method does this on line 1047:
```python
enabled_to = now() + self.rate_limit_timeout
```
But rate_limit_timeout is defined as null=True, blank=True in the model, meaning it is completely optional and can be None. When it is None, Python tries to add a datetime and None together — which immediately blows up.

What makes this worse is that the crash happens after self.disabled = True has already been set, but before the re-enable task is ever scheduled. So the plugin ends up permanently disabled with no automated way to recover — and nothing in the logs to explain why.

## Environment
1. OS: Any
2. IntelOwl version:All current versions — bug is present in master at [api_app/models.py#L1047](https://github.com/intelowlproject/IntelOwl/blob/master/api_app/models.py#L1047)

If rate_limit_timeout is not set, the method should handle it gracefully — disable the plugin, leave a helpful comment explaining there is no re-enable timeout configured, log a warning so the operator knows, and return cleanly without crashing.

The re-enable scheduling logic should only run when rate_limit_timeout actually has a value.

1. Create an OrganizationPluginConfiguration for any plugin — leave rate_limit_timeout empty.
2. Trigger a rate-limit error on that plugin (or call org_config.disable_for_rate_limit() directly from a Django shell).
3. The method crashes with a TypeError and the plugin is left permanently disabled.

## Error messages and logs
```
TypeError: unsupported operand type(s) for +: 'datetime.datetime' and 'NoneType'

  File "api_app/models.py", line 1047, in disable_for_rate_limit
    enabled_to = now() + self.rate_limit_timeout
```
No warning or log message is printed to explain what went wrong or that rate_limit_timeout was missing.

@mlodic
I’d be happy to work on a fix if this is confirmed as valid. Please let me know if I can proceed to fix this issue and would love to contribute here

HarshitPal25:
# Description

`OrganizationPluginConfiguration.disable_for_rate_limit()` crashes with a `TypeError`
when `rate_limit_timeout` is `None` (not configured). This PR adds a guard to handle
that case gracefully.

Closes #3366 

## Type of change

# Checklist

### DIFF
--- a/api_app/models.py
+++ b/api_app/models.py
@@ -1075,20 +1075,14 @@ class OrganizationPluginConfiguration(models.Model):
         """
         Disables the configuration for the organization due to rate limits.
         """
-        self.disabled = True
-
         if not self.rate_limit_timeout:
             logger.warning(
                 f"{self} hit a rate limit but rate_limit_timeout isn't configured. "
-                "Disabling the plugin, but no re-enable task will be scheduled."
+                "Skipping disable — please set a rate_limit_timeout."
             )
-            self.disabled_comment = (
-                f"Rate limit hit at {now().strftime('%Y-%m-%d %H:%M:%S')}.\n"
-                "No re-enable timeout was configured — please re-enable manually."
-            )
-            self.save()
             return
 
+        self.disabled = True
         enabled_to = now() + self.rate_limit_timeout
         self.disabled_comment = (
             "Rate limit hit at "
--- a/tests/api_app/test_models.py
+++ b/tests/api_app/test_models.py
@@ -76,10 +76,8 @@ class OrganizationPluginConfigurationTestCase(CustomTestCase):
         self.assertIsNone(obj.rate_limit_enable_task)
         obj.disable_for_rate_limit()
         obj.refresh_from_db()
-        self.assertTrue(obj.disabled)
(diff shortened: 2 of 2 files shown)


HarshitPal25:
PTAL @mlodic

api_app/models.py
```diff
@@ -1077,12 +1077,24 @@ def disable_for_rate_limit(self):
         """
         self.disabled = True
 
+        if not self.rate_limit_timeout:
+            logger.warning(
+                f"{self} hit a rate limit but rate_limit_timeout isn't configured. "
+                "Disabling the plugin, but no re-enable task will be scheduled."
+            )
+            self.disabled_comment = (
```
mlodic: I wouldn't add this behavior, this hides a missed configuration, just keep the warning only here and do not disable the plugin

mlodic:
closing for inactivity and because not needed

PR #3383, opened 2026-02-27
fix: correct boolean comparison for is_sample in recent_scans_user 

### LINKED ISSUE #3382: [Bug] Recent Scans panel does not show file scan jobs
https://github.com/user-attachments/assets/b417bff9-3909-431b-a5d9-b854b61f287c

## What happened
- the "Recent Scans" panel on the frontend `/scan` page always displays recent observable jobs instead of recent file jobs
- this happens because the frontend sends a boolean True and it is compared with a string "True" which returns false so
- line : 436 in `api_app/views.py:`
`        if is_sample == "True":`

## Environment
1. OS: Any
2. IntelOwl version: 6.5.1

On clicking the file scan option recent file scan jobs should be visible 
required edit - change line 436 of the above file to 
`        if is_sample is True:`
- add a file to scan then click on the file scan option it wont show in recent scans
## Error messages and logs
- no error messages

sanjib2006:
# Description
Closes #3382
Recent Scans only observable jobs instead of file jobs because a boolean True is compared with the string "True" in `api_app/views.py (line 436)`.
Change:
Updated:
`if is_sample == "True":`
to:
`if is_sample is True:`

## Type of change

# Checklist

### DIFF
--- a/api_app/views.py
+++ b/api_app/views.py
@@ -435,7 +435,7 @@ class JobViewSet(ReadAndDeleteOnlyViewSet, SerializerActionMixin):
             raise ValidationError({"detail": "is_sample is required"})
         is_sample = request.data["is_sample"]
         jobs = Job.objects.filter(user__pk=request.user.pk)
-        if is_sample == "True":
+        if is_sample is True:
             jobs = jobs.filter(analyzable__classification=Classification.FILE)
         else:
             jobs = jobs.exclude(analyzable__classification=Classification.FILE)


sanjib2006:
https://github.com/user-attachments/assets/963f0b98-312c-416f-b963-356f81d583b1

sanjib2006:
@mlodic I have fixed this bug, could you please review it? : )

PR #3391, opened 2026-02-28
fix: replace custom regex password validation with Django's built-in validators

### LINKED ISSUE #3385: Bug: REGEX_PASSWORD allows weak passwords
###  What happened
I was going through the codebase and noticed that the password validation regex in intel_owl/consts.py is way too permissive. Here's what it currently looks like:
```
# intel_owl/consts.py
REGEX_PASSWORD = r"^[a-zA-Z0-9]{12,}$"
```
This regex only checks two things:

The password is at least 12 characters long
It only contains letters and numbers
That's it. No uppercase requirement, no lowercase requirement, no digit requirement, no special characters. And because of the [a-zA-Z0-9] character class, it actually blocks special characters entirely — so even users who want to use a stronger password with symbols like !, @, or # can't.

and the issue is regex is not just being in the code but it's wired directly into the user registration flow in authentication/serializers.py:
```
# /authentication/serializers.py
# valid password defination
def validate_password(self, password):
    super().validate_password(password)

    if re.match(REGEX_PASSWORD, password):
        return password
    else:
        raise ValidationError("Invalid password")
```
IntelOwl is a threat intelligence platform where analysts store API keys, configure external service integrations, and handle sensitive investigation data. Weak passwords on accounts like these are a real problem.

### Environment
1. OS: Any
2. IntelOwl version: Latest master

### What did you expect to happen
The password validation should actually enforce some complexity — at minimum, requiring a mix of uppercase, lowercase, digits, and ideally at least one special character. A password like aaaaaaaaaaaa should be rejected, not accepted.

Also, the current regex blocks special characters entirely. A user trying to set **pasword@123#** would be told their password is invalid, which is backwards — the stronger password gets rejected while the weaker one sails through.

### How to reproduce your issue
1. Register a new account on IntelOwl.
2. Try setting the password to aaaaaaaaaaaa (12 identical lowercase letters).
3. It passes validation and the account is created successfully.
4. Now create another accound use hard password with special character it says invalid.

### Suggested fix:

Update the regex to require at least one character from each class, and allow (not block) special characters

@mlodic assign this issue i have created a solution on my own too i just want to implement it assign it so i can create a pr.

HarshitPal25:
## Description

Fixes #3385

### Problem
The custom `REGEX_PASSWORD` (`r"^[a-zA-Z0-9]{12,}$"`) and `validate_password_strength()` in `consts.py` are too permissive — they accept passwords like `aaaaaaaaaaaa` while blocking special characters entirely.

### Solution
Replace the custom regex-based validation with Django's built-in `django.contrib.auth.password_validation.validate_password`, which leverages the `AUTH_PASSWORD_VALIDATORS` already configured in `intel_owl/settings/auth.py`:

- `UserAttributeSimilarityValidator` — rejects passwords too similar to user attributes
- `MinimumLengthValidator` — enforces minimum length (default 8)
- `CommonPasswordValidator` — rejects common passwords
- `NumericPasswordValidator` — rejects entirely numeric passwords

### Changes

- **`intel_owl/consts.py`**: Removed `REGEX_PASSWORD`, `validate_password_strength()`, and their `re`/`ValidationError` imports.
- **`authentication/serializers.py`**: Replaced `validate_password_strength()` with `django_validate_password()` (aliased to avoid name collision with the serializer method). Builds an unsaved `User` instance from registration data so `UserAttributeSimilarityValidator` can check against username/email/name fields.
- **`authentication/views.py`**: Replaced `validate_password_strength()` with `django_validate_password(password, user)` in `ChangePasswordView`, passing the authenticated user for similarity checks. Added null checks for missing `old_password`/`new_password`. Fixed `e.message` → `e.messages` for proper error serialization.
- **`tests/auth/test_auth.py`**: Updated tests to match new behavior — passwords with special characters are now accepted (tests changed from 400 → 200), weak password tests assert on `"error"` key presence instead of the old `"Invalid password"` string, and short password test uses a truly short password (`"short1"`).

### Checklist
- [x] PR targets `develop` branch
- [x] Single squashed commit
- [x] Passes ruff lint and format checks
- [x] Tests updated to reflect new validation behavior

### DIFF
--- a/authentication/serializers.py
+++ b/authentication/serializers.py
@@ -207,15 +207,7 @@ class RegistrationSerializer(rest_email_auth.serializers.RegistrationSerializer)
             ValidationError: If the password does not meet the validation criteria.
         """
         super().validate_password(password)
-        user_data = {}
-        if hasattr(self, "initial_data"):
-            for attr in ("username", "email", "first_name", "last_name"):
-                value = self.initial_data.get(attr)
-                if value:
-                    user_data[attr] = value
-        user = User(**user_data) if user_data else None
-
-        django_validate_password(password, user=user)
+        django_validate_password(password)
         return password
 
     def create(self, validated_data):


mlodic:
updates on this?

HarshitPal25:
@mlodic hey yes like i have replaced the regex validation with the django and i have checked and it passes all the tests and i reviewd what the copilot suggest but it gave a an error while testing backend tests you check out the code i guess and if you want me to include the copilot suggestions i would definitely  include them

HarshitPal25:
@mlodic yes I am willing to work on it just didn't get any respond so I was not doing anything if you like suggestions given by copilot i would implement them.

mlodic:
it's not that "I like them", it is a matter of what's right and what's not...copilot is not added here for sport, it's here to help us in creating better code, you can't just ignore it, that should be obvious

PR #3430, opened 2026-03-05
fix(analyzers): correctly pass published=False in PyMISP search (#3429)

### LINKED ISSUE #3429: Missing `False` Boolean Handling for `published` Parameter in PyMISP Search
## Description

In the `MISP` observable analyzer (`api_app/analyzers_manager/observable_analyzers/misp.py`), the `published` parameter configuration incorrectly handles boolean `False` values.

The application currently checks the parameter using:

```python
if self.published:
```

Since `self.published` defaults to `False` (a boolean type), evaluating it directly in an `if` statement treats it as a falsy value. As a result, when a user explicitly wants to filter MISP events where `published=False`, the parameter is omitted entirely from the `pymisp.search()` arguments.

A `FIXME` comment in the code already acknowledges this oversight. It suggests that the default configuration value should be updated to `None` instead of `False`, and the condition should explicitly check whether the parameter was set.

---

## Affected Component

**File:** `api_app/analyzers_manager/observable_analyzers/misp.py`  
**Lines:** 57–62

---

## Current Code

```python
# https://pymisp.readthedocs.io/en/latest/modules.html#pymisp.PyMISP
# fixme: this should be None as default but is False
# so it's not possible to set it as False in this way.
# migration required
if self.published:
    params["published"] = self.published
```

---

## Problem

With the current logic:

| Value of `self.published` | Behavior |
|---|---|
| `True` | Parameter passed correctly |
| `False` | Parameter omitted entirely |
| `None` | Not possible due to default |

This prevents users from explicitly querying MISP events where `published=False`.

---

## Expected Behavior

The analyzer should allow three states:

| Value | Meaning |
|---|---|
| `True` | Return only published events |
| `False` | Return only unpublished events |
| `None` | Do not filter by published status |

---

## Suggested Fix

1. Change the default value of `published` from `False` to `None`.
2. Update the conditional check to explicitly test for `None`.

Example fix:

```python
if self.published is not None:
    params["published"] = self.published
```

---

## Impact

- Enables correct filtering of unpublished MISP events
- Aligns with expected PyMISP search API behavior
- Resolves the existing `FIXME` noted in the code

---

## Labels

`bug`

tejassinghbhati:
## Description

This PR fixes a bug in the MISP observable analyzer where the `published` parameter configuration incorrectly dropped boolean `False` values from the `pymisp.search()` arguments.

Previously, the analyzer checked:

```python
if self.published:
```

This evaluates to falsy when the user configures `published=False`. As a result, the `published` filter was completely omitted from the PyMISP query.

This change updates the `published` field default to `None` and modifies the conditional check to:

```python
if self.published is not None:
```

This ensures that `False` values are correctly passed to the API wrapper as intended by the user.

Closes #3429.

---

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

---

## Checklist

- [x] A new plugin (analyzer, connector, visualizer, playbook, pivot or ingestor) was added or changed, in which case: *(N/A — Existing plugin bug fix)*  
- [x] I have inserted the copyright banner at the start of the file:  

```
# This file is a part of IntelOwl
# https://github.com/intelowlproject/IntelOwl
```

- [x] If external libraries/packages with restrictive licenses were added, they were added in the Legal Notice section.  
- [x] Linters (Ruff) gave 0 errors. If you have correctly installed pre-commit, it does these checks and adjustments on your behalf.  
- [x] I have added tests for the feature/bug I solved (see tests folder). All the tests (new and old ones) gave 0 errors.  
- [x] If the GUI has been modified: *(N/A)*  
- [x] After you had submitted the PR, if DeepSource, Django Doctors or other third-party linters have triggered any alerts during the CI checks, I have solved those alerts.

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/misp.py
+++ b/api_app/analyzers_manager/observable_analyzers/misp.py
@@ -51,8 +51,7 @@ class MISP(classes.ObservableAnalyzer):
             return [cls]
 
         raise AnalyzerConfigurationException(
-            f"Observable {cls} not supported."
-            "Currently supported are: ip, domain, hash, url, generic."
+            f"Observable {cls} not supported. Currently supported are: ip, domain, hash, url, generic."
         )
 
     def run(self):


mlodic:
you need to read and follow the PR and contribution guidelines

PR #3443, opened 2026-03-07
Establish dynamic Data Model visualization.

### LINKED ISSUE #3420: Visualizers are silently hiding valuable API metadata due to hardcoded data extraction (Data Ghosting)
# What happened
  Currently, our visualizer scripts rely on static, hardcoded logic to extract metadata from raw API reports. Because of this, valuable threat intelligence that the backend successfully fetches is completely hidden from the Visualizer UI.

For example, in api_app/visualizers_manager/visualizers/ip_reputation_services.py, the _abuse_ipdb method explicitly extracts only two fields from the metadata: isp and usageType.

However, the AbuseIPDB API documentation, fields such as abuseConfidenceScore, countryCode, domain, and totalReports are typically included in the response, returns a wealth of other critical intel inside its data object, such as abuseConfidenceScore, countryCode, domain, and totalReports. Because these fields aren't explicitly typed out in the Python script, they are dropped from the Visualizer.
This creates a "data ghosting" effect. Analysts are forced to manually dig through the Raw JSON tab which often contains hundreds of lines of nested arrays just to find a simple data point like the confidence score.

https://github.com/user-attachments/assets/744a818d-4a6f-4a0d-b985-a450a59b5df6

#  Environment
   1. OS: Linux
   2. IntelOwl version: v6.5.1

#  What did you expect to happen

A cleaner approach would be adding a helper method (like _get_dynamic_metadata) to the base Visualizer class. This method could iterate through the JSON response, filter out the fields we are already explicitly handling (like isp), and automatically group the rest into an "Additional Metadata" list in the UI.
It could also format the keys to be more readable on the fly for example, converting abuseConfidenceScore to "Abuse Confidence Score".
This would make the visualizers much easier to maintain. If an external API adds a new data field in the future, IntelOwl will just pick it up and display it automatically, without a developer needing to update the visualizer script.

 # How to reproduce your issue
   1. Go to New Analysis and enter an IP observable (e.g., 7.7.7.7).
   2. Run the analysis using the AbuseIPDB analyzer.
   3. Open the completed report and look at the Raw tab. If you scroll past the massive reports array, you will see that the data dictionary contains many fields like abuseConfidenceScore, domain, and totalReports.
   4. Switch to the Visualizer tab.
   5. Notice that under the AbuseIPDB Meta section, only the isp and usageType are displayed. All other valuable metadata is missing from the GUI.

#  Error messages and logs
  There are no explicit system errors or crashes. This is a logic and visibility issue where data is successfully retrieved by the backend but silently ignored by the frontend UI.

ayushgupta704:
# Description

I've implemented a comprehensive refactor to resolve the "data ghosting" issue by establishing the Data Model as the unified interface for both analysis storage and visualization.

Instead of relying on brittle, hardcoded extraction logic within individual visualizers, this PR introduces a Dynamic Discovery Engine in the base Visualizer class. This engine automatically detects, humanizes, and renders all metadata stored in the additional_info JSON field of any Data Model. This ensures that no valuable API data is left hidden from the user.

  Key Improvements:
   * Dynamic Metadata Rendering: Implemented _get_dynamic_metadata and _humanize_key in the Visualizer base class. This allows the UI to stay in sync with API changes automatically  without requiring visualizer code updates.
   * Normalized the analyzers: I updated AbuseIPDB, GreyNoise, MaxMind, and OTX to make sure they are actually populating the Data Model correctly. I mapped the standard fields (like ISP and Country) and dumped the rest into additional_info.
   * Cleaned up the Reputation tab: I simplified the Reputation visualizer. It’s now just a "quick look" with status badges. I moved all the deep, forensic details over to the DataModel tab to keep the UI from getting cluttered.
   * Fixed false positives: I also refined the logic so that clean/trusted IPs (like 1.1.1.1) don't trigger red badges anymore.

https://github.com/user-attachments/assets/d2f16e1f-9d8c-453a-ae66-655ceda43288

  Type of change

  Checklist
   - [x] I read the contributing rules.
   - [x] This is for the develop branch.
   - [x] I followed the "How to create a Plugin" guide.
   - [x] I added the data migration using dumpplugin.
   - [x] Added the copyright banner.
   - [x] Ruff linting is green (0 errors).
   - [x] All 56 backend tests passed locally.
   - [x] Fixed all DeepSource and CodeQL alerts (B904, B008, E501).

Closes #3420


mlodic:
thanks for the efforts but I don't think this change should be merged.
The way you populate the data models and the visualizer does compromise the original scope of them.
Data models are made to extract only some of the data, that is common to other analyzers as well, not to be filled with all the information available in the analyzer.
Also, visualizers should be predictable to the users. If dynamically created, they would create too much noise and provide too chaotic results. They should be static and clear. That is intended. 
Visualizers and data models have not been created to just make the raw JSON data more affordable to read, they are made to select and organize only the right data that matters to make it more usable.

mlodic:
Also, you should wait to be assigned to the issue next time. I just did it 2 minutes ago but I already found the PR opened. That is what happens if you rush to implementation without waiting to have a proper discussion.

PR #3453, opened 2026-03-11
fix: Talos IP blocklist URL returns HTML T&C page. Closes #3451

### LINKED ISSUE #3451: Deprecate and remove TalosReputation analyzer
## What happened
I am currently assigned working on refactoring the analyzers so that the external datasets are stored in DB #3427 

while doing it i found out that 
in file `api_app/analyzers_manager/observable_analyzers/talos.py`, line 42-44
``` python
url = "https://snort.org/downloads/ip-block-list"
            r = requests.get(url)
            r.raise_for_status()
``` 

This url requires accepting t&c and then redirects to amazonaws hosted file. read online that cisco updated to add this page in between in 2024.

So what should we do about it, use some other source or what?

## Environment
1. OS: Any
2. IntelOwl version: 6.5.1

The file should have been downloaded as .txt

Just visit the url [here](https://snort.org/downloads/ip-block-list)

devs6186:
(Closes #3451)

# Description

The Talos analyzer's `update()` method fetches the IP blocklist from `https://snort.org/downloads/ip-block-list`. As of 2024, Cisco/Snort requires users to accept terms and conditions in a browser before downloading. Unauthenticated `GET` requests now receive an HTML T&C page instead of the expected plain-text blocklist. The previous code wrote this HTML directly to the local database file, silently corrupting it and causing all subsequent lookups to fail.

## Type of change

# Checklist

---

## Root Cause

`Talos.update()` called `requests.get(url)` with no Content-Type check. When the response is an HTML page (the T&C gate), the code decoded and wrote it to the blocklist file without any validation. There was also no request timeout, so the method could hang indefinitely.

## Solution

After `raise_for_status()`, inspect the `Content-Type` response header. If it contains `text/html`, log a warning that includes the manual download URL and the expected local file path (`MEDIA_ROOT/talos_ip_blacklist.txt`), then return `False`. This lets the analyzer fall back to an existing cached file rather than overwriting it with garbage. A 60-second timeout was added to the `requests.get()` call. A module docstring was added explaining the manual download path for environments where automatic download is not possible.

## Changes Made

- `api_app/analyzers_manager/observable_analyzers/talos.py`
  - Added module docstring documenting the T&C limitation and manual download path
  - Added `timeout=60` to `requests.get()`
  - Added Content-Type check: if `text/html` is present, logs a warning and returns `False`

## Testing

- Manually confirmed the Content-Type branch is reached when the URL returns an HTML page (simulated by pointing at an HTML-serving URL)
- Confirmed the analyzer still functions correctly when a valid local file is present and `update()` returns `False`

## Edge Cases

- If `Content-Type` header is absent, `headers.get("Content-Type") or ""` returns an empty string, so the HTML check evaluates to `False` and the existing path proceeds normally
- The `return False` path preserves any previously cached valid blocklist file

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/talos.py
+++ b/api_app/analyzers_manager/observable_analyzers/talos.py
@@ -1,6 +1,15 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
+"""
+Talos IP blocklist analyzer (Cisco/Snort).
+
+Note: As of 2024, the blocklist at snort.org requires accepting terms and conditions
+in a browser. If automatic download fails (e.g. HTML T&C page returned), download
+the list manually from https://snort.org/downloads/ip-block-list and place it at
+MEDIA_ROOT/talos_ip_blacklist.txt.
+"""
+
 import logging
 import os
 
@@ -39,8 +48,18 @@ class Talos(classes.ObservableAnalyzer):
         try:
             logger.info("starting download of db from talos")
             url = "https://snort.org/downloads/ip-block-list"
-            r = requests.get(url)
+            r = requests.get(url, timeout=60)
             r.raise_for_status()
+
+            content_type = (r.headers.get("Content-Type") or "").lower()
+            if "text/html" in content_type:
+                logger.warning(
+                    "Talos blocklist URL returned HTML (likely T&C page). "
+                    "Download the list manually from %s and place it at %s",
+                    url,
+                    database_location,
+                )
+                return False
 
             with open(database_location, "w", encoding="utf-8") as f:
                 f.write(r.content.decode())


sanjib2006:
Hey @devs6186 , thanks for the quick fix! 

But the root cause is not resolved, probably it would be better to use another source to get that list.
I'm currently working on a refactor of local-DB analyzers. Once `mlodic` decides on a replacement URL, I can include it in that refactor.
Let's wait for mlodic's input on the best path forward.

sanjib2006:
I'm also a contributor here. I noticed you've opened some PRs previously.

Just a small heads-up: the project usually prefers creating new branches from the **develop** branch (it looks like this one might be based on `master`). Also, PRs are typically opened for issues that you're assigned to.

You can check the contributing guide [here](https://intelowlproject.github.io/docs/IntelOwl/contribute/)

Feel free to reach out if you need help with initial setup or configuration :) btw the docs are quite good and cover most of what you need.

mlodic:
not following guidelines

PR #3341, opened 2026-02-20
feat: optimize read_configured_params. Closes #3313

berardifra:
Closes #3313

# Description

**Before:** Redundant database operations with `.exists()` + `.first()` pattern (2 queries instead of 1), no `select_related` causing lazy-load N+1 on `param.python_module.module` access in error path, and redundant boolean condition `not settings.STAGE_CI or settings.STAGE_CI and not param.value`.

**After:** Clean optimized implementation with:
- Single `.first()` call + `is not None` check instead of `.exists()` + `.first()` (2→1 DB queries)
- `select_related("python_module")` on the filtered queryset to prevent lazy-load query on error path
- Simplified boolean: `not settings.STAGE_CI or not param.value` (logically equivalent, removes redundancy)
- 5 dedicated unit tests covering all branches (`ReadConfiguredParamsTestCase`)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue).

# Checklist

- [x] I have read and understood the rules about [how to Contribute](https://intelowlproject.github.io/docs/IntelOwl/contribute/) to this project
- [x] The pull request is for the branch `develop`
- [x] I have inserted the copyright banner at the start of the file
- [x] Please avoid adding new libraries as requirements whenever it is possible. No new libraries added.
- [x] Linters (`Ruff`) gave 0 errors.
- [x] I have added tests for the feature/bug I solved (see `tests` folder). All the tests (new and old ones) gave 0 errors.
- [x] After you had submitted the PR, if `DeepSource`, `Django Doctors` or other third-party linters have triggered any alerts during the CI checks, I have solved those alerts.

### Important Rules
- If you miss to compile the Checklist properly, your PR won't be reviewed by the maintainers.
- Everytime you make changes to the PR and you think the work is done, you should explicitly ask for a review by using GitHub's reviewing system detailed [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review).

PR #3454, opened 2026-03-11
perf: eliminate redundant DB query in read_configured_params. Closes #3437

### LINKED ISSUE #3437: Optimization: Redundant Database Query in PythonConfig.read_configured_params()

devs6186:
(Closes #3437)

# Description

`PythonConfig.read_configured_params()` checked for unconfigured required parameters using two separate queryset operations: `.exists()` followed by `.first()`. Both hit the database independently, performing two `SELECT ... LIMIT 1` queries when only one is needed.

## Type of change

# Checklist

- [x] Please avoid adding new libraries as requirements whenever it is possible.

---

## Root Cause

The original code:

```python
not_configured_params = params.filter(required=True, configured=False)
if not_configured_params.exists():
    param = not_configured_params.first()
    if not settings.STAGE_CI or ...:
        raise TypeError(...)
```

Django's `.exists()` executes `SELECT (1) AS "a" FROM ... LIMIT 1` and `.first()` executes `SELECT ... LIMIT 1` — two round-trips for the same logical check. Since `.first()` returns `None` when the queryset is empty, the `.exists()` guard is entirely redundant.

## Solution

Replace the two-call pattern with a single `.first()` and an `if param is not None` guard:

```python
param = not_configured_params.first()
if param is not None:
    if not settings.STAGE_CI or ...:
        raise TypeError(...)
```

This is the idiomatic Django pattern and halves the number of DB queries on every plugin health check and job initialization that calls `read_configured_params()`.

## Changes Made

- `api_app/models.py` — `PythonConfig.read_configured_params()`: replaced `exists()` + `first()` with a single `first()` call and `if param is not None` guard.

## Testing

- Confirmed the behavior is identical: the `TypeError` is raised when and only when a required unconfigured parameter exists

## Edge Cases

- When no unconfigured required parameters exist, `.first()` returns `None` and the `if` block is skipped — same behavior as before
- The `STAGE_CI` logic inside the block is unchanged

### DIFF
--- a/api_app/models.py
+++ b/api_app/models.py
@@ -1684,9 +1684,8 @@ class PythonConfig(AbstractConfig):
             self, user, config_runtime
         )
         not_configured_params = params.filter(required=True, configured=False)
-        # TODO to optimize
-        if not_configured_params.exists():
-            param = not_configured_params.first()
+        param = not_configured_params.first()
+        if param is not None:
             if not settings.STAGE_CI or settings.STAGE_CI and not param.value:
                 raise TypeError(
                     f"Required param {param.name} "


devs6186:
Closing this PR because the fix has already been incorporated into the `develop` branch as part of #3341 (commit `05889958`). That commit applied the same optimization — replacing `.exists()` + `.first()` with a single `.first()` call and an `if param is not None` guard — and additionally added `.select_related("python_module")` to avoid the lazy-load on `param.python_module.module` in the error path, along with a simplification of the boolean condition.

Issue #3437 tracks this exact redundancy. Since the fix is already in `develop`, maintainers may want to close that issue too (or it will be closed once the next release is cut from `develop`).

PR #3455, opened 2026-03-11
fix: semantic_version_comp crashes on 'v'-prefixed versions. Closes #3446

### LINKED ISSUE #3446: [BUG] initialize.sh: semantic_version_comp crashes on versions with "v" prefix
## What happened
1. The `semantic_version_comp` function in `initialize.sh` crashes when either version string passed as an argument contains a `v` prefix (e.g., `v2.3.4`).
2. `initialize.sh` fails to detect the Docker Compose version correctly if the output of `docker compose version` does not contain a `v` prefix before the version number (e.g., `Docker Compose version 5.0.2`). This is due to a fragile `cut` command at line 121.

The crash in `semantic_version_comp` occurs because the function attempts to perform arithmetic comparison on a string containing `v`, leading to:
`((: 10#v2: value too great for base (error token is "10#v2")`

The failure at line 121 results in an empty `docker_compose_version` variable, which then causes `semantic_version_comp` to receive an empty string, leading to incorrect "lessThan" evaluation or further crashes.

## Environment
1. OS: linux
2. IntelOwl version: v6.5.1-32-g99f82032
3. Docker Compose version: 5.0.2 (which doesn't start with 'v')

The function should correctly strip the `v` prefix and compare the version numbers as integers.

You can reproduce this by calling `semantic_version_comp` with a `v`-prefixed version:

```bash
# In initialize.sh or a separate script
semantic_version_comp "v2.3.4" "1.0.0"
```

A standalone reproduction script:
```bash
#!/usr/bin/env bash
semantic_version_comp () {
  ver1="${1//v/}"
  ver2="${2//v/}"
  local IFS=.
  read -ra ver1_arr <<< "$1" # Bug: should be $ver1
  read -ra ver2_arr <<< "$2" # Bug: should be $ver2
  if ((10#${ver1_arr[0]} > 0)); then echo "OK"; fi
}
semantic_version_comp "v2.3.4" "1.0.0"
```

## Error messages and logs
```
/tmp/repro_bug.sh: line 22: ((: 10#v2: value too great for base (error token is "10#v2")
```

Also, if the version parsing at line 121 fails, the script might output:
`Error: Docker compose version is too old. Please upgrade to at least 2.3.4.` (even if the version is newer).

devs6186:
(Closes #3446)

# Description

`semantic_version_comp` in `initialize.sh` crashed with a bad substitution error when version strings had a "v" prefix (e.g. `v2.3.4`). Additionally, newer versions of docker compose no longer include a "v" in their version output, causing the version extraction to silently return an empty string.

## Type of change

# Checklist

---

## Root Cause

Two separate bugs:

**Bug 1 — `semantic_version_comp`:** The function stripped the "v" prefix into local variables `ver1`/`ver2` but then passed the original positional parameters `$1`/`$2` to `read -ra`:

```bash
ver1="${1//v/}"
ver2="${2//v/}"
local IFS=.
read -ra ver1 <<< "$1"   # BUG: uses $1, not the stripped $ver1
read -ra ver2 <<< "$2"   # BUG: uses $2, not the stripped $ver2
```

This populated the arrays with elements like `"v2"`, causing `((10#v2 > ...))` to fail with `value too great for base`.

**Bug 2 — docker compose version parsing:** The version was extracted with `cut -d'v' -f3`, which assumes the output contains a "v". Docker Compose v2.20+ may output `"Docker Compose version 5.0.2"` (no "v"), returning an empty string from `cut`.

## Solution

- Renamed the arrays to `ver1_arr`/`ver2_arr` and populated them from the stripped variables: `read -ra ver1_arr <<< "$ver1"`
- Updated all array references in the comparison loop to use `ver1_arr`/`ver2_arr`
- Replaced `cut -d'v' -f3` with `grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1` which extracts the semver regardless of prefix
- Added an empty-string guard: if extraction produces an empty string, the error path is taken

## Changes Made

- `initialize.sh` — `semantic_version_comp`: renamed arrays and fixed `read -ra` source variables
- `initialize.sh` — docker compose version detection: replaced `cut`-based extraction with `grep -oE`

## Testing

Verified manually with version strings: `v2.3.4`, `2.3.4`, `v19.03.0`, `Docker Compose version 5.0.2`, `Docker Compose version v2.3.4`.

## Edge Cases

- Versions without a "v" prefix continue to work unchanged
- If `docker compose version` produces unexpected output, `grep -oE` returns empty, which is caught by the `[[ -z "$docker_compose_version" ]]` guard and results in a clear error message

### DIFF
--- a/initialize.sh
+++ b/initialize.sh
@@ -12,30 +12,29 @@ semantic_version_comp () {
       return
   fi
 
-  # Remove "v" prefix if present
-  ver1="${1//v/}"  # Used parameter substitution instead of sed (SC2001)
+  # Remove "v" prefix if present (e.g. v2.3.4 or Docker Compose version 5.0.2)
+  ver1="${1//v/}"
   ver2="${2//v/}"
 
-  # Convert version numbers to arrays
+  # Convert version numbers to arrays using stripped values
   local IFS=.
-  read -ra ver1 <<< "$1"
-  read -ra ver2 <<< "$2"
-  # Fill empty fields in ver1 with zeros
-  for ((i=${#ver1[@]}; i<${#ver2[@]}; i++)); do
-      ver1[i]=0
+  read -ra ver1_arr <<< "$ver1"
+  read -ra ver2_arr <<< "$ver2"
+  # Fill empty fields in ver1_arr with zeros
+  for ((i=${#ver1_arr[@]}; i<${#ver2_arr[@]}; i++)); do
+      ver1_arr[i]=0
   done
 
   # Compare version numbers
-  for ((i=0; i<${#ver1[@]}; i++)); do
-      if [[ -z ${ver2[i]} ]]; then
-          # Fill empty fields in ver2 with zeros
-          ver2[i]=0
+  for ((i=0; i<${#ver1_arr[@]}; i++)); do
+      if [[ -z ${ver2_arr[i]} ]]; then
+          ver2_arr[i]=0
       fi
-      if ((10#${ver1[i]} > 10#${ver2[i]})); then
+      if ((10#${ver1_arr[i]} > 10#${ver2_arr[i]})); then
           echo "greaterThan"
           return
       fi
-      if ((10#${ver1[i]} < 10#${ver2[i]})); then
+      if ((10#${ver1_arr[i]} < 10#${ver2_arr[i]})); then
           echo "lessThan"
           return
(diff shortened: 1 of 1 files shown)


mlodic:
respect the rules, you weren't be assigned

PR #3457, opened 2026-03-11
feat: deduplicate identical data models via content hash. Closes #3450

### LINKED ISSUE #3450: Reusing identical Data Models to prevent Data Bloat (Found a # TODO)
# What happened
While I was working on some visualizers and checking the Django Admin panel, I noticed that IntelOwl is creating a lot of redundant rows in the Data Model tables (IPDataModel, DomainDataModel, etc.).

Basically, every time an analyzer runs successfully, it creates a brand new row in the database. If I run three different analyzers for the same IP (like 1.1.1.1), and they all find the same ASN and ISP info, I end up with three identical rows in the database instead of just one. Over time, this is going to make the database much larger and slower than it
needs to be.

# Environment
- OS: Linux 

If the intelligence data (the dictionary) returned by an analyzer is exactly the same as something we already have in the database, the system should be smart enough to just link to the existing record instead of printing a new one.

# How to reproduce your issue

1. Run a few scans for the same IP (e.g., 8.8.8.8) using analyzers like AbuseIPDB.
2. Go to the Django Admin panel at /admin/data_model_manager/ipdatamodel/.
3. You'll see multiple rows that have different IDs (PKs) but the exact same data (same ASN, same Org Name, etc.).

Image
Image

- I actually found a note in the code where the maintainers already caught this! In api_app/analyzers_manager/models.py on line 124, there's a       # TODO that says we don't need to actually crate a new object every time. if the report is the same of the previous one, we can just link it.

Image

## Error messages and logs
There aren't any errors per se because the code doesn't crash, but you can see the bloat 
in the database records. Here is an example of what my IPDataModel table looks like after a
few scans of the same IP:
- PK 230 | ASN 13335 | ORG Cloudflare | COUNTRY US
- PK 231 | ASN 13335 | ORG Cloudflare | COUNTRY US
- PK 232 | ASN 13335 | ORG Cloudflare | COUNTRY US

As you can see, IDs 230, 231, and 232 are identical. We should probably be using something
like get_or_create with a fingerprint/hash to handle this.

devs6186:
(Closes #3450)

# Description

Every successful analyzer run that produced the same content (e.g. the same ASN, country, and evaluation for a well-known IP across multiple jobs) created a brand-new data model row via `objects.create()` followed by `merge()`. Over time this causes significant table bloat and wastes storage on rows that are identical in all meaningful fields. A `TODO` comment in `create_data_model()` explicitly acknowledged this and asked for a fix.

## Type of change

# Checklist

- [x] Please avoid adding new libraries as requirements whenever it is possible.

---

## Root Cause

`AnalyzerReport.create_data_model()` always called `self.data_model_class.objects.create()` regardless of whether an identical data model already existed. There was no deduplication mechanism.

## Solution

Added a `content_hash` field to `BaseDataModel` — a SHA-256 hash of the normalized content dictionary used to populate the model. Before creating a new row, `create_data_model()` computes this hash and checks for an existing row with the same hash. If one exists, the report is linked to it and returned immediately (no INSERT). If not, the new row is created, merged, and the hash stored as usual.

**Normalization** (`_normalize_for_content_hash`):
- Dicts → sorted by key recursively
- Lists/sets/tuples → sorted by `str()` representation (avoids `TypeError` with mixed-type lists)
- ORM model instances → their primary key
- Everything else → value as-is

`json.dumps(..., sort_keys=True, default=str)` on the normalized output ensures consistent serialization before hashing.

**Migration:** `0012_basedatamodel_content_hash` adds the `content_hash` column (nullable, unique, indexed) to `ipdatamodel`, `domaindatamodel`, and `filedatamodel`. Nullable so existing rows are unaffected — no data migration needed.

## Changes Made

- `api_app/data_model_manager/models.py` — added `content_hash` field to `BaseDataModel`
- `api_app/analyzers_manager/models.py` — added `_normalize_for_content_hash()`, `_compute_data_model_content_hash()`, and reuse-by-hash logic in `create_data_model()`
- `api_app/data_model_manager/migrations/0012_basedatamodel_content_hash.py` — new migration

## Testing

- Confirmed that two `create_data_model()` calls with identical dictionaries produce the same hash and reuse the first row
- Confirmed that calls with different dictionaries produce different hashes and create separate rows
- Confirmed existing rows (null `content_hash`) are unaffected

## Edge Cases

- Hash collision is astronomically unlikely with SHA-256 (2^256 space)
- If two concurrent jobs produce the same hash simultaneously, one `objects.create()` may violate the unique constraint; this is a rare race condition that can be addressed separately (e.g. `get_or_create`) if it becomes a problem in practice
- The `unique=True` constraint ensures no two rows ever share a hash, maintaining data integrity

### DIFF
--- a/api_app/analyzers_manager/models.py
+++ b/api_app/analyzers_manager/models.py
@@ -1,8 +1,9 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
+import hashlib
 import json
 from logging import getLogger
-from typing import Dict, Optional, Type, Union
+from typing import Any, Dict, Optional, Type, Union
 
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
@@ -20,6 +21,25 @@ from api_app.fields import ChoiceArrayField
 from api_app.models import AbstractReport, PythonConfig, PythonModule
 
 logger = getLogger(__name__)
+
+
+def _normalize_for_content_hash(value: Any) -> Any:
+    """Normalize a value for stable content hashing (for data model deduplication)."""
+    if value is None:
+        return None
+    if hasattr(value, "pk"):
+        return value.pk
+    if isinstance(value, dict):
+        return {k: _normalize_for_content_hash(v) for k, v in sorted(value.items())}
+    if isinstance(value, (list, set, tuple)):
+        return sorted((_normalize_for_content_hash(x) for x in value), key=str)
+    return value
+
+
+def _compute_data_model_content_hash(dictionary: Dict) -> str:
+    """Compute a stable hash of the data model dictionary for deduplication."""
+    normalized = _normalize_for_content_hash(dictionary)
+    return hashlib.sha256(json.dumps(normalized, sort_keys=True, default=str).encode()).hexdigest()
 
 
 class AnalyzerReport(AbstractReport):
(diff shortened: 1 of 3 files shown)


mlodic:
don't open PRs if not assigned

PR #3458, opened 2026-03-12
Fix semantic_version_comp crash on "v" prefixed versions and simplify Docker Compose version parsing. Closes #3446

### LINKED ISSUE #3446: [BUG] initialize.sh: semantic_version_comp crashes on versions with "v" prefix

chauhan-varun:
# Description

This PR fixes two related bugs in `initialize.sh` that cause the script to fail during installation:

### Bug 1: `semantic_version_comp` crashes on `v`-prefixed versions

Lines 16-17 correctly strip the `v` prefix from version strings into `ver1`/`ver2`, but lines 21-22 overwrite them by reading from the original `$1`/`$2` arguments instead of the stripped values. When a version like `v2.3.4` is passed, `read -ra ver1 <<< "v2.3.4"` produces `["v2","3","4"]`, and the subsequent arithmetic comparison `((10#${ver1[0]}))` with `ver1[0]="v2"` causes a bash error:

image

**Fix:** Changed `read -ra ver1 <<< "$1"` → `read -ra ver1 <<< "$ver1"` (and same for `ver2`) so the stripped values are actually used.

### Bug 2: Docker Compose version parsing fails without `v` prefix

Line 121 uses `cut -d 'v' -f3` to extract the version number from `docker compose version` output. This assumes the output always contains a `v` before the version number (e.g., `Docker Compose version v2.3.4`). However, newer versions of Docker Compose output `Docker Compose version 5.0.2` (without the `v`), causing `cut` to return an empty string. This leads to an incorrect "version too old" error even when the version is valid.
## Before
image

## After
image

**Fix:** Replaced the fragile `cut` command with `docker compose version --short`, which reliably returns just the version number regardless of format.

Closes #3446

## Type of change

# Checklist

    - N/A
- [ ] I have inserted the copyright banner at the start of the file: N/A (shell script, not a Python module)
- [ ] If external libraries/packages with restrictive licenses were added, they were added in the [Legal Notice](https://github.com/certego/IntelOwl/blob/master/.github/legal_notice.md) section. N/A
- [ ] If the GUI has been modified: N/A

### DIFF
--- a/initialize.sh
+++ b/initialize.sh
@@ -18,8 +18,8 @@ semantic_version_comp () {
 
   # Convert version numbers to arrays
   local IFS=.
-  read -ra ver1 <<< "$1"
-  read -ra ver2 <<< "$2"
+  read -ra ver1 <<< "$ver1"
+  read -ra ver2 <<< "$ver2"
   # Fill empty fields in ver1 with zeros
   for ((i=${#ver1[@]}; i<${#ver2[@]}; i++)); do
       ver1[i]=0
@@ -118,7 +118,7 @@ if ! docker compose version; then
   fi
 else
   # docker compose exists
-  docker_compose_version="$(docker compose version | cut -d 'v' -f3)"
+  docker_compose_version="$(docker compose version --short)"
   if [[ $(semantic_version_comp "$docker_compose_version" "$MINIMUM_DOCKER_COMPOSE_VERSION") == "lessThan" ]]; then
     echo "Error: Docker compose version is too old. Please upgrade to at least $MINIMUM_DOCKER_COMPOSE_VERSION." >&2
     exit 1


chauhan-varun:
@mlodic please review the PR

mlodic:
show screenshot that prove that this works as intended with the cited case. We can merge it only with proof that you tested it locally and that this is reproducible

chauhan-varun:
@mlodic i've added the screenshots in the PR discription please take a look

PR #3422, opened 2026-03-03
Auto-search artifacts after new evaluation submission

lvb05:
## Description

When a user adds a new evaluation on the Artifacts page, the search results were not automatically updated. The user had to manually type and search for the IOC to view its evaluation, requiring extra clicks.
This PR fixes that by wiring an `onEvaluationSuccess` callback in `Analyzables.jsx`  that triggers automatically after a successful submission in `UserEventModal`. If a single IOC was submitted, it auto-populates the search field and triggers `doSearch` with that value. If multiple IOCs were submitted, only the first one is searched, as suggested in the issue.

Closes #3148

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue).

## Checklist

- [x] I have read and understood the rules about how to Contribute to this project
- [x] The pull request is for the branch `develop`
- [x] I have added tests for the feature/bug I solved. All the tests (new and old ones) gave 0 errors.
- [x] If the GUI has been modified:
    - [x] I have created new frontend tests for the new component or updated existing ones.
- [x] Linters gave 0 errors.

PR #3466, opened 2026-03-13
fix: auto-search for brand new artifacts after evaluation submission. Closes #3148

### LINKED ISSUE #3148: Visualization of artifacts search results after addition of a new one
In case we are in the Artifacts page and we add a new evaluation, at the end of the evaluation we get a toast telling us that everything was fine but I would have expected to also automatically get a search for that added IOC in the page so that I can see all its evaluation. Without having this, additional clicks and inputs are required.

In case more than an IOC is involved in this, just avoid this feature or just search the first one

lvb05:
## Description

This PR fixes an edge case missed in #3422: the auto-search after evaluation submission was not working for completely new artifacts that had never been analyzed or searched. When a user clicks "New evaluation" from the top right of the Artifacts page and types a brand new IOC, `inputState[analyzable]` is `undefined` because the debounce hasn't fired yet, or the artifact has never been seen before. This caused a silent crash in `onSubmit` before the `onSubmitCallback` could fire, so the artifact never appeared in the results.
The fix adds a safe `??` fallback so that if `inputState[analyzable]` is undefined, it defaults to `{ type: ANALYZABLE, eventId: null }` and correctly proceeds with `axios.post` to create the new evaluation and trigger the auto-search.

Closes #3148
This is a follow-up to #3422

## Type of change

# Checklist

- [x] I have read and understood the rules about how to contribute to this project
- [x] Linters gave 0 errors.
    - [x] I have provided a screenshot of the result in the PR.
    
image

image

### DIFF
--- a/frontend/src/components/userEvents/UserEventModal.jsx
+++ b/frontend/src/components/userEvents/UserEventModal.jsx
@@ -204,26 +204,24 @@ export function UserEventModal({
       const failed = [];
       Promise.allSettled(
         formik.values.analyzables.map((analyzable) => {
-          if (inputState[analyzable].type === UserEventTypes.IP_WILDCARD)
+          const state = inputState[analyzable] ?? {
+            type: UserEventTypes.ANALYZABLE,
+            eventId: null,
+          };
+          if (state.type === UserEventTypes.IP_WILDCARD)
             evaluation.network = analyzable;
-          else if (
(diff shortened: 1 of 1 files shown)


lvb05:
@mlodic  @fgibertoni 

https://github.com/user-attachments/assets/b286e778-863a-4d88-a474-a14963849789

PR #3499, opened 2026-03-20
fix(forms): assign str.replace() result in _cleaning_and_multiline

Akkhan88:
## Description
Fixed a bug in `api_app/forms.py` where `str.replace()` results were being discarded in `_cleaning_and_multiline()`.

Python strings are **immutable** — `.replace()` returns a new string and does not modify in place. The current code calls `line.replace()` without assigning the result back, so carriage returns (`\r`) and double quotes (`"`) are never actually removed.

## Proof

```python
line = 'hello "world"\r'
line.replace("\r", "")         # ❌ line is still 'hello "world"\r'
line = line.replace("\r", "")  # ✅ line is now 'hello "world"'
```

## Changes
- `api_app/forms.py` line 41: `line.replace("\r", "")` → `line = line.replace("\r", "")`
- `api_app/forms.py` line 42: `line.replace('"', "")` → `line = line.replace('"', "")`

## Checklist
- [x] Bug fix (non-breaking change that fixes an issue)
- [x] No new dependencies added
- [x] Minimal 2-line change

### DIFF
--- a/api_app/forms.py
+++ b/api_app/forms.py
@@ -38,8 +38,8 @@ class MultilineJSONField(forms.JSONField):
         if value is not None and "\n" in value:
             cleaned_value = []
             for line in value.splitlines():
-                line.replace("\r", "")
-                line.replace('"', "")
+                line = line.replace("\r", "")
+                line = line.replace('"', "")
                 line = line + "\\n"
                 cleaned_value.append(line)
             value = '"' + "".join(cleaned_value) + '"'


Akkhan88:
Hi, I wanted to confirm whether removing double quotes here is the intended behavior.

Looking at the `_cleaning_and_multiline` method, it processes multiline JSON input. If this input is expected to contain JSON-like content, removing quotes could potentially affect parsing downstream (since `"` is structurally significant in JSON).

The docstring says "remove carriage returns and quotes" — so the *intent* seems clear, but the original code never actually did it (due to the immutability bug). So this fix might surface new behavior that wasn't happening before.

Happy to adjust the fix based on the expected input format. Let me know if any edge cases need handling!

PR #3501, opened 2026-03-20
feat: Refine HTTP timeouts across numerous analyzers and ingestors. Closes #3495

### LINKED ISSUE #3495: [STABILITY] Potential Worker Starvation due to Missing Request Timeouts
## What happened
Across the codebase, specifically in many analyzers and connectors, the `requests` library is frequently used without an explicit `timeout` parameter. 

### Affected Components (Examples):
- `api_app/analyzers_manager/observable_analyzers/stratosphere.py` (Line 66)
- `api_app/analyzers_manager/observable_analyzers/abuseipdb.py` (Line 29)
- Many other analyzers and connectors using `requests.get()` or `requests.post()`.

Because IntelOwl uses a fixed-size pool of Celery workers, a malicious or slow external service can keep a request open indefinitely (a "slow-tail" attack). If an attacker targets multiple analyzers at once, they can occupy every available worker, effectively causing a Denial of Service (DoS) where no legitimate scans can be processed.

## Environment
1. OS: Linux
2. IntelOwl version: Current develop branch

All outgoing HTTP requests should have a reasonable, explicit timeout (e.g., `requests.get(url, timeout=30)`).

1. Point an analyzer (like the Stratosphere or any web-based analyzer) to a server designed to never close the connection (e.g., `slowloris` or a simple infinite loop server).
2. Observe that the Celery worker for that task stays in a "Started" state forever and never returns to the pool.
3. Repeat for all worker slots to halt the entire system.

## Error messages and logs
No error messages are generated because the worker simply hangs. This is visible in Celery monitoring tools like Flower as tasks that never progress.

## Suggested Fix
1. Define a global default timeout for all HTTP requests within IntelOwl's utility modules.
2. Iterate through all analyzers and connectors to ensure they use this default timeout or an explicit one.
3. Consider using a wrapper for `requests` or switching to a library that enforces timeouts.
or messages are generated because the worker simply hangs. This is visible in Celery monitoring tools like Flower as tasks that never progress.

chauhan-varun:
# Description

This PR implements a centralized HTTP timeout mechanism project-wide to prevent indefinite hangs in Celery workers, as said in issue #3495.

Key changes:
- Introduced a global `HTTP_TIMEOUT` setting in `intel_owl/settings/commons.py` (default 30 seconds).
- Created a robust wrapper in `api_app/http_utils.py` for the `requests` library which automatically applies the default timeout to all requests (including a new `Session` wrapper).
- Migrated over 125 files, including analyzers, connectors, and ingestors, from direct `requests` calls to `http_utils`.
- Refined the `health_check` method in `api_app/analyzers_manager/classes.py` to use `settings.HTTP_TIMEOUT`.
- Fixed several secondary bugs caught during migration:
    - Added missing `quote` and `urljoin` imports in `mmdb_server.py` and `malprob.py`.
    - Ensured proper URL encoding of observable names in these analyzers.
- Performed a project-wide cleanup with `Ruff`, ensuring 0 linting errors.
- **Verified in Docker**: Ran the full analyzer unit test suite (199 tests) inside the container, resolving all regressions (`AttributeError`, `TypeError`) caused by the migration.

## Type of change

# Checklist

    - [x] If the plugin interacts with an external service, I have created an attribute called precisely `url` that contains this information. This is required for Health Checks (HEAD HTTP requests). 
    - [ ] If a new analyzer has beed added, I have created a unittest for it in the appropriate dir. I have also mocked all the external calls, so that no real calls are being made while testing.
    - [ ] I have added that raw JSON sample to the `get_mocker_response()` method of the unittest class. This serves us to provide a valid sample for testing. 
- [x] I have inserted the copyright banner at the start of the file: ```# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl # See the file 'LICENSE' for copying permission.``` (Retained in migrated files)
- [x] Please avoid adding new libraries as requirements whenever it is possible. Use new libraries only if strictly needed to solve the issue you are working for. In case of doubt, ask a maintainer permission to use a specific library. (Used `urllib.parse` which is standard library)
- [x] If external libraries/packages with restrictive licenses were added, they were added in the [Legal Notice](https://github.com/certego/IntelOwl/blob/master/.github/legal_notice.md) section. (None added)

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/firehol_iplist.py
+++ b/api_app/analyzers_manager/observable_analyzers/firehol_iplist.py
@@ -5,7 +5,6 @@ import ipaddress
 import logging
 import traceback
 
-from django.conf import settings
 from django.db import transaction
 from django.utils import timezone
 
--- a/api_app/analyzers_manager/observable_analyzers/spamhaus_drop.py
+++ b/api_app/analyzers_manager/observable_analyzers/spamhaus_drop.py
@@ -3,7 +3,6 @@ import ipaddress
 import json
 import logging
 
-from django.conf import settings
 from django.db import transaction
 
 from api_app import http_utils
--- a/api_app/analyzers_manager/observable_analyzers/stratosphere.py
+++ b/api_app/analyzers_manager/observable_analyzers/stratosphere.py
@@ -3,7 +3,6 @@
 
 import logging
 
-from django.conf import settings
 from django.db import transaction
 
 from api_app import http_utils


chauhan-varun:
Hey @mlodic, I've finished implementing the centralized `HTTP` timeouts. I've migrated all the analyzers to a new `http_utils` wrapper and verified them with a full test run all 199 passed.

The default timeout is now set to **30s** as requested. Please have a look and let me know if everything looks good! 

image
image

mlodic:
it seems a bit overkill to be honest and could result in possible unwanted behaviors. Not sure if it is worth adding this.
We already have a soft time limit in place which would kill the timeout anyway

chauhan-varun:
I totally understand the concern about breaking long-running analyzers. The reason I thought the HTTP-level timeout was useful is that it allows the analyzer to catch the timeout exception and fail gracefully (showing a nice error in the UI) rather than having Celery kill the entire task, which can be a bit more abrupt.

That said, since my `http_utils` wrapper still allows passing an explicit `timeout`, any analyzer that needs more than 30s can just override the default.

chauhan-varun:
To address the 'wanted behavior' concern, what if I increase the default timeout to 60 or 90 seconds? That would still prevent truly 'infinite' hangs while being safe for slow analyzers.

Or, if you prefer, I can revert the mass migration and only apply this new `http_utils` to the most critical analyzers. Let me know what you think is best!

mlodic:
90 seconds is definitively better

intel_owl/settings/commons.py
```diff
@@ -58,3 +58,10 @@
     "https://api.github.com/repos/intelowlproject/IntelOwl/releases/latest",
 )
 INTEL_OWL_VERSION = VERSION
+try:
```
mlodic, line 61: this would require adding this parameter in the file env_app used by Docker to load env variables and documented its usage

api_app/http_utils.py
```diff
@@ -0,0 +1,97 @@
...
+    the default HTTP_TIMEOUT to all requests.
+    """
+
+    def request(self, method, url, **kwargs):
+        if "timeout" not in kwargs:
+            kwargs["timeout"] = settings.HTTP_TIMEOUT
+        return super().request(method, url, **kwargs)
+
+
+def verify_timeout(func):
```
mlodic, line 85: new contributors won't know about this. Please update the official doc with a statement that says that this helper should be preferred

intel_owl/settings/commons.py
chauhan-varun: Done! Updated the default to 90s and added the HTTP_TIMEOUT variable to the `env_file_app` templates as requested

api_app/http_utils.py
chauhan-varun: created a issue in docs [here](https://github.com/intelowlproject/docs/issues/63) I'll create the pr there once this get merged

api_app/analyzers_manager/classes.py
```diff
@@ -454,7 +455,7 @@ def health_check(self, user: User = None) -> bool:
         basic health check: if instance is up or not (timeout - 10s)
         """
         try:
-            requests.head(self.url, timeout=10)
+            http_utils.head(self.url, timeout=settings.HTTP_TIMEOUT)
```
mlodic: why this have been changed?

chauhan-varun:
The only spots where I'd accidentally changed an existing short timeout were the health_check methods. I've now reverted those back to the original 10s so they stay fast.

api_app/analyzers_manager/classes.py
chauhan-varun: I changed it to settings `HTTP_TIMEOUT` to make sure everything used the same global setting

chauhan-varun:
Hi @mlodic, I've resolved the conflicts and verified the fixes with unit tests in Docker. Ready for another look!

mlodic:
considering the amount of file changed I would prefer to postpone this after the release of April.
I'll keep this open and reconsider later together with the other PRs that provide substantial changes

chauhan-varun:
Thanks, sounds good. Happy to revisit this after the April release

Issue #3460, opened 2026-03-12
[Performance] UserEventQuerySet.decay() runs 2×N database queries instead of a single bulk SQL UPDATE


mannubaveja007:
## What happened
In `api_app/user_events_manager/queryset.py`, the `decay()` method iterates over
UserEvent objects in Python and calls `obj.save()` + `obj.data_model.save()` for each one.
The routine leads to 2* N database queries when dealing with N decaying events; this will strongly affect the performance for a large number of decays.
performance degradation as the number of user events grows.

## Environment
1. OS: Any
2. IntelOwl version: latest develop

## What did you expect to happen
The decay logic should be rewritten using Django `F()` expressions with `Case/When`
update all objects in a `QuerySet` and model fields using bulk `UPDATE` instead of the default `SELECT`/`UPDATE` approach by applying `QuerySet.update()` or `Model.update()`. This can be especially effective with large `QuerySet` instances.
avoiding the N+1 query problem.

## How to reproduce your issue
1. Create many UserEvent records with `next_decay` set to a past datetime
2. Trigger the decay task
3. Enable Django query logging — observe 2×N UPDATE queries being executed

## Error messages and logs
No error. This is a known TODO left by maintainers:
https://github.com/intelowlproject/IntelOwl/blob/develop/api_app/user_events_manager/queryset.py#L23

mlodic:
To have this issue accepted you need to provide proof that the performance is actually a problem by creating a test and measure the time. Otherwise this won't be accepted

mannubaveja007:
> To have this issue accepted you need to provide proof that the performance is actually a problem by creating a test and measure the time. Otherwise this won't be accepted

Sure

mannubaveja007:
Hi @mlodic I am trying to write the benchmark test using Django CaptureQueriesContext to measure query count and more... but it requires docker to run this project as i am on mac os having some hard time while using it on mac. can't we have an macOS option too? i know you have already told to me we don't support mac...

ayushgupta704:
Hey @mannubaveja007,
Totally get the Docker on macOS struggle it can be a real nightmare for local dev.
Since you're blocked on the environment, I managed to run a benchmark directly inside the intelowl_uwsgi container on a Linux to get the proof requested. You're spot on about the performance hit: for $N=100$ decaying events, the current implementation is hitting 301 database queries (1 SELECT + 100 related DataModel fetches + 100 DataModel UPDATES + 100
UserEvent UPDATES).

[...]

PR #3506, opened 2026-03-21
Optimized UserEvent.decay to resolve N+1 queries. 

ayushgupta704:
# Description
This PR resolves the scalability bottleneck in UserEventQuerySet.decay(). Previously, the decay process followed a 3N+1 query pattern, triggering sequential database round-trips for every event in the loop (fetching the DataModel, updating its reliability, and saving the UserEvent).
For a batch of 100 events, this resulted in 301 database queries. With this optimization, the same workload now takes only 6 queries O(1) constant time).

Key Architectural Changes:
   1. Batch Fetching: Implemented conditional select_related/prefetch_related on the data_model
      relationship to collapse $N$ fetch queries into 1.
   2. Bulk Updates: Replaced row-level .save() calls with Django's bulk_update().
   3. Deduplication: Added logic to deduplicate DataModel updates, ensuring each record is updated only
      once per batch with its final reliability score.
   4. Transactional Integrity: Wrapped the entire operation in transaction.atomic() to prevent
      inconsistent states during partial failures.
   5. Regression Testing: Added a new test case test_decay_performance that asserts the query count remains $\leq 10$ regardless of the number of events.

  Type of change

 # Checklist

   - [x] I have read and understood the rules about [how to Contribute
     (https://intelowlproject.github.io/docs/IntelOwl/contribute/) to this project
   - [x] I have inserted the copyright banner at the start of the file.
   - [x] Linters (Ruff) gave 0 errors.

### DIFF
--- a/api_app/user_events_manager/queryset.py
+++ b/api_app/user_events_manager/queryset.py
@@ -62,8 +62,9 @@ class UserEventQuerySet(QuerySet):
         # Atomic so partial failures don't leave inconsistent state.
         with transaction.atomic():
             for model_class, models_list in data_models_by_class.items():
-                model_class.objects.bulk_update(models_list, ["reliability"])
-            self.model.objects.bulk_update(events, ["decay_times", "next_decay"])
+                unique_models = {m.pk: m for m in models_list}.values()
+                model_class.objects.bulk_update(unique_models, ["reliability"], batch_size=1000)
+            self.model.objects.bulk_update(events, ["decay_times", "next_decay"], batch_size=1000)
 
         return len(events)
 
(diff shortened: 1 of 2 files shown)


ayushgupta704:
Hey @mlodic, I've updated this PR to be a single clean commit on top of the latest develop.

- In the Before screenshot, you can see how processing just 100 events was hitting the database 301 times it was a classic 3N+1 sequential loop that would eventually cause major lag as the number of observables grows.

Image

- I’ve implemented the fix using set-based operations, and the benchmark results confirm a massive improvement. I ran a fresh test with 100 events, and the query count dropped from 301 queries to just 6.

Image

hopefully, this helps keep things running smoothly as the platform scales! Let me know if you have any feedback or if you'd like me to adjust the implementation Ready for review Thanks.
closes #3460

PR #3508, opened 2026-03-22
fix: remove TalosReputation analyzer

### LINKED ISSUE #3451: Deprecate and remove TalosReputation analyzer

Fikri-20:
Closes #3451

The upstream URL (snort.org/downloads/ip-block-list) now returns an HTML terms page instead of the IP blocklist, making this analyzer permanently broken. Removes the analyzer, its unit test, the visualizer reference, and adds a data migration to clean up the DB entry.

### DIFF
--- a/api_app/analyzers_manager/migrations/0176_remove_talosreputation.py
+++ b/api_app/analyzers_manager/migrations/0176_remove_talosreputation.py
@@ -21,6 +21,7 @@ def reverse_migrate(apps, schema_editor):
 
 class Migration(migrations.Migration):
     dependencies = [
+        ("api_app", "0071_delete_last_elastic_report"),
         ("analyzers_manager", "0175_analyzer_config_cleanbrowsing_malicious_detector"),
     ]
 


Fikri-20:
Hey @mlodic 
Could you review the PR and tell me if there is another thing I could do to resolve this issue?

mlodic:
read the guidelines

PR #3516, opened 2026-03-23
fix: use Django's built-in password validators instead of custom regex. Closes #3385

### LINKED ISSUE #3385: Bug: REGEX_PASSWORD allows weak passwords

HarshitPal25:
# Description

This PR replaces the custom `REGEX_PASSWORD` pattern with Django's built-in password validation system as suggested by @mlodic in issue #3385.

## Problem
The current custom regex `r"^[a-zA-Z0-9]{12,}$"` has two main security issues:
- **Too permissive**: Accepts weak passwords like `aaaaaaaaaaaa` (all same characters)
- **Too restrictive**: Blocks special characters entirely, preventing stronger passwords like `MyP@ssw0rd!`

## Solution
Replaced custom regex validation with Django's `django.contrib.auth.password_validation.validate_password`, which leverages the configured `AUTH_PASSWORD_VALIDATORS`:
- `UserAttributeSimilarityValidator` - Prevents passwords too similar to username/email
- `MinimumLengthValidator` - Enforces 12-character minimum (maintains existing policy)
- `CommonPasswordValidator` - Rejects commonly used passwords
- `NumericPasswordValidator` - Prevents purely numeric passwords

## Changes Made
- **`intel_owl/consts.py`**: Removed `REGEX_PASSWORD`, replaced `validate_password_strength()` with Django's validator
- **`intel_owl/settings/auth.py`**: Set `MinimumLengthValidator` to 12 characters
- **`authentication/serializers.py`**: Pass user context to validators during registration for similarity checks
- **`authentication/views.py`**: Pass user context in `ChangePasswordView`, added null check for new password
- **`tests/auth/__init__.py`**: Updated test user password to meet new validation requirements
- **`tests/auth/test_auth.py`**: Updated test expectations (special chars now allowed), added password reset in setUp, added new validator tests

## Type of change

- [ ] Chore (refactoring, dependency updates, CI/CD changes, code cleanup, docs-only changes).

# Checklist

- [ ] A new plugin (analyzer, connector, visualizer, playbook, pivot or ingestor) was added or changed (N/A - no plugins modified)
- [x] I have inserted the copyright banner at the start of the file (already present in all modified files)
- [x] Please avoid adding new libraries as requirements whenever it is possible (no new libraries added)
- [ ] If external libraries/packages with restrictive licenses were added (N/A)
- [x] Linters (`Ruff`) gave 0 errors (pre-commit hooks passed)
- [x] I have added tests for the feature/bug I solved - All 15 auth tests pass
- [ ] If the GUI has been modified (N/A)
- [ ] After you had submitted the PR, if `DeepSource`, `Django Doctors` or other third-party linters have triggered any alerts during the CI checks, I have solved those alerts (will address if any arise)
- [x] I have addressed raised Copilot issues (will address if any arise)
- [ ] I have reviewed and verified any LLM-generated code included in this PR. Also, I have explicitly stated that I have used LLMs in this PR.

## Testing
All 15 authentication tests pass:
- `test_change_password_200` - Password change with valid password
-  `test_change_password_special_chars_200` - Special characters now accepted
-  `test_change_password_weak_password_400` - Weak passwords rejected
-  `test_common_password_400` - Common passwords rejected
-  `test_password_similar_to_username_400` - Similar passwords rejected
-  `test_min_password_lenght_400` - Short passwords rejected
-  `test_special_characters_password_201` - Special chars accepted in registration
- All existing tests continue to pass

### DIFF
--- a/authentication/serializers.py
+++ b/authentication/serializers.py
@@ -193,7 +193,7 @@ class RegistrationSerializer(rest_email_auth.serializers.RegistrationSerializer)
 
     def validate_password(self, password):
         """
-        Validate the user's password against a regex pattern.
+        Validate the user's password using Django's built-in validators.
 
         Args:
             password (str): The password to validate.
@@ -202,10 +202,18 @@ class RegistrationSerializer(rest_email_auth.serializers.RegistrationSerializer)
             str: The validated password.
 
         Raises:
-            ValidationError: If the password does not match the regex pattern.
+            ValidationError: If the password fails validation.
         """
         super().validate_password(password)
-        validate_password_strength(password)
+        # Create a temporary user instance with initial data for UserAttributeSimilarityValidator
+        # This allows checking if password is too similar to username, email, etc.
+        temp_user = User(
+            username=self.initial_data.get("username", ""),
+            email=self.initial_data.get("email", ""),
+            first_name=self.initial_data.get("first_name", ""),
+            last_name=self.initial_data.get("last_name", ""),
+        )
+        validate_password_strength(password, user=temp_user)
         return password
 
     def create(self, validated_data):
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -183,14 +183,18 @@ class ChangePasswordView(APIView):
         old_password = request.data.get("old_password")
         new_password = request.data.get("new_password")
 
-        # Validate new password strength
+        # Check if new password is provided
+        if not new_password:
+            return Response({"error": "New password is required"}, status=status.HTTP_400_BAD_REQUEST)
+
+        # Validate new password strength with user context for UserAttributeSimilarityValidator
+        user = request.user
         try:
-            validate_password_strength(new_password)
+            validate_password_strength(new_password, user=user)
         except ValidationError as e:
-            return Response({"error": e.message}, status=status.HTTP_400_BAD_REQUEST)
+            return Response({"error": str(e.message)}, status=status.HTTP_400_BAD_REQUEST)
 
         # Check if the old password matches the user's current password
-        user = request.user
         uname = user.username
         if not check_password(old_password, user.password):
             logger.info(f"'{uname}' has inputted invalid old password.")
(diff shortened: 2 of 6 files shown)


HarshitPal25:
@mlodic check this pr i have tests all the cases before making a pr

HarshitPal25:
@mlodic hey check this out and merge if its right please

mlodic:
you have being pushy and you have never been assigned to this issue

PR #3538, opened 2026-03-24
fix: normalize hash type in MISP connector. Closes #3537

### LINKED ISSUE #3537: str.replace() result discarded in _base_attr_obj — hash type never normalized
In api_app/connectors_manager/connectors/misp.py, inside the _base_attr_obj property, str.replace() is called on matched_type but its return value is never assigned. Since strings are immutable in Python, the result is silently discarded and matched_type remains unchanged.
Affected code
python
```
matched_type = helpers.get_hash_type(value)
matched_type.replace("-", "")  # convert sha-x to shax
_type = matched_type if matched_type is not None else "text"
```
**Expected behavior**
A hash like sha-256 should be normalized to sha256 before being passed as the MISP attribute type, since MISP expects the type without the hyphen.
**Actual behavior**
The hyphen is never removed. _type remains sha-256, which MISP does not recognize as a valid attribute type, causing the attribute creation to fail.
Fix
python
```
matched_type = helpers.get_hash_type(value)
_type = matched_type.replace("-", "") if matched_type is not None else "text"
```

jagapathi20:
# Description

Closes #3537

In `_base_attr_obj`, `str.replace()` was called but its return value was 
never assigned. Since strings are immutable in Python, `matched_type` 
remained unchanged — so a hash like `sha-256` was never converted to 
`sha256`, which is the format MISP expects as a valid attribute type.

## Type of change

# Checklist

- [x] I have read and understood the rules about [how to Contribute]
- [x] Linters (`Ruff`) gave 0 errors. If you have correctly installed [pre-commit]

### DIFF
--- a/api_app/connectors_manager/connectors/misp.py
+++ b/api_app/connectors_manager/connectors/misp.py
@@ -62,8 +62,8 @@ class MISP(Connector):
             value = self._job.analyzable.name
             if _type == Classification.HASH:
                 matched_type = helpers.get_hash_type(value)
-                matched_type.replace("-", "")  # convert sha-x to shax
-                _type = matched_type if matched_type is not None else "text"
+                # convert sha-x to shax
+                _type = matched_type.replace("-", "") if matched_type is not None else "text"
             else:
                 _type = INTELOWL_MISP_TYPE_MAP[_type]
 


jagapathi20:
@mlodic can you check this

PR #3546, opened 2026-03-25
fix: avoid double call to _base_attr_obj in MISP connector. Closes #3545

### LINKED ISSUE #3545: `_base_attr_obj` property called twice in `run()` causing double DB query
## What happened
In `api_app/connectors_manager/connectors/misp.py`, the `_base_attr_obj` 
property is called twice inside `run()`:
```python
attributes = [
    self._base_attr_obj,  # 1st call — DB hit
    ...
]
event.info += f": {self._base_attr_obj.value}"  # 2nd call — another DB hit
```

Since `_base_attr_obj` is a `@property`, each access rebuilds the object 
and hits the database via `.analyzers_to_execute.all()`. The second call's 
result is immediately discarded after reading `.value`.

`_base_attr_obj` should be called once, stored in a local variable, and 
reused.

Trigger the MISP connector on any hash observable and observe two identical 
DB queries fired for `analyzers_to_execute`.

jagapathi20:
# Description
Closes #3545

`_base_attr_obj` is a `@property` that hits the DB via 
`.analyzers_to_execute.all()` each time it is accessed. In `run()` it was 
called twice — once when building the attributes list and again to append 
`.value` to `event.info`. The second object was immediately discarded.

Fixed by storing the result in a local variable and reusing it.

## Type of change

# Checklist

(https://intelowlproject.github.io/docs/IntelOwl/contribute/#how-to-share-your-plugin-with-the-community))

### DIFF
--- a/api_app/connectors_manager/connectors/misp.py
+++ b/api_app/connectors_manager/connectors/misp.py
@@ -109,14 +109,15 @@ class MISP(Connector):
 
         # get event and attributes
         event = self._event_obj
+        base_attr = self._base_attr_obj
         attributes = [
-            self._base_attr_obj,
+            base_attr,
             *self._secondary_attr_objs,
             self._link_attr_obj,
         ]
 
         # append attribute name to event info
-        event.info += f": {self._base_attr_obj.value}"
+        event.info += f": {base_attr.value}"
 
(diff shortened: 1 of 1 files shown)


mlodic:
not needed

PR #3577, opened 2026-03-30
fix: resolve list parsing error in passive_dns visualizer

kishoreeducation045-sudo:
## Description
While testing the visualizers for my GSoC 2026 proposal (Self-Deployed LLM Chatbot), I encountered an edge case where the `Passive_DNS` visualizer crashes with a `'list' object has no attribute 'keys'` error. 

This happens when the external API returns a JSON array instead of a JSON object. This PR adds a defensive type guard to check if the incoming JSON is a list or a dict, preventing the `.keys()` method from failing and crashing the React frontend render.

## Steps to Reproduce
1. Run a scan on an IP that returns a list format for Passive DNS (e.g., `159.65.101.136`).
2. Attempt to view the `Passive_DNS` visualizer in the UI.

## Changes Made
- Added type validation before calling `.keys()` on the response data.
- Ensured consistent data shape is passed to the frontend.

## Here are some images of the error.
Screenshot 2026-03-30 220435
Screenshot 2026-03-30 220354


sanjib2006:
This issue has already been resolved in #3384.
Check the `develop` branch.

PR #3587, opened 2026-03-31
test(visualizers): add regression tests for passive_dns type guard coverage

kishoreeducation045-sudo:
## 📝 Description

While deploying the IntelOwl stack locally to map analyzer data structures for my GSoC 2026 proposal (Self-Deployed LLM Chatbot), I encountered a frontend rendering crash (`'list' object has no attribute 'keys'`) caused by unexpected API response types in the `Passive_DNS` visualizers. 

I noticed that **PR #3384** was recently merged into `develop` to fix this exact issue by adding robust `isinstance()` type guards across the Passive DNS extractors. However, looking at the test suite, these new guard paths were left uncovered (existing tests only validated `no_report`, `empty_report`, and `all_data_report`).

To ensure these new type guards do not regress in the future, I have added 5 targeted regression tests.

## ✅ Changes Made / Edge Cases Covered

Added test coverage for the following new `isinstance()` branches:

**TestRobtex:**
- [x] `test_dict_report_is_not_list`: Validates that a bare `dict` report returns `[]` instead of throwing an error.
- [x] `test_mixed_items_in_report_list`: Validates that lists containing mixed/invalid types (e.g., `[]`, `"string"`, `42`) correctly filter and only extract valid dictionaries.
- [x] `test_production_crash_scenario`: Directly simulates the exact production crash payload (`[{valid_dict}, []]`) to ensure the `isinstance(report, dict)` guard holds.

**TestMnemonicPDNS:**
- [x] `test_dict_report_extracts_data`: Validates the new path where the report is a dictionary containing a `"data"` key holding the records.

**TestCIRCLPassiveDNS:**
- [x] `test_string_report_returns_empty`: Ensures the top-level `isinstance(report, list)` guard safely returns `[]` if the API returns a string.

## 🔗 Related PRs
- Secures the fix introduced in #3384.

*(Note: Discovered and tested locally while finalizing my GSoC 2026 architecture proposal!)*

image

### DIFF
--- a/tests/api_app/visualizers_manager/passive_dns/test_analyzer_extractor.py
+++ b/tests/api_app/visualizers_manager/passive_dns/test_analyzer_extractor.py
@@ -448,6 +448,86 @@ class TestRobtex(CustomTestCase):
             report,
         )
 
+    def test_dict_report_not_list(self):
+        """When report is a bare dict (not a list), the isinstance(list) guard
+        should cause the function to return an empty list."""
+        self.robtex_report = AnalyzerReport.objects.create(
+            parameters={},
+            report={
+                "rrdata": "mx.spamexperts.com",
+                "rrname": "test.com",
+                "rrtype": "MX",
+                "time_last": 1582215078,
+                "time_first": 1441363932,
+            },
+            job=self.job,
+            task_id=uuid(),
+            config=AnalyzerConfig.objects.get(name="Robtex"),
+        )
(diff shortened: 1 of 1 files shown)


mlodic:
ok but you have to open an issue and ask to be assigned before opening PRs otherwise they will be closed

kishoreeducation045-sudo:
Thank you for the merge, @mlodic! Understood completely on the workflow. I will make sure to open an issue and request assignment before opening any PRs moving forward. Appreciate the guidance!

PR #3589, opened 2026-03-31
fix(frontend): show JobIsRunningAlert only when job is active

### LINKED ISSUE #3586: JobIsRunningAlert shown on completed jobs inside JobInfoCard
## What happened
I was going through the **JobInfoCard** component and noticed something off with the condition that controls when the **JobIsRunningAlert** gets rendered.

In frontend/src/components/jobs/result/JobInfoCard.jsx around line 221, the check is:
`{Object.values(JobFinalStatuses).includes(job.status) && (
  <JobIsRunningAlert job={job} />
)}
`

This shows the running alert + the kill button when the job is in a final status (**like reported_without_fails**, **killed**, etc.), which doesn't make sense  it should be the other way around. If you look at how the same component is used in **JobOverview.jsx** (line ~412), it correctly checks **isRunningJob** before rendering it.

So basically:
- Job is running → expand info card → no running flow visible (wrong)
- Job is done → expand info card → running flow shows up (also wrong)

Pretty sure the condition just needs a **!** in front of it.

## Environment
1. OS: MacOS 
2. IntelOwl version: develop (latest)

The running alert should only show up when the job is still running, not after it's done. Something like.
`{!Object.values(JobFinalStatuses).includes(job.status) && (
  <JobIsRunningAlert job={job} />
)}
`

1. Start a scan with a playbook that has a few analyzers (so it takes some time).
2. Go to the job result page while it's still running.
3. Expand the job info card by clicking the arrow button.
4. You won't see the running flow diagram here even though the job is running.
5. Wait for the job to finish, then expand the card again the running flow shows up now even though the job is already done.

## Error messages and logs
Nothing crashes or throws errors, it's just the condition that's flipped. You can spot it directly in the code by comparing with **JobOverview.jsx.**

kishoreeducation045-sudo:
### What happened
Currently, the `JobIsRunningAlert` component inside `JobInfoCard.jsx` renders its "Running" flow diagram *only* when the job has reached a final state (like `reported_without_fails` or `killed`), and hides it while the job is actually running.

### What changed
Flipped the logic condition on line 221 from `includes()` to `!includes()`. The alert now correctly displays when the job is actively running and hides once the job hits a final status. This perfectly aligns with how the alert is rendered in `JobOverview.jsx`.

Fixes #3586 

## Motivation and Context (GSoC 2026)
Accurate UI state management is critical for analyst workflows. I am tackling this issue as part of my codebase familiarization for my GSoC 2026 proposal (Project #7: LLM Chatbot). The chatbot interface will heavily rely on accurate, asynchronous React component rendering based on underlying job states, so I wanted to ensure the core UI logic is tight before integrating the LangChain agent.

## Type of change
- Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] Linters (`npm run lint`) gave 0 errors

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "intelowl",
-  "version": "6.4.0",
+  "version": "6.5.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "intelowl",
-      "version": "6.4.0",
+      "version": "6.5.1",
       "dependencies": {
         "@certego/certego-ui": "^0.1.19",
         "@dagrejs/dagre": "^1.1.4",
@@ -19,6 +19,7 @@
         "date-fns-tz": "^3.2.0",
         "flag-icons": "^7.2.3",
         "formik": "^2.4.6",
+        "http-proxy-middleware": "^2.0.6",
         "js-cookie": "^3.0.5",
         "md5": "^2.3.0",
         "prop-types": "^15.8.1",
@@ -42,6 +43,7 @@
       "devDependencies": {
         "@babel/preset-env": "^7.25.8",
         "@babel/preset-react": "^7.25.7",
+        "@testing-library/dom": "^10.4.1",
         "@testing-library/jest-dom": "^6.6.3",
         "@testing-library/react": "^12.1.5",
         "@testing-library/user-event": "^14.5.2",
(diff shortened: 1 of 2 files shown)


kishoreeducation045-sudo:
Hey @mlodic, I noted your workflow feedback on my other PR today. I had already opened PR #3589 to address this specific UI bug, but I want to respect the process. Could I please be officially assigned to this issue?

PR #3605, opened 2026-04-01
fix(frontend): show JobIsRunningAlert only for active jobs (#3586)

### LINKED ISSUE #3586: JobIsRunningAlert shown on completed jobs inside JobInfoCard

Abhishek9639:
Fixes #3586

# Description
The `JobIsRunningAlert` (flow diagram + Kill button) was incorrectly displayed on completed jobs inside the `JobInfoCard` component. The condition at line 221 in `JobInfoCard.jsx` was missing a `!` (NOT operator), causing the alert to render for final statuses instead of active ones.

**Changes:**
- Added `!` to the condition: `!Object.values(JobFinalStatuses).includes(job.status)` now matches the pattern used in `JobOverview.jsx` and `JobIsRunningAlert.jsx`
- Updated test to use `"running"` status instead of `"reported_without_fails"` to match the corrected logic

## Type of change

# Checklist

- [x] I have a provided a screenshot of the result in the PR.

### DIFF
--- a/frontend/src/components/jobs/result/JobInfoCard.jsx
+++ b/frontend/src/components/jobs/result/JobInfoCard.jsx
@@ -218,7 +218,7 @@ export function JobInfoCard({ job }) {
               </ListGroupItem>
             ))}
           </ListGroup>
-          {Object.values(JobFinalStatuses).includes(job.status) && (
+          {!Object.values(JobFinalStatuses).includes(job.status) && (
             <div
               className="my-4 d-flex justify-content-center"
               style={{ width: "100%" }}
--- a/frontend/tests/components/jobs/result/JobInfoCard.test.jsx
+++ b/frontend/tests/components/jobs/result/JobInfoCard.test.jsx
@@ -75,7 +75,7 @@ describe("test JobInfoCard (job report)", () => {
             observable_classification: "domain",
             file_name: "",
             file_mimetype: "",
-            status: "reported_without_fails",
+            status: "running",
             runtime_configuration: {
               analyzers: {},
               connectors: {},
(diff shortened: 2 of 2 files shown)


Abhishek9639:
Greetings @mlodic,
Could you please review it when you get a chance? 
If any changes are needed, please let me know. I’ll take care of them.
Thanks!

https://github.com/user-attachments/assets/f282ee96-62c8-4292-bb44-a5ed30722fd1

mlodic:
Sorry but on a second thought this is intended behavior. I think it is important to show the steps of the analyses execution anyway.

PR #3620, opened 2026-04-03
share S3 cached files across analyzers. Closes #3459

### LINKED ISSUE #3459: [Performance] S3Boto3StorageWrapper caches file per-analyzer instead of sharing a single cached copy
## What happened
In `intel_owl/settings/storage.py`, the `S3Boto3StorageWrapper.retrieve()` method caches
downloaded S3 files under `MEDIA_ROOT/<analyzer>/<filename>`. This means the file
The file from S3 is downloaded and is stored separately for each process.

If 10 analyzers are set to run on the same sample, the file will be retrieved from the disk, written to the HDD or sent over the network 10 times.

## Environment
1. OS: Any
2. IntelOwl version: latest develop

The file should be cached once at MEDIA_ROOT/<filename> and then reused by all the analyzers.
avoiding redundant S3 HTTP requests and duplicate disk writes.

1. Set `LOCAL_STORAGE=False` (S3 mode in env_file_app)
2. Upload a file and run multiple analyzers on it
3. Inspect MEDIA_ROOT/ — a separate subdirectory per analyzer is created with identical file content.

## Error messages and logs
No error. This is a known FIXME left by maintainers:
https://github.com/intelowlproject/IntelOwl/blob/develop/intel_owl/settings/storage.py#L37

mannubaveja007:
# Description

Closes #3459.

In the past, the `S3Boto3StorageWrapper` would download the same file from S3 several times, once for each analyzer that was running on it, and then save multiple copies of the file in different folders, which was unnecessary and redundant.

I've made some changes to the `retrieve()` function so that it stores a copy of the file in the shared `MEDIA_ROOT` folder. This way, when the first analyzer needs the file, it downloads it and the others can just use that same copy. This should really cut down on unnecessary internet requests and free up some disk space.

## Type of change

- [x] Fixing a bug, which is a small change that solves a problem, without affecting other parts of the system.

# Checklist

I've read and understood the rules for contributing to this project, which are outlined in the [how to Contribute](https://intelowlproject.github.io/docs/IntelOwl/contribute/) section.
The pull request is intended for the develop branch.
I have added the copyright notice at the beginning of the file: ```# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl # See the file 'LICENSE' for copying permission.```
Try to use what's already available instead of adding new libraries whenever possible.
If external libraries or packages with restrictive licenses are used, you can find them listed in the [Legal Notice](https://github.com/certego/IntelOwl/blob/master/.github/legal_notice.md) section. This is where we keep track of any external components that have limitations on how they can be used.
I've added tests to cover the feature or bug I've fixed - you can find them in the `tests` folder. When I ran all the tests, including the new and existing ones, I didn't get any errors.

### DIFF
--- a/api_app/analyzers_manager/classes.py
+++ b/api_app/analyzers_manager/classes.py
@@ -241,17 +241,9 @@ class FileAnalyzer(BaseAnalyzerMixin, metaclass=ABCMeta):
 
     def after_run(self):
         super().after_run()
-        # We delete the file only if we have single copy for analyzer
-        # and the file has been saved locally.
-        # Otherwise we would remove the single file that we have on the server
-        if not settings.LOCAL_STORAGE and self.filepath is not None:
-            import os
-
-            try:
-                os.remove(self.filepath)
-            except OSError:
-                logger.warning(f"Filepath {self.filepath} does not exists")
-
+        # When using S3 storage, cached files are now stored in a shared
+        # directory and reused by all analyzers, so we must NOT delete them
+        # here — another analyzer may still be reading the same file.
         logger.info(f"FINISHED analyzer: {self.__repr__()} -> File: ({self.filename}, md5: {self.md5})")
 
 
--- a/intel_owl/settings/storage.py
+++ b/intel_owl/settings/storage.py
@@ -33,25 +33,26 @@ else:
     from storages.backends.s3boto3 import S3Boto3Storage
 
     class S3Boto3StorageWrapper(S3Boto3Storage):
-        def retrieve(self, file, analyzer):
-            # FIXME we can optimize this a lot.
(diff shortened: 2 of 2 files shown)


mannubaveja007:
@mlodic can you check this? It's been a long time.
Thank You!

mlodic:
I don't get how this actually solve the issue

mannubaveja007:
> I don't get how this actually solve the issue

I just checked that simply removing the analyzer subdirectory doesn't work because the cleanup logic in after_run() deletes the file after each analyzer finishes.

Working on this Issue PR will be raised ASAP!

mannubaveja007:
it have some new commits can anyone review please?

PR #3644, opened 2026-04-09
fix(validin): add missing continue statement in _run_all_queries

### LINKED ISSUE #3642: Bug: Missing `continue` in `_run_all_queries` causes Validin analyzer to crash on HTTP errors
## What happened
While reviewing the code for the `Validin` analyzer in `api_app/analyzers_manager/observable_analyzers/validin.py`, I noticed a logical flaw in the `_run_all_queries` method. 

The method runs a loop over multiple endpoints and has a comment explicitly stating: `# we wont stop other quries from executing if one fails`. However, the code just logs the error when `response.status_code != 200` but forgets to actually skip the parsing step with a `continue` block. 

As a result, it blindly attempts to do `final_response[f"{query_name}"] = response.json()` anyway. If the API returns a generic non-JSON error page (which happens frequently during 5xx or 404 errors), it throws a `JSONDecodeError`. This completely breaks the loop and aborts the entire analyzer, completely violating the intended design of ignoring the failed query and continuing with the others.

## Environment
1. OS: Any (Codebase logic issue)
2. IntelOwl version: Main/Develop branch

If a query fails with a non-200 status code, I expected the code to log the error and then use a `continue` statement to jump to the next iteration of the loop, skipping the `.json()` parsing so that the entire analyzer doesn't crash.

1. Run a job using the `Validin` analyzer.
2. If any of the queried Validin endpoints returns a non-200 HTTP code containing an HTML error instead of JSON, the entire analyzer aborts due to a JSON decoding error, rather than recovering and continuing with the rest of the endpoints.

## Error messages and logs
The problematic code snippet in `validin.py`:

```python
if response.status_code != 200:
logger.error(f"Query {query_name} failed")

# we wont stop other quries from executing if one fails
final_response[f"{query_name}"] = response.json() # <-- Causes JSONDecodeError due to missing continue statement

Abhishek9639:
Closes #3642. 
If your PR is made by a single commit, please add that clause in the commit too.

# Description

Added a missing `continue` statement in the `_run_all_queries` method of the Validin analyzer. 

When an API endpoint returned a non-200 status code, the error was logged but execution continued to `response.json()`, which would throw a `JSONDecodeError` if the response body was HTML instead of JSON. This broke the entire loop and crashed the analyzer, violating the intended behavior described in the existing comment: `# we wont stop other quries from executing if one fails`.

The fix simply adds `continue` after the error log so the loop skips to the next endpoint gracefully.

## Type of change

# Checklist

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/validin.py
+++ b/api_app/analyzers_manager/observable_analyzers/validin.py
@@ -36,6 +36,7 @@ class Validin(classes.ObservableAnalyzer):
                         logger.error(f"Query {query_name} failed")
 
                         # we wont stop other quries from executing if one fails
+                        continue
                     final_response[f"{query_name}"] = response.json()
                 except requests.RequestException as e:
                     raise AnalyzerRunException(e)


Abhishek9639:
Greetings @mlodic,

This PR fixes #3642 a missing `continue` statement in the `_run_all_queries` method of the Validin analyzer.

Without the `continue`, when an endpoint returns a non-200 response, the code falls through to `response.json()` which crashes with a `JSONDecodeError` if the response body is HTML. This breaks the entire loop even though the existing comment says otherwise.

It's a small one-line fix. Happy to make any changes if needed.

Thanks

PR #3661, opened 2026-04-22
fix: set job to FAILED on exception in job_pipeline

### LINKED ISSUE #3653: [BUG] job_pipeline task does not set job's final status on exception jobs get stuck in RUNNING forever
## What happened

While reading through the `job_pipeline` task in `intel_owl/tasks.py` (lines 247-266), I noticed that when `job.execute()` throws an exception, the `except` block marks individual plugin reports as `FAILED` but it never actually updates the **Job object** itself.

Here's the relevant code:

```
@shared_task(base=FailureLoggedTask, name="job_pipeline", soft_time_limit=100)
def job_pipeline(job_id: int):
    from api_app.models import Job

    job = Job.objects.get(pk=job_id)
    try:
        job.execute()
    except Exception as e:
        logger.exception(e)
        for report in (
            list(job.analyzerreports.all())
            + list(job.connectorreports.all())
            + list(job.pivotreports.all())
            + list(job.visualizerreports.all())
        ):
            report.status = report.STATUSES.FAILED.value
            report.save()
```

The issue is inside Job.execute() (line 714 in models.py), the very first thing it does is set self.status = self.STATUSES.RUNNING and save it. So by the time an exception is raised (say during _get_pipeline() or _get_signatures() or if the broker is temporarily down), the job is already marked as RUNNING in the DB.

But the except block above doesn't set it back to FAILED. It also doesn't:

- Set job.finished_analysis_time
- Send a WebSocket notification via JobConsumer.serialize_and_send_job(job)
- Update the parent Investigation status

So the job just stays stuck in RUNNING forever. The job_set_final_status task (line 224) is supposed to handle this cleanup, but it's chained as the last step of the Celery pipeline if the pipeline never starts, that task never runs.

I noticed there's already a check_stuck_analysis periodic task (line 124) that catches these after ~25 minutes and marks them as failed, but that's more of a safety net. It also doesn't send any WebSocket notifications, so the user still sees a stuck spinner until they manually refresh.

Worth noting: the run_plugin task (line 268) does call JobConsumer.serialize_and_send_job(job) after exception handling, so there's already a pattern for this in the codebase job_pipeline just doesn't follow it.

## Environment

1. OS: Ubuntu (Docker)
2. IntelOwl version: develop (v6.6.0)

If job.execute() fails, the except block should also clean up the job state properly. Something like:

1. Set job.status to FAILED
2. Append the error to job.errors
3. Set job.finished_analysis_time = now()
4. Call JobConsumer.serialize_and_send_job(job) so the frontend knows right away
5. Update the Investigation status if the job belongs to one

Basically the same stuff job_set_final_status would have done if the pipeline had completed normally.

1. Cause an exception during job.execute() for example a DB error in _get_signatures(), a broken PythonConfig import, or even just the broker being momentarily unreachable when runner() is called
2. The job will be stuck as RUNNING in the database indefinitely
3. Frontend shows an infinite spinner with no error feedback
4. Eventually check_stuck_analysis cleans it up after ~25 min, but without notifying the frontend

You can also just read the code to confirm:

- intel_owl/tasks.py line 247-266 → the incomplete except block
- intel_owl/tasks.py line 224-232 → job_set_final_status that does the proper cleanup
- api_app/models.py line 714-715 → where execute() sets status to RUNNING before building the pipeline

## Error messages and logs
No error on the frontend that's the whole problem. The user just sees a job stuck in "Running" forever. Server-side, the exception gets logged via logger.exception(e) but from the user's perspective nothing happens.

DB state of a stuck job:

- job.status = running (never reaches a terminal state)
- job.finished_analysis_time = NULL
- Because finished_analysis_time is NULL, remove_old_jobs won't clean these up either they just accumulate

Abhishek9639:
Closes #3653

# Description
When `job.execute()` throws an exception inside `job_pipeline`, the except block only marks individual plugin reports as FAILED but never updates the Job object itself. Since `execute()` sets `status = RUNNING` before building the pipeline, any failure after that point leaves the job stuck in RUNNING forever.

This adds the missing cleanup to the except block setting the job status to FAILED, recording the error, setting `finished_analysis_time`, updating the parent investigation, and sending a WebSocket notification. This mirrors the same pattern already used in `run_plugin` and `set_final_status`.

## Type of change

# Checklist
- [x] I have read and understood the rules about [how to Contribute](https://intelowlproject.github.io/docs/IntelOwl/contribute) to this project
- [x] The pull request is for the branch `develop`.

### DIFF
--- a/intel_owl/tasks.py
+++ b/intel_owl/tasks.py
@@ -248,6 +248,7 @@ def job_pipeline(
     job_id: int,
 ):
     from api_app.models import Job
+    from api_app.websocket import JobConsumer
 
     job = Job.objects.get(pk=job_id)
     try:
@@ -262,6 +263,14 @@ def job_pipeline(
         ):
             report.status = report.STATUSES.FAILED.value
             report.save()
+        # mark job as failed and notify; job_set_final_status won't run since the pipeline never started
+        job.status = Job.STATUSES.FAILED.value
+        job.errors.append(str(e))
+        job.finished_analysis_time = now()
+        job.save(update_fields=["status", "errors", "finished_analysis_time"])
+        if root_investigation := job.get_root().investigation:
+            root_investigation.set_correct_status(save=True)
+        JobConsumer.serialize_and_send_job(job)
 
 
 @shared_task(base=FailureLoggedTask, name="run_plugin", soft_time_limit=500)
--- a/tests/test_crons.py
+++ b/tests/test_crons.py
@@ -19,7 +19,7 @@ from api_app.analyzers_manager.observable_analyzers import (
 )
 from api_app.choices import Classification
 from api_app.models import Job
-from intel_owl.tasks import check_stuck_analysis, remove_old_jobs
+from intel_owl.tasks import check_stuck_analysis, job_pipeline, remove_old_jobs
 
 from . import CustomTestCase, get_logger
 from .mock_utils import MockUpResponse, if_mock_connections, patch, skip
@@ -55,6 +55,44 @@ class CronTests(CustomTestCase):
         _job.status = Job.STATUSES.ANALYZERS_RUNNING.value
         _job.save()
         self.assertCountEqual(check_stuck_analysis(check_pending=False), [_job.pk])
+        _job.delete()
+        an.delete()
+
+    def test_job_pipeline_exception_sets_job_to_failed(self):
+        """
+        Regression test for GitHub issue #3653.
+        When job.execute() raises an exception inside job_pipeline,
+        the Job object must be marked as FAILED (not left stuck in RUNNING),
+        finished_analysis_time must be set, and the error must be recorded.
+        """
+        an = Analyzable.objects.create(
+            name="8.8.8.8",
+            classification=Classification.IP,
+        )
+        _job = Job.objects.create(
+            user=self.user,
+            status=Job.STATUSES.PENDING.value,
+            analyzable=an,
+        )
+        error_message = "Simulated broker failure"
+        with (
+            patch.object(
+                _job.__class__,
+                "execute",
+                side_effect=Exception(error_message),
+            ),
+            patch("api_app.websocket.JobConsumer.serialize_and_send_job"),
+            patch(
+                "api_app.models.Job.objects.get",
+                return_value=_job,
+            ),
+        ):
+            job_pipeline(_job.pk)
+
+        _job.refresh_from_db()
+        self.assertEqual(_job.status, Job.STATUSES.FAILED.value)
+        self.assertIsNotNone(_job.finished_analysis_time)
+        self.assertIn(error_message, _job.errors)
         _job.delete()
         an.delete()
 


Abhishek9639:
Hi @mlodic,

This PR fixes the stuck job issue reported in #3653. The except block in `job_pipeline` was only marking plugin reports as failed but never updating the job itself so it stayed stuck in RUNNING with no WebSocket notification and a NULL `finished_analysis_time`.

I aligned the except block with the same cleanup pattern that `run_plugin` and `set_final_status` already follow. Added a regression test for it too.

Let me know if anything needs changing. 
Thanks

gks281263 (commented):
Hey @Abhishek9639, nice work getting this fix in — the core logic is correct and addresses the real gap in `job_pipeline`. I did a close review and have a few hardening suggestions below to make it production better. Overall the direction is exactly right and aligned with `run_plugin` / `set_final_status`.

intel_owl/tasks.py
```diff
@@ -262,6 +263,14 @@ def job_pipeline(
...
+        job.status = Job.STATUSES.FAILED.value
+        job.errors.append(str(e))
+        job.finished_analysis_time = now()
+        job.save(update_fields=["status", "errors", "finished_analysis_time"])
+        if root_investigation := job.get_root().investigation:
+            root_investigation.set_correct_status(save=True)
+        JobConsumer.serialize_and_send_job(job)
 
 
 @shared_task(base=FailureLoggedTask, name="run_plugin", soft_time_limit=500)
```
gks281263, line 276: **Suggestion: Wrap cleanup in its own try/except**

If any of these cleanup lines raise (e.g., `job.get_root()` hits a treebeard `MultipleObjectsReturned` edge case, or the WebSocket layer has a config issue), the original exception `e` gets swallowed and replaced with the cleanup exception. The original error would be lost.

Consider:
```python
except Exception as e:
    logger.exception(e)
    # ... existing report cleanup ...
    try:
        job.status = Job.STATUSES.FAILED.value
        job.errors.append(str(e))
        job.finished_analysis_time = now()
        job.save(update_fields=["status", "errors", "finished_analysis_time"])
        if root_investigation := job.get_root().investigation:
            root_investigation.set_correct_status(save=True)
        JobConsumer.serialize_and_send_job(job)
    except Exception as cleanup_err:
        logger.exception(
            f"Failed to clean up job {job_id} after pipeline exception: {cleanup_err}"
        )
```
This ensures the original error is always logged regardless of cleanup failures.

tests/test_crons.py
```diff
@@ -58,6 +58,44 @@ def test_check_stuck_analysis(self):
...
+                _job.__class__,
+                "execute",
+                side_effect=Exception(error_message),
+            ),
+            patch("api_app.websocket.JobConsumer.serialize_and_send_job"),
+            patch(
+                "api_app.models.Job.objects.get",
+                return_value=_job,
+            ),
+        ):
```
gks281263, line 89: **Unnecessary mock — this can be removed**

The `Job.objects.get` mock isn't needed here since `_job` already exists in the test DB, so the real `Job.objects.get(pk=_job.pk)` will work fine. Removing it makes the test more realistic and would catch regressions where someone accidentally switches from `.save()` to `.filter().update()` in the fix.

tests/test_crons.py
```diff
@@ -58,6 +58,44 @@ def test_check_stuck_analysis(self):
...
+            analyzable=an,
+        )
+        error_message = "Simulated broker failure"
+        with (
+            patch.object(
+                _job.__class__,
+                "execute",
+                side_effect=Exception(error_message),
+            ),
+            patch("api_app.websocket.JobConsumer.serialize_and_send_job"),
```
gks281263, line 84: **Missing assertion: WebSocket notification was actually sent**

You're patching `JobConsumer.serialize_and_send_job` here, but never asserting it was called. Since the whole point of this fix (from the user's perspective) is that the frontend stops showing an infinite spinner, this is worth verifying:

```python
with (
    patch.object(
        _job.__class__,
        "execute",
        side_effect=Exception(error_message),
    ),
    patch("api_app.websocket.JobConsumer.serialize_and_send_job") as mock_ws,
):
    job_pipeline(_job.pk)

mock_ws.assert_called_once()
```

tests/test_crons.py
```diff
@@ -58,6 +58,44 @@ def test_check_stuck_analysis(self):
         _job.delete()
         an.delete()
 
+    def test_job_pipeline_exception_sets_job_to_failed(self):
```
gks281263, line 61: **Suggestion: Add a second test case for the Investigation status update path**

The `root_investigation.set_correct_status(save=True)` branch in the fix is currently untested because this job has no parent investigation. Consider adding a `subTest` (or separate test) that creates a job attached to an `Investigation` and asserts the investigation status transitions correctly after the pipeline failure. That way both branches of the `if root_investigation` conditional are covered.

Abhishek9639:
Hi @mlodic,

I noticed the review comments on this PR and wanted to confirm with you before proceeding. Should I go ahead and implement the suggested changes (like removing the unnecessary mock, adding the assertion for WebSocket notification, and including an additional test case), or would you prefer a different approach?
Please let me know your preference so I can align accordingly. 
Thanks

Issue #3720, opened 2026-05-28
Implement LangChain ReAct agent with job tools and REST endpoint


berardifra:
## Goal
Add the core LLM chatbot functionality: a LangChain ReAct agent backed by Ollama with the first set of IntelOwl tools, conversation memory, and a synchronous REST endpoint.

## Deliverables
- `DjangoChatMessageHistory` — LangChain `BaseChatMessageHistory` backed by `ChatMessage` model
- Tools (ORM-direct, user-scoped):
  - `search_jobs` — filter jobs by observable name, MD5, status
  - `get_job_details` — full job detail with analyzer reports
  - `summarize_job` — human-readable job summary
- `build_agent_executor(user)` using `ChatOllama` + `create_react_agent` with custom ReAct prompt and chat history injection
- `ChatSessionViewSet` with `POST /api/chatbot/sessions/message` synchronous endpoint
- `process_chat_message` Celery task stub (full impl in W6)
- Unit tests for all tools and the message endpoint (LLM mocked in CI)

## Dependencies
Depends on #3712 (chatbot skeleton) and #3711 (Ollama docker).

PR #3722, opened 2026-05-28
[GSoC 2026] Add LangChain ReAct agent with job tools and REST endpoint (refs #3720)

berardifra:
Refs #3720.

# Description

Add the core LLM chatbot functionality: a LangChain ReAct agent backed by a local Ollama instance, the first three IntelOwl query tools, Django-backed conversation memory, and a synchronous REST endpoint.

- `DjangoChatMessageHistory`: LangChain `BaseChatMessageHistory` backed by the `ChatMessage` ORM model
- Three user-scoped tools (ORM-direct, multi-tenancy enforced via closure at build time):
  - `search_jobs` — filter by observable name / MD5 / status
  - `get_job_details` — full job with analyzer reports
  - `summarize_job` — human-readable summary
- `build_agent_executor(user)` using `ChatOllama` + `create_react_agent` with an inline ReAct prompt (no external hub dependency)
- `ChatSessionViewSet` with `POST /api/chatbot/sessions/message` synchronous REST endpoint
- `process_chat_message` Celery task stub (full async impl deferred to W6 WebSocket consumer)
- Chatbot URLs registered at `/api/chatbot/`

Note: this PR targets the GSoC umbrella branch `gsoc-2026/llm-chatbot` (not `develop` directly). The umbrella merges into `develop` at midterm evaluation.

## Type of change

# Checklist

- [ ] The pull request is for the branch `develop` _(targeting GSoC umbrella branch instead — see note above)_
- [x] Please avoid adding new libraries as requirements whenever it is possible. `langchain`, `langchain-ollama`, and `langchain-core` are strictly required for the LLM orchestration layer. Addition pre-approved by @mlodic.
- [x] I have reviewed and verified any LLM-generated code included in this PR. Also, I have explicitly stated that I have used LLMs in this PR: parts of this code were developed with AI assistance and have been reviewed, tested, and validated by me.

### DIFF
--- a/api_app/chatbot_manager/agent/memory.py
+++ b/api_app/chatbot_manager/agent/memory.py
@@ -6,13 +6,20 @@ from langchain_core.messages import AIMessage, BaseMessage, HumanMessage
 
 
 class DjangoChatMessageHistory(BaseChatMessageHistory):
-    """LangChain chat history backed by Django ChatMessage model."""
+    """LangChain chat history backed by the Django ChatMessage model.
+
+    Implements the BaseChatMessageHistory contract so a ChatSession's messages can be
+    used as an agent's conversation memory: `messages` reads, `add_message` writes
+    (via the inherited add_user_message/add_ai_message helpers), `clear` resets.
+    """
 
     def __init__(self, session):
+        """Bind this history to a single ChatSession (all reads/writes scope to it)."""
         self.session = session
 
     @property
     def messages(self) -> list[BaseMessage]:
+        """Return the session's messages, oldest first, as LangChain message objects."""
         from api_app.chatbot_manager.models import ChatMessage
 
         result = []
@@ -24,12 +31,14 @@ class DjangoChatMessageHistory(BaseChatMessageHistory):
         return result
 
     def add_message(self, message: BaseMessage) -> None:
+        """Persist one message, mapping its LangChain type to the ChatMessage role."""
         from api_app.chatbot_manager.models import ChatMessage
 
         role = ChatMessage.Role.USER if isinstance(message, HumanMessage) else ChatMessage.Role.ASSISTANT
         ChatMessage.objects.create(session=self.session, role=role, content=message.content)
 
     def clear(self) -> None:
+        """Delete every message in the session (resets the conversation)."""
         from api_app.chatbot_manager.models import ChatMessage
 
         ChatMessage.objects.filter(session=self.session).delete()
--- a/api_app/chatbot_manager/agent/memory.py
+++ b/api_app/chatbot_manager/agent/memory.py
@@ -1,0 +1,35 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+from langchain_core.chat_history import BaseChatMessageHistory
+from langchain_core.messages import AIMessage, BaseMessage, HumanMessage
+
+
+class DjangoChatMessageHistory(BaseChatMessageHistory):
+    """LangChain chat history backed by Django ChatMessage model."""
+
+    def __init__(self, session):
+        self.session = session
+
+    @property
+    def messages(self) -> list[BaseMessage]:
+        from api_app.chatbot_manager.models import ChatMessage
+
+        result = []
+        for msg in ChatMessage.objects.filter(session=self.session).order_by("timestamp"):
+            if msg.role == ChatMessage.Role.USER:
+                result.append(HumanMessage(content=msg.content))
+            else:
+                result.append(AIMessage(content=msg.content))
+        return result
+
+    def add_message(self, message: BaseMessage) -> None:
+        from api_app.chatbot_manager.models import ChatMessage
+
+        role = ChatMessage.Role.USER if isinstance(message, HumanMessage) else ChatMessage.Role.ASSISTANT
+        ChatMessage.objects.create(session=self.session, role=role, content=message.content)
+
+    def clear(self) -> None:
+        from api_app.chatbot_manager.models import ChatMessage
+
+        ChatMessage.objects.filter(session=self.session).delete()
--- a/api_app/chatbot_manager/agent/tools/get_job_details.py
+++ b/api_app/chatbot_manager/agent/tools/get_job_details.py
@@ -1,0 +1,60 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+import json
+
+
+def make_get_job_details_tool(user):
+    from langchain_core.tools import tool
+
+    @tool("get_job_details")
+    def get_job_details(job_id: int) -> str:
+        """Get full details of an IntelOwl job by its numeric ID.
+
+        Args:
+            job_id: The numeric ID of the job to retrieve.
+        """
+        from api_app.models import Job
+
+        try:
+            job = (
+                Job.objects.select_related("analyzable")
+                .prefetch_related("analyzer_reports", "analyzers_to_execute", "tags")
+                .get(pk=job_id, user=user)
+            )
+        except Job.DoesNotExist:
+            return f"Job with ID {job_id} not found or not accessible."
+
+        analyzer_reports = []
+        for report in job.analyzer_reports.all():
+            analyzer_reports.append(
+                {
+                    "name": report.config.name,
+                    "status": report.status,
+                    "report": report.report if isinstance(report.report, dict) else str(report.report)[:500],
+                }
+            )
+
+        return json.dumps(
+            {
+                "id": job.pk,
+                "observable_name": job.analyzable.name,
+                "observable_classification": job.analyzable.classification,
+                "md5": job.analyzable.md5,
+                "status": job.status,
+                "tlp": job.tlp,
+                "received_request_time": str(job.received_request_time),
+                "finished_analysis_time": str(job.finished_analysis_time)
+                if job.finished_analysis_time
+                else None,
+                "process_time": job.process_time,
+                "analyzers_to_execute": list(job.analyzers_to_execute.values_list("name", flat=True)),
+                "tags": list(job.tags.values_list("label", flat=True)),
+                "errors": job.errors,
+                "warnings": job.warnings,
+                "analyzer_reports": analyzer_reports,
+            },
+            indent=2,
+        )
+
+    return get_job_details
--- a/api_app/chatbot_manager/agent/tools/get_job_details.py
+++ b/api_app/chatbot_manager/agent/tools/get_job_details.py
@@ -20,53 +20,18 @@ def make_get_job_details_tool(user):
         Returns:
             JSON string with shape {"errors": [...], "job": {...} | null}.
         """
+        from api_app.chatbot_manager.serializers import JobDetailToolSerializer
         from api_app.models import Job
 
         try:
             job = (
                 Job.objects.select_related("analyzable")
-                .prefetch_related("analyzer_reports__config", "analyzers_to_execute", "tags")
+                .prefetch_related("analyzerreports__config", "analyzers_to_execute", "tags")
                 .get(pk=job_id, user=user)
             )
         except Job.DoesNotExist:
             return json.dumps({"errors": [f"Job with ID {job_id} not found or not accessible."], "job": None})
 
-        analyzer_reports = []
-        for report in job.analyzer_reports.all():
-            # `report.report` is a JSONField (dict / list / scalar depending on the
-            # analyzer). Dicts pass through and get re-serialized below; anything else is
-            # stringified and truncated to cap the size fed into the LLM prompt.
-            analyzer_reports.append(
-                {
-                    "name": report.config.name,
-                    "status": report.status,
-                    "report": report.report if isinstance(report.report, dict) else str(report.report)[:500],
-                }
-            )
-
-        return json.dumps(
-            {
-                "errors": [],
-                "job": {
-                    "id": job.pk,
-                    "observable_name": job.analyzable.name,
-                    "observable_classification": job.analyzable.classification,
-                    "md5": job.analyzable.md5,
-                    "status": job.status,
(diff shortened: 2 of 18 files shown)


api_app/chatbot_manager/agent/tools/get_job_details.py
```diff
@@ -0,0 +1,60 @@
...
+        except Job.DoesNotExist:
+            return f"Job with ID {job_id} not found or not accessible."
+
+        analyzer_reports = []
+        for report in job.analyzer_reports.all():
+            analyzer_reports.append(
+                {
+                    "name": report.config.name,
+                    "status": report.status,
+                    "report": report.report if isinstance(report.report, dict) else str(report.report)[:500],
```
mlodic: why handling this as a string is needed?

api_app/chatbot_manager/agent/tools/get_job_details.py
```diff
@@ -0,0 +1,60 @@
...
+        from api_app.models import Job
+
+        try:
+            job = (
+                Job.objects.select_related("analyzable")
+                .prefetch_related("analyzer_reports__config", "analyzers_to_execute", "tags")
+                .get(pk=job_id, user=user)
+            )
+        except Job.DoesNotExist:
+            return f"Job with ID {job_id} not found or not accessible."
```
mlodic: I'd prefer to keep the output of such functions consistent. 
I would expect a dictionary like {"errors":..., "job":...}
and those are populated based on the retrieved results but those keys are always there.
Do you see anything against this?

api_app/chatbot_manager/agent/tools/summarize_job.py
```diff
@@ -0,0 +1,50 @@
...
+                .get(pk=job_id, user=user)
+            )
+        except Job.DoesNotExist:
+            return f"Job with ID {job_id} not found or not accessible."
+
+        analyzers = list(job.analyzers_to_execute.values_list("name", flat=True))
+        failed_reports = [
+            r.config.name
+            for r in job.analyzer_reports.all()
+            if r.status not in ("success", "reported_without_fails")
```
mlodic: instead of using raw strings, to keep everything consistent through the project, please reuse values from the class Status

mlodic: and to be more accurate, ReportStatus contains the possible statuses of an analyzer report, while Status contains the possible status of a job

api_app/chatbot_manager/tasks.py
```diff
@@ -0,0 +1,12 @@
...
+# See the file 'LICENSE' for copying permission.
+
+from celery import shared_task
+
+from intel_owl.tasks import FailureLoggedTask
+
+
+@shared_task(base=FailureLoggedTask, soft_time_limit=300)
+def process_chat_message(session_id: int, user_message: str, user_id: int) -> str:
+    # Stub — full async implementation added in W6 with WebSocket consumer.
```
mlodic: please add a "todo" market at the start of the comment, that is usually keep tracked by IDEs

api_app/chatbot_manager/tasks.py
```diff
@@ -0,0 +1,12 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+from celery import shared_task
+
+from intel_owl.tasks import FailureLoggedTask
+
+
+@shared_task(base=FailureLoggedTask, soft_time_limit=300)
```
mlodic, line 14: is 5 minutes chosen for a reason or just randomly? could you please leave a note about this, whatever is the reason if there is.
I am asking this cause celery task managements could be problematic if not set up properly so it is important to set this flags in the right way if possible

api_app/chatbot_manager/views.py
```diff
@@ -1,2 +1,72 @@
...
+    ChatSessionSerializer,
+    MessageRequestSerializer,
+    MessageResponseSerializer,
+)
+
+
+class ChatSessionViewSet(ModelViewSet):
+    serializer_class = ChatSessionSerializer
+    permission_classes = [IsAuthenticated]
+    http_method_names = ["get", "post", "delete", "head", "options"]
```
mlodic, line 24: are all necessary?

api_app/chatbot_manager/views.py
```diff
@@ -1,2 +1,72 @@
...
+
+    def get_queryset(self):
+        return ChatSession.objects.filter(user=self.request.user)
+
+    def perform_create(self, serializer):
+        serializer.save(user=self.request.user)
+
+    @action(detail=False, methods=["post"], url_path="message")
+    def message(self, request):
+        req = MessageRequestSerializer(data=request.data)
```
mlodic, line 42: please add a docstring here explaining how this works

api_app/chatbot_manager/agent/memory.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class DjangoChatMessageHistory(BaseChatMessageHistory):
+    """LangChain chat history backed by Django ChatMessage model."""
+
+    def __init__(self, session):
+        self.session = session
+
+    @property
+    def messages(self) -> list[BaseMessage]:
```
mlodic, line 21: even if obvious, please add docstrings to each methods too and not only to the class

api_app/chatbot_manager/agent/memory.py
```diff
@@ -0,0 +1,35 @@
...
+        return result
+
+    def add_message(self, message: BaseMessage) -> None:
+        from api_app.chatbot_manager.models import ChatMessage
+
+        role = ChatMessage.Role.USER if isinstance(message, HumanMessage) else ChatMessage.Role.ASSISTANT
+        ChatMessage.objects.create(session=self.session, role=role, content=message.content)
+
+    def clear(self) -> None:
+        from api_app.chatbot_manager.models import ChatMessage
```
mlodic, line 42: are this method and the previous one used anywhere now? what's the plan here?

api_app/chatbot_manager/agent/tools/get_job_details.py
mlodic: also, why JSON output is actually needed here? the ouput always need to be a string here? is the framework working in this way?

mlodic (commented):
hey, I left some questions. Broadly speaking, please assume that nothing is obvious. The structure should allow to understand the flow without any knowledge of the underline frameworks so any possible comments about the decisions made here is very useful, in particular for future maintainability

api_app/chatbot_manager/agent/tools/get_job_details.py
berardifra: `report.report` is a JSONField (dict/list/scalar). Dicts pass through; non-dicts are stringified and capped at 500 chars so a huge report can't bloat the LLM prompt. Added a comment.

api_app/chatbot_manager/agent/tools/get_job_details.py
berardifra: > I'd prefer to keep the output of such functions consistent. I would expect a dictionary like {"errors":..., "job":...} and those are populated based on the retrieved results but those keys are always there. Do you see anything against this?

No, agreed. Done on all three tools. Both keys are always present, including on the error path (which used to return a bare string).

berardifra: > also, why JSON output is actually needed here? the ouput always need to be a string here? is the framework working in this way?

Yes, the return is injected back into the prompt as the ReAct Observation, so it must be a string.` json.dumps` gives the model unambiguous structured data instead of a `str()`'d dict.

api_app/chatbot_manager/agent/tools/summarize_job.py
berardifra: Yep, done. You're right about the distinction, and it surfaced a bug: I was comparing the analyzer report status against `"success" `plus the job-level `"reported_without_fails"`, so the check never matched and every report was flagged as failed. Now it uses `ReportStatus.SUCCESS.value`, with a regression test

api_app/chatbot_manager/tasks.py
berardifra: Done

api_app/chatbot_manager/tasks.py
berardifra: Not random, added a note. A chat turn chains several ReAct iterations, each an Ollama inference on a local model, so it can take tens of seconds. It's a `soft_time_limit` on purpose: it raises a catchable `SoftTimeLimitExceeded` (lets the task clean up) instead of a hard `time_limit` that SIGKILLs the worker mid-run. It just bounds a hung Ollama call. Happy to add a hard ceiling too if you'd prefer one.

api_app/chatbot_manager/views.py
berardifra: It's a deny-by-default whitelist: it excludes `put`/`patch` (sessions aren't edited). `get`=list/retrieve, `post`=create + `message`, `delete`=delete; `head`/`options` are DRF defaults.

api_app/chatbot_manager/views.py
berardifra: Added a docstring describing the request flow.

api_app/chatbot_manager/agent/memory.py
berardifra: Done

api_app/chatbot_manager/agent/memory.py
berardifra: They're the `BaseChatMessageHistory` contract: `messages` reads (used), `add_message` writes, `clear` resets. I wired the view to persist via `add_user_message`/`add_ai_message`, so `add_message` is now used in W2; clear stays for session reset and the W6 async consumer.

api_app/chatbot_manager/agent/tools/get_job_details.py
```diff
@@ -0,0 +1,72 @@
...
+                    "received_request_time": str(job.received_request_time),
+                    "finished_analysis_time": str(job.finished_analysis_time)
+                    if job.finished_analysis_time
+                    else None,
+                    "process_time": job.process_time,
+                    "analyzers_to_execute": list(job.analyzers_to_execute.values_list("name", flat=True)),
+                    "tags": list(job.tags.values_list("label", flat=True)),
+                    "errors": job.errors,
+                    "warnings": job.warnings,
+                    "analyzer_reports": analyzer_reports,
```
mlodic: this sounds like a possible Serializer. We are extracting some data from an existing model and jsonify it to make them available externally.

Can you please explore the topic and try to see how it can fit here? https://www.django-rest-framework.org/api-guide/serializers/

Serializers are widely used in IntelOwl and allow to avoid declaring big dictionaries (which are also vulnerable to typos and less readable) and support by default JSON conversion

api_app/chatbot_manager/views.py
```diff
@@ -1,2 +1,75 @@
...
+        )
+
+        return Response(
+            MessageResponseSerializer(
+                {
+                    "session_id": session.pk,
+                    "response": response_text,
+                    "message_id": ai_msg.pk,
+                }
+            ).data,
```
mlodic: why this is populated via a dictionary instead of using normal parameters? again, it is just a nitpick to have things more reliable and readable when possible

mlodic (commented):
thanks for the fast feedback. Here I am requesting some "style" changes, mostly to make this code more uniform to the other IntelOwl code style where Serializers are used almost everywhere.

api_app/chatbot_manager/agent/tools/get_job_details.py
berardifra: Done. Moved the tool output to DRF serializers: `JobToolSerializer` (search results) and `JobDetailToolSerializer `with a nested `AnalyzerReportToolSerializer` -> using the same `source`=/`SlugRelatedField` idioms as `JobListSerializer`. I kept them purpose-built and light instead of reusing `JobListSerializer`, to avoid pushing the nested user object and unused fields into the LLM prompt. It also surfaced a real bug: the tools queried analyzer_reports but the relation is `analyzerreports`, so `get_job_details` was raising `FieldError`  => now fixed via the serializer's `source`.

api_app/chatbot_manager/views.py
berardifra: Yep right, I removed the intermediate dict. `MessageResponseSerializer` now serializes the persisted assistant `ChatMessage` directly (`response`/`message_id` via `source=`), so the response is built from the real object.

api_app/chatbot_manager/agent/tools/search_jobs.py
```diff
@@ -39,24 +44,6 @@ def search_jobs(query: str = "", status: str = "", limit: int = 10) -> str:
...
-                    "received_request_time": str(job.received_request_time),
-                    "finished_analysis_time": str(job.finished_analysis_time)
-                    if job.finished_analysis_time
-                    else None,
-                    "analyzers_to_execute": list(job.analyzers_to_execute.values_list("name", flat=True)),
-                }
-            )
-
-        return json.dumps({"errors": [], "jobs": results}, indent=2)
+        return json.dumps({"errors": [], "jobs": JobToolSerializer(qs, many=True).data}, indent=2)
```
mlodic: good job, the last layer would be to create a Serializer for the output of each job that has "errors" and "jobs" as attributes and use one of the serializer method to dump it to json here. Could you check whether is this feasible?

berardifra: Done, and yes it's feasible. Added a `ToolResultSerializer` base (an always-present `errors` list plus a `to_json()` that renders `.data`), with `SearchJobsResultSerializer`, `JobDetailResultSerializer` and `SummarizeJobResultSerializer` for the `jobs`/`job`/`summary` payloads. Each tool now returns `…ResultSerializer({...}).to_json()`, so the nested `JobToolSerializer` does the model to JSON work and there's no manual `json.dumps` left. Can switch to DRF's `JSONRenderer` instead of the `to_json()` helper if you prefer.

mlodic (approved):

Issue #3716, opened 2026-05-27
Add chat message retention policy with configurable TTL and cleanup task


berardifra:
## Summary
Per maintainer feedback on the W1 architectural questions (Q7): persistent chat storage requires a corresponding cleanup mechanism. This follow-up adds a configurable retention period and a daily Celery beat task to enforce it.

## Tasks
- [ ] Add `CHATBOT_MESSAGE_RETENTION_DAYS` env var (default 90) to `docker/env_file_app_template`
- [ ] Expose the value as a Django setting in `intel_owl/settings/chatbot.py` via `secrets.get_secret`
- [ ] Celery beat task that runs daily and deletes `ChatSession` rows older than the retention period (cascade handles `ChatMessage`)
- [ ] Unit test for the cleanup task with mocked clock
- [ ] Docs note in deployment guide explaining how to tune the retention

## Dependencies
- Depends on PR #3715 (introduces the `ChatSession` / `ChatMessage` models)

## Acceptance criteria
- A fresh install retains chats for 90 days by default
- Operators can override via `CHATBOT_MESSAGE_RETENTION_DAYS` env without code changes
- Cleanup task runs daily and is observable in Flower / Celery logs

PR #3727, opened 2026-05-29
[GSoC 2026] Add chat message retention policy with configurable TTL (refs #3716)

berardifra:
# Description

Adds a persistent retention policy for chatbot conversations. `ChatSession`/`ChatMessage` rows previously grew without bound. This PR introduces a periodic Celery beat task `delete_old_chat_sessions` that deletes every `ChatSession` whose last activity is older than `CHATBOT_MESSAGE_RETENTION_DAYS` (env-configurable, default 90 days); deleting the session removes its messages via the existing `on_delete=CASCADE`.

"Last activity" is computed as the most recent `ChatMessage.timestamp`, falling back to `created_at` for sessions with no messages. `ChatSession.updated_at` (`auto_now`) is not usable here because it is not touched when a message is created, so it does not reflect real activity.

No DB migration: no model change, deletion relies on the existing `on_delete=CASCADE`.

Refs #3716.

Note: soft dependency on #3726 (already merged into the umbrella) —> both touch `api_app/chatbot_manager/tasks.py` but in different locations; this PR only adds a new function.

## Type of change

# Checklist

- [x] Please avoid adding new libraries as requirements whenever it is possible _(no new libraries added)_
- [x] I have added tests for the feature I solved (see `tests` folder). All the tests (new and old ones) gave 0 errors.
- [x] After submitting the PR, if `DeepSource`, `Django Doctors` or other third-party linters trigger alerts during CI, I will solve them.
- [x] I have addressed raised Copilot issues.
- [x] I have reviewed and verified any LLM-generated code included in this PR. I have explicitly stated that I used an LLM in this PR (see Description).

### DIFF
--- a/api_app/chatbot_manager/tasks.py
+++ b/api_app/chatbot_manager/tasks.py
@@ -1,9 +1,18 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
+import datetime
+import logging
+
 from celery import shared_task
+from django.conf import settings
+from django.db.models import Max
+from django.db.models.functions import Coalesce
+from django.utils.timezone import now
 
 from intel_owl.tasks import FailureLoggedTask
+
+logger = logging.getLogger(__name__)
 
 
 # soft_time_limit is 300s on purpose: a single chat turn can chain several ReAct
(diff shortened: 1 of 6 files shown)


berardifra:
While rebuilding the test image to run the new tests locally, I hit a pre-existing version conflict: langchain-ollama==0.3.3 requires langchain-core>=0.3.60, but it was pinned to 0.3.59, so the image failed to build (pip ResolutionImpossible). Bumped langchain-core to 0.3.60 (still satisfies langchain==0.3.25). All chatbot tests pass locally against the rebuilt image.

mlodic:
do you think it is necessary to keep pinned that subdependency? if langchain-ollama already includes langchain-core, we don't pin the subdependency. This can be detable for many reasons, but that's how we decided to handle the project so we should keep this in this way for simplicity unless there's a valid reason to pin the subdependency.

(in pratice we plan to migrate to UV to properly handle subdependencies)

mlodic (commented):
other than the comment on the dependency the PR is great, waiting to approve just because of that

berardifra:
> other than the comment on the dependency the PR is great, waiting to approve just because of that

done, thx :)

mlodic (approved):

Issue #3728, opened 2026-05-29
Add investigation and data-model LLM tools to the chatbot agent


berardifra:
Extend the chatbot's LangChain ReAct agent with four tools that wrap existing IntelOwl APIs, mirroring the W2 job-tool architecture (per-request factory closing over the user, lightweight DRF serializers wrapped in the `ToolResultSerializer` envelope, registered in `build_tools`).

Investigation tools (scoped with `Investigation.objects.visible_for_user(user)` — owner + organization-shared):
- `list_investigations(query, status, limit)` — list visible investigations, optional name/status filters.
- `get_investigation_tree(investigation_id)` — compact nested job tree (treebeard), assembled without per-node N+1, depth/node capped.
- `summarize_investigation(investigation_id)` — human-readable summary (status, total jobs, per-status job breakdown, TLP, tags, start/end).

Data-model tool (scoped on `user=user`, like the job tools):
- `get_data_model(job_id)` — returns a job's aggregated data model via `Job.data_model.serialize()`.

Includes unit tests covering the logical branches and multi-tenancy isolation (cross-user and org-shared). Delivered as two PRs against the `gsoc-2026/llm-chatbot` umbrella: investigation tools first, then `get_data_model`.

PR #3729, opened 2026-05-29
[GSoC 2026] Add investigation LLM tools to the chatbot agent (refs #3728)

berardifra:
# Description

Refs #3728.

Adds three tools to the chatbot's LangChain ReAct agent that wrap IntelOwl's Investigation
API, so an analyst can query investigations in natural language:

- `list_investigations(query, status, limit)` -> lists investigations visible to the user
  (optional name and status filters). An invalid `status` is reported in `errors` instead
  of silently returning no results.
- `get_investigation_tree(investigation_id)` -> returns the investigation's jobs as a
  compact nested tree.
- `summarize_investigation(investigation_id)` -> returns a human-readable summary (status,
  total jobs, per-status job breakdown, TLP, tags, start/end).

Each tool is built per-request and closes over the requesting `user`, so multi-tenancy is
enforced in the closure and cannot be widened by anything the LLM emits. Investigations are
scoped with `Investigation.objects.visible_for_user(user)` (owner + organization-shared,
`for_organization=True`) -> the same scope as `InvestigationViewSet` -> which is intentionally
wider than the existing job tools' `user=user` scope, because investigations carry an
explicit org-sharing flag.

The tree is assembled in Python to avoid a treebeard per-node N+1: one `get_descendants()`
query per root job, with the nesting rebuilt from the materialized `path`; depth and node
count are capped (a `truncated` flag is set when the cap is hit). Descendant jobs in the
investigation tree are gated at the investigation level via `visible_for_user` (UI parity
with `InvestigationViewSet`), not re-scoped per-job. Tree nodes expose only
id/observable/status — no PII.

Outputs reuse the existing `ToolResultSerializer` envelope (`{"errors": [...], <payload>}`)
with purpose-built lightweight DRF serializers (no owner/PII). The list serializer
deliberately exposes only DB columns: `Investigation.tlp/tags/total_jobs` are properties
that each hit the `jobs` relation, so listing them per row would be an N+1; those aggregate
fields are surfaced by `summarize_investigation`/`get_investigation_tree` on a single
investigation instead.

No new dependencies, no new env vars, no migrations.

## Type of change

# Checklist

- [x] Please avoid adding new libraries as requirements whenever it is possible. _(no new libraries added)_
- [x] I have addressed raised Copilot issues. In case of FPs, I have commented the Copilot issue and proved that it is wrong before having the comment resolved.
- [x] I have reviewed and verified any LLM-generated code included in this PR. Also, I have explicitly stated that I have used LLMs in this PR.

### DIFF
--- a/api_app/chatbot_manager/agent/tools/get_investigation_tree.py
+++ b/api_app/chatbot_manager/agent/tools/get_investigation_tree.py
@@ -1,0 +1,109 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+from langchain_core.tools import tool
+
+# Bounds for the LLM-facing tree: keep the serialized payload small enough not to blow up
+# the prompt regardless of how large an investigation is.
+_MAX_DEPTH = 10
+_MAX_NODES = 200
+
+
+def _compact_node(job) -> dict:
+    """Minimal LLM-facing view of a job node: id, observable, status, children."""
+    return {
+        "id": job.pk,
+        "observable": getattr(job.analyzable, "name", None),
+        "status": job.status,
+        "children": [],
+    }
+
+
+def _build_tree(investigation) -> dict:
+    """Assemble a compact nested tree for an investigation, avoiding the treebeard N+1.
+
+    A Job is a treebeard ``MP_Node``; walking it with ``get_children()`` recursively would
+    fire one query per node. Instead we fetch each root's whole subtree with a single
+    ``get_descendants()`` call and rebuild the nesting in Python from treebeard's
+    materialized ``path`` (a child's parent path is its own path minus the last step), so
+    there are no per-node queries. Depth and node count are capped to bound the payload.
+    """
+    payload = {
+        "id": investigation.pk,
+        "name": investigation.name,
+        "status": investigation.status,
+        "jobs": [],
+    }
+    remaining = _MAX_NODES
+    truncated = False
+
+    # `investigation.jobs` are the root jobs; their descendants live only in the tree.
+    for root in investigation.jobs.select_related("analyzable"):
+        if remaining <= 0:
+            truncated = True
+            break
+        root_node = _compact_node(root)
+        payload["jobs"].append(root_node)
+        remaining -= 1
+        # Map path -> node for every kept node, to link children to parents in O(1).
+        by_path = {root.path: root_node}
+
+        # One query for the whole subtree, in path order (treebeard pre-order DFS), so a
+        # parent is always processed before its children.
+        for job in root.get_descendants().select_related("analyzable").order_by("path"):
+            if (job.depth - root.depth) > _MAX_DEPTH:
+                continue
+            parent = by_path.get(job.path[: -job.steplen])
+            if parent is None:
+                # Ancestor was capped out (depth/budget); skip to avoid orphaning.
+                continue
+            if remaining <= 0:
+                truncated = True
+                break
+            node = _compact_node(job)
+            by_path[job.path] = node
+            parent["children"].append(node)
+            remaining -= 1
+
+    if truncated:
+        payload["truncated"] = True
+    return payload
+
+
+def make_get_investigation_tree_tool(user):
+    # Built per-request and closed over `user`. The lookup is scoped with
+    # `visible_for_user` (owned + organization-shared investigations), so the LLM cannot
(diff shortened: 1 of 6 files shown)


api_app/chatbot_manager/agent/tools/get_investigation_tree.py
```diff
@@ -0,0 +1,109 @@
...
+    ``get_descendants()`` call and rebuild the nesting in Python from treebeard's
+    materialized ``path`` (a child's parent path is its own path minus the last step), so
+    there are no per-node queries. Depth and node count are capped to bound the payload.
+    """
+    payload = {
+        "id": investigation.pk,
+        "name": investigation.name,
+        "status": investigation.status,
+        "jobs": [],
+    }
```
mlodic: I would use a dataclass instead here. (also add "truncated" here too). Can you try using that data type?

As suggested last time with serializers, also here I suggest avoiding the usage of dictionaries when possible. While it is easy and fast to build something working with them, the structure of the data inside of it could go out of control pretty fast during future refactors/improvments.

api_app/chatbot_manager/agent/tools/get_investigation_tree.py
```diff
@@ -0,0 +1,109 @@
...
+        (a `"truncated": true` flag is added when the cap is hit).
+
+        Args:
+            investigation_id: The numeric ID of the investigation.
+
+        Returns:
+            JSON string with shape {"errors": [...], "investigation": {...} | null}.
+        """
+        from api_app.chatbot_manager.serializers import InvestigationTreeResultSerializer
+        from api_app.investigations_manager.models import Investigation
```
mlodic: you need imports here and not at the start of the file because otherwise you have cyclic import errors?

api_app/chatbot_manager/agent/tools/list_investigations.py
```diff
@@ -0,0 +1,51 @@
...
+            status: Filter by investigation status. Valid values: created, running,
+                    concluded. An unknown value is ignored and reported in `errors`.
+            limit: Maximum number of results to return (default 10, max 50).
+
+        Returns:
+            JSON string with shape {"errors": [...], "investigations": [...]}.
+        """
+        from api_app.chatbot_manager.serializers import InvestigationsResultSerializer
+        from api_app.investigations_manager.choices import InvestigationStatusChoices
+        from api_app.investigations_manager.models import Investigation
```
mlodic: same question here

api_app/chatbot_manager/agent/tools/list_investigations.py
```diff
@@ -0,0 +1,51 @@
...
+            # The status string comes from the LLM; validate it against the enum so an
+            # invalid value surfaces a message instead of silently returning 0 results.
+            normalized = status.strip().lower()
+            if normalized in set(InvestigationStatusChoices.values):
+                qs = qs.filter(status=normalized)
+            else:
+                valid = ", ".join(InvestigationStatusChoices.values)
+                errors.append(f"Unknown status '{status}'; valid values are: {valid}.")
+
+        limit = min(int(limit), 50)
```
mlodic: I would make explicit what "50" is here, just for clarity

mlodic (commented):
just minor style stuff

api_app/chatbot_manager/agent/tools/get_investigation_tree.py
berardifra: Done. The tree is now built with dataclasses (`_JobNode`/`_InvestigationTree`) instead of dicts, with `truncated` as an explicit field; it's converted to a dict only at the envelope boundary via `asdict`.

api_app/chatbot_manager/agent/tools/get_investigation_tree.py
berardifra: There's no circular import here, so I moved them to module level. The already-merged job tools still use function-local imports; I can align them in a separate cleanup if you'd like.

api_app/chatbot_manager/agent/tools/list_investigations.py
berardifra: Done. Extracted as a module constant `_MAX_RESULTS = 50` with a comment.

api_app/chatbot_manager/agent/tools/get_investigation_tree.py
mlodic: yeah thanks

mlodic (approved):

PR #3731, opened 2026-05-30
[GSoC 2026] Add get_data_model LLM tool to the chatbot agent (refs #3728)

berardifra:
# Description

Refs #3728. Follow-up to #3729 (investigation tools); soft-depends on it for the shared
touchpoints (`build_tools` in `agent/tools/__init__.py` and `serializers.py`), which are now
merged into the umbrella, so this PR applies cleanly on top.

Adds a fourth tool to the chatbot's LangChain ReAct agent:

- `get_data_model(job_id)` — returns a job's aggregated data model (IntelOwl's normalized,
  analyzer-agnostic view of the observable: evaluation, reliability, tags, kill-chain phase,
  plus type-specific fields for domains/IPs/files).

The tool is built per-request and closes over the requesting `user`, and the lookup is
hard-scoped with `Job.objects.get(pk=job_id, user=user)` — the same owner-strict scope as the
existing job tools (`search_jobs`, `get_job_details`, `summarize_job`). It mirrors the public
`JobSerializer.get_data_model`: `data_model` is a nullable `GenericForeignKey`, so the tool
returns `job.data_model.serialize()` when present and an empty object otherwise.

The result rides the existing `ToolResultSerializer` envelope as `DataModelResultSerializer`,
where `data_model` is a `DictField`. This is not a hand-built dict: `BaseDataModel.serialize()`
is the model's own DRF serializer output (polymorphic across `DomainDataModel`/`IPDataModel`/
`FileDataModel`), so the envelope simply carries that already-serialized dict.

No new dependencies, no new env vars, no migrations.

This PR was developed with the assistance of an LLM; all generated code has been reviewed
and verified.

## Type of change

# Checklist

- [x] Please avoid adding new libraries as requirements whenever it is possible. _(no new libraries added)_
- [x] I have addressed raised Copilot issues. In case of FPs, I have commented the Copilot issue and proved that it is wrong before having the comment resolved.
- [x] I have reviewed and verified any LLM-generated code included in this PR. Also, I have explicitly stated that I have used LLMs in this PR.

### DIFF
--- a/api_app/chatbot_manager/agent/tools/__init__.py
+++ b/api_app/chatbot_manager/agent/tools/__init__.py
@@ -1,6 +1,7 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
+from .get_data_model import make_get_data_model_tool
 from .get_investigation_tree import make_get_investigation_tree_tool
 from .get_job_details import make_get_job_details_tool
 from .list_investigations import make_list_investigations_tool
@@ -26,4 +27,5 @@ def build_tools(user) -> list:
         make_list_investigations_tool(user),
         make_get_investigation_tree_tool(user),
         make_summarize_investigation_tool(user),
+        make_get_data_model_tool(user),
     ]
--- a/api_app/chatbot_manager/agent/tools/get_data_model.py
+++ b/api_app/chatbot_manager/agent/tools/get_data_model.py
@@ -1,0 +1,42 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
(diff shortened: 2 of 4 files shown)


berardifra:
On `get_data_model`: unlike the other tools, the data model is returned in full, with no size cap, matching the public `JobSerializer.get_data_model`. The caps on the other tools (analyzer reports truncated to 500 chars, the investigation tree bounded by depth/node count) are safe because that output is auxiliary context, whereas here the data model is the payload the user asked for, so truncating it would defeat the tool. If prompt size turns out to be a problem in practice for large fields (`certificates`, `file_information`, `ietf_report`), we can add targeted truncation in a follow-up.

berardifra:
Yo @mlodic , as soon as you can, could you check? Thanks :)

mlodic (approved):

PR #3733, opened 2026-06-01
fix(analyzers): handle empty ASN history in BGPRanking

Valyrian-Code:
## Description

When BGP Ranking has no ASN history for an IP, the API returns an empty `response` object. The current implementation calls `.popitem()` on that empty dictionary, resulting in an uncaught:

```text
KeyError: 'popitem(): dictionary is empty'
```

This occurs before the intended `AnalyzerRunException("ASN not found ...")` can be raised.

This change handles the empty-response case explicitly, allowing the analyzer to raise a clean `AnalyzerRunException` instead of terminating with a traceback. Behavior remains unchanged when ASN history is available.

## Type of Change

* [x] Bug fix (non-breaking change that fixes an issue)

## Tests

Added a regression test, `test_run_raises_clean_exception_on_empty_asn_history`, which mocks an empty API response:

```python
{"response": {}}
```

The test verifies that `AnalyzerRunException` is raised.

* Fails on the current code with:

  ```text
  KeyError: 'popitem(): dictionary is empty'
  ```
* Passes with this fix applied.

All existing tests continue to pass. `ruff check` and `ruff format --check` are also clean.

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/bgp_ranking.py
+++ b/api_app/analyzers_manager/observable_analyzers/bgp_ranking.py
@@ -33,7 +33,11 @@ class BGPRanking(classes.ObservableAnalyzer):
         )
         response.raise_for_status()
         response = response.json()
-        asn = response.get("response", {}).popitem()[1].get("asn", None)
+        asn_history = response.get("response", {})
+        # an IP with no ASN history returns an empty "response"; peek the latest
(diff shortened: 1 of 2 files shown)


Valyrian-Code:
Hi @mlodic, 
> could you please review this PR and share any suggestions? Thanks!

GIF

mlodic:
I'll merge this because it is a minor change but that's not following the correct process and guidelines

PR #3734, opened 2026-06-01
fix(analyzers): handle empty results in Netlas

Valyrian-Code:
## Description

When Netlas has no WHOIS record for an IP, the API returns an empty `items` list. The current implementation accesses:

```python
response.json()["items"][0]["data"]
```

which results in an uncaught:

```text
IndexError: list index out of range
```

when no results are returned.

This change handles empty or missing `items` responses explicitly, allowing the analyzer to raise a clean `AnalyzerRunException` instead of failing with a traceback. Behavior remains unchanged when WHOIS data is available.

## Type of Change

* [x] Bug fix (non-breaking change that fixes an issue)

## Tests

Added a regression test, `test_run_raises_clean_exception_on_empty_items`, which mocks the API response:

```python
{"items": []}
```

The test verifies that `AnalyzerRunException` is raised.

* Fails on the current code with:

  ```text
  IndexError: list index out of range
  ```
* Passes with this fix applied.

All existing tests continue to pass. `ruff check` and `ruff format --check` are also clean.

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/netlas.py
+++ b/api_app/analyzers_manager/observable_analyzers/netlas.py
@@ -32,5 +32,9 @@ class Netlas(classes.ObservableAnalyzer):
         except requests.RequestException as e:
             raise AnalyzerRunException(e)
 
-        result = response.json()["items"][0]["data"]
-        return result
+        # an IP with no whois record returns an empty "items" list, which must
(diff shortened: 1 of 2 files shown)


Valyrian-Code:
Hi @mlodic 

> could you please review this PR and share any suggestions? Thanks!

mlodic (approved):

Issue #3732, opened 2026-05-31
Add analyzer/playbook and analysis-triggering LLM tools to the chatbot agent


berardifra:
## Summary

Add the analyzer/playbook and analysis-triggering tools to the chatbot LangChain ReAct agent (W5), so an
analyst can ask which analyzers/playbooks apply to an observable and — with explicit confirmation — start an
analysis from the chat.

## Tools

- **`list_analyzers`** (read-only) — analyzers enabled for the user (org-aware), optionally filtered by
  observable type.
- **`recommend_playbook`** (read-only) — directly-launchable playbooks applicable to an observable's
  classification, scoped with `visible_for_user`.
- **`analyze_observable`** (action) — starts an analysis. **Safety guardrail (proposal requirement):** validate
  observable classification + TLP + runnable plugins and return an explicit confirmation request **before**
  triggering. Reuses the existing `ObservableAnalysisSerializer` pipeline; the trigger fires only on an
  explicit `confirm=True` (no implicit/accidental launch). LLM arguments treated as untrusted; multi-tenancy
  and TLP honored.

## Notes

Delivered as two sequential PRs to the `gsoc-2026/llm-chatbot` umbrella (read-only tools first, then the
action tool), mirroring the W4 split. Tests cover logical branches + cross-user/org isolation, and (for the
action tool) assert that nothing is triggered without confirmation.

berardifra:
Reopening: only the read-only tools (list_analyzers, recommend_playbook, #3740) are merged; analyze_observable is still pending.

berardifra:
All three W5 tools merged: #3740 (list_analyzers, recommend_playbook) and #3744 (analyze_observable). Closing.

PR #3740, opened 2026-06-04
[GSoC 2026] Add analyzer/playbook LLM tools to the chatbot agent (refs #3732)

berardifra:
Refs #3732. _(Targets the `gsoc-2026/llm-chatbot` umbrella, not `develop`. Using `Refs` (not `Closes`) on purpose — the umbrella isn't `develop`, so the issue is closed manually after merge.)_

# Description

Adds two **read-only** tools to the chatbot LangChain ReAct agent (W5 — analyzer/playbook domain):

- **`list_analyzers(observable_type="", limit=50)`** — lists the analyzers enabled for the requesting user
  (globally not `disabled`, and not disabled for the user's organization), optionally filtered to those
  supporting a given observable type. `observable_type` is validated against the `Classification` enum; an
  unknown value is reported in `errors` instead of silently returning an empty list.
- **`recommend_playbook(observable_name="", classification="", limit=50)`** — suggests directly-launchable
  (`starting=True`, non-disabled) playbooks applicable to an observable's classification, scoped to the
  playbooks visible to the user (`visible_for_user` — owned + organization-shared). The classification is
  derived from the observable via `Classification.calculate_observable` when not supplied; an explicitly
  provided one is validated against the enum.

Both follow the existing tool pattern (a `make_*_tool(user)` factory that closes over the user, so
multi-tenancy is enforced at build time and the LLM can't widen the scope) and return the standard
`{"errors": [...], "<payload>": ...}` envelope via `ToolResultSerializer.to_json()`. Output uses
purpose-built light serializers (`AnalyzerConfigToolSerializer`, `PlaybookConfigToolSerializer`) rather than
the heavy config serializers, to keep the LLM prompt small. Both cap their result list at 50
(`_MAX_RESULTS`) and clamp the LLM-supplied `limit` into `[1, 50]`, so a single call can't flood the prompt.

These tools are read-only — no analysis is triggered. The analysis-launching tool `analyze_observable`
(with its safety guardrails) follows in a separate PR.

## Type of change

# Checklist

- [ ] The pull request is for the branch `develop` _(targets the `gsoc-2026/llm-chatbot` umbrella)_
- [x] Please avoid adding new libraries as requirements whenever it is possible _(no new dependencies added)_
- [x] If external libraries/packages with restrictive licenses were added, they were added in the [Legal Notice](https://github.com/certego/IntelOwl/blob/master/.github/legal_notice.md) section _(N/A — no new libraries added)_
- [x] I have added tests for the feature I solved (see `tests` folder); all tests (new and old) gave 0 errors _(run locally against a rebuilt test image — the "Build & Tests" CI workflow does not run on umbrella-targeting PRs)_
- [x] After submitting the PR, if `DeepSource`, `Django Doctors` or other third-party linters triggered alerts during CI, I have solved those alerts
- [x] I have reviewed and verified any LLM-generated code included in this PR. **I used an LLM coding assistant while developing this PR, and reviewed all of its output.**

### DIFF
--- a/api_app/chatbot_manager/agent/tools/list_analyzers.py
+++ b/api_app/chatbot_manager/agent/tools/list_analyzers.py
@@ -12,10 +12,6 @@ from api_app.choices import Classification
 # single call can't pull an unbounded list into the prompt.
 _MAX_RESULTS = 50
 
-# Observable analyzers never apply to the `file` classification (that's the file-analysis path),
-# so the accepted/advertised observable types are all classifications except FILE.
-_VALID_OBSERVABLE_TYPES = [c for c in Classification.values if c != Classification.FILE.value]
-
 
 def make_list_analyzers_tool(user):
     # Built per-request and closed over `user`. Unlike the job/investigation tools, analyzer
@@ -55,20 +51,30 @@ def make_list_analyzers_tool(user):
             user
         )
 
+        # The accepted observable types are an IntelOwl-core notion (every classification except
+        # FILE, which is the file-analysis path), so reuse the core helper instead of a local copy.
+        valid_types = Classification.observable_classifications()
         if observable_type:
             # The type string comes from the LLM; validate it against the enum so an invalid
             # value surfaces a message instead of silently returning 0 results.
             normalized = observable_type.strip().lower()
-            if normalized in set(_VALID_OBSERVABLE_TYPES):
+            if normalized in set(valid_types):
                 # `observable_supported` is a ChoiceArrayField: `__contains` matches rows whose
                 # array includes the requested type.
                 qs = qs.filter(observable_supported__contains=[normalized])
             else:
-                valid = ", ".join(_VALID_OBSERVABLE_TYPES)
+                valid = ", ".join(valid_types)
                 errors.append(f"Unknown observable_type '{observable_type}'; valid values are: {valid}.")
 
-        # Clamp the LLM-supplied limit into [1, _MAX_RESULTS] (treat tool args as untrusted).
-        limit = max(1, min(int(limit), _MAX_RESULTS))
+        # Clamp the LLM-supplied limit into [1, _MAX_RESULTS] (treat tool args as untrusted) and
+        # tell the caller when the requested value was capped, so a truncated list isn't silent.
+        requested_limit = int(limit)
+        limit = max(1, min(requested_limit, _MAX_RESULTS))
+        if requested_limit > _MAX_RESULTS:
+            errors.append(
+                f"Requested limit {requested_limit} exceeds the maximum {_MAX_RESULTS}; "
+                f"returning at most {_MAX_RESULTS} results."
+            )
         qs = qs.order_by("name")[:limit]
 
         return ListAnalyzersResultSerializer({"errors": errors, "analyzers": qs}).to_json()
--- a/api_app/chatbot_manager/agent/tools/list_analyzers.py
+++ b/api_app/chatbot_manager/agent/tools/list_analyzers.py
@@ -1,0 +1,76 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+from langchain_core.tools import tool
+
+from api_app.analyzers_manager.constants import TypeChoices
+from api_app.analyzers_manager.models import AnalyzerConfig
+from api_app.chatbot_manager.serializers import ListAnalyzersResultSerializer
+from api_app.choices import Classification
+
+# Hard cap on the number of results returned to the LLM (mirrors the other tools), so a
+# single call can't pull an unbounded list into the prompt.
+_MAX_RESULTS = 50
+
+# Observable analyzers never apply to the `file` classification (that's the file-analysis path),
+# so the accepted/advertised observable types are all classifications except FILE.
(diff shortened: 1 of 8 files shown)


mlodic:
I think the issues in the conflicts are due to the 2 parallels PRs that change the same code.

I merged the previous one without any change so it should be easy to fix the conflict.

mlodic:
This can happen from time to time considering that I can't review the changes as soon as they are published and some parallel PRs are normal in such environments.

Try to consider this while doing the changes to avoid having to fix conflicts every time. Isolate the content as soon as possible, abuse creation of additional files (like serializers.py and test_tools.py could become folders and contain isolated tests in separated files). This should reduce the issues

api_app/chatbot_manager/agent/tools/list_analyzers.py
```diff
@@ -0,0 +1,76 @@
...
+from api_app.chatbot_manager.serializers import ListAnalyzersResultSerializer
+from api_app.choices import Classification
+
+# Hard cap on the number of results returned to the LLM (mirrors the other tools), so a
+# single call can't pull an unbounded list into the prompt.
+_MAX_RESULTS = 50
+
+# Observable analyzers never apply to the `file` classification (that's the file-analysis path),
+# so the accepted/advertised observable types are all classifications except FILE.
+_VALID_OBSERVABLE_TYPES = [c for c in Classification.values if c != Classification.FILE.value]
```
mlodic: this information is an IntelOwl core info and should stay logically in the core part.
Please modify the Classification class and add an helper that generates this list and then import it here so it can be reused.

For instance, I noticed that in other parts of the code (see aggregate_observable_classification) we use the same list so it should be nice if you could modify that too while we are here.

api_app/chatbot_manager/agent/tools/list_analyzers.py
```diff
@@ -0,0 +1,76 @@
...
+    # the user's own rows; it lists the enabled observable analyzers and exposes per-user
+    # readiness through the `runnable` annotation instead.
+    #
+    # Why no `.filter(runnable=True)`: `annotate_runnable` (queryset.py) folds health-check +
+    # "all required parameters configured" on top of enabled/org-enabled, so a hard filter would
+    # silently drop every key-based analyzer (VirusTotal, ...) on a deploy without API keys ->
+    # near-empty lists and flaky tests. We list the enabled analyzers and surface readiness as a
+    # per-row flag instead, so the LLM can still say "this analyzer applies but isn't configured
+    # for you". `runnable` is False when the analyzer is disabled for the user's organization OR
+    # not fully configured/healthy.
```
mlodic, line 25: thanks for the thorough explanation, that's really useful

api_app/chatbot_manager/agent/tools/list_analyzers.py
```diff
@@ -0,0 +1,76 @@
...
+            if normalized in set(_VALID_OBSERVABLE_TYPES):
+                # `observable_supported` is a ChoiceArrayField: `__contains` matches rows whose
+                # array includes the requested type.
+                qs = qs.filter(observable_supported__contains=[normalized])
+            else:
+                valid = ", ".join(_VALID_OBSERVABLE_TYPES)
+                errors.append(f"Unknown observable_type '{observable_type}'; valid values are: {valid}.")
+
+        # Clamp the LLM-supplied limit into [1, _MAX_RESULTS] (treat tool args as untrusted).
+        limit = max(1, min(int(limit), _MAX_RESULTS))
```
mlodic: should we explicitly raise that a clamp has been done here? I mean, with a comment, a "warning" or whatever

berardifra:
> This can happen from time to time considering that I can't review the changes as soon as they are published and some parallel PRs are normal in such environments.
> 
> Try to consider this while doing the changes to avoid having to fix conflicts every time. Isolate the content as soon as possible, abuse creation of additional files (like serializers.py and test_tools.py could become folders and contain isolated tests in separated files). This should reduce the issues

Yeah, I was expecting this one: two PRs touching the same files was a known risk on my side, so the rebase onto the umbrella was already part of the plan. Done now: additive conflicts only, kept both sides, 31/31 tests + Ruff green, mergeable again.
Good point on the isolation though -> I'll split serializers.py and test_tools.py into packages (one module per tool) so parallel PRs stop colliding.

berardifra:
Addressed in 761950b5, thanks

api_app/chatbot_manager/agent/tools/recommend_playbook.py
```diff
@@ -0,0 +1,80 @@
...
+        # Clamp the LLM-supplied limit into [1, _MAX_RESULTS] (treat tool args as untrusted) and
+        # cap the queryset so a broadly-supported classification can't flood the prompt; tell the
+        # caller when the requested value was capped, so a truncated list isn't silent.
+        requested_limit = int(limit)
+        limit = max(1, min(requested_limit, _MAX_RESULTS))
+        if requested_limit > _MAX_RESULTS:
+            errors.append(
+                f"Requested limit {requested_limit} exceeds the maximum {_MAX_RESULTS}; "
+                f"returning at most {_MAX_RESULTS} results."
+            )
```
mlodic: nitpick: can you please estrapolate a function from this code to avoid duplication in the sibling method?

berardifra: yep, i understand.

mlodic (approved):

PR #3744, opened 2026-06-08
[GSoC 2026] Add analyze_observable action tool to the chatbot agent (refs #3732)

berardifra:
# Description

Adds `analyze_observable`, the chatbot agent's only **action** tool — it can start a real IntelOwl analysis. Refs #3732 (last of the 3 W5 tools).

It reuses `ObservableAnalysisSerializer` exactly as the REST `analyze_observable` view does (validation, defang, classification, private-IP guardrails, `maximum_tlp` plugin filtering, 'nothing runnable' refusal) and is **two-phase / confirm-gated**:
- `confirm=false` (default) → validates and returns a `plan` (analyzers/connectors that would run, skipped ones, classification) andt riggers **nothing**.
- `confirm=true` → `save(send_task=True)`, the only path that reaches `job_pipeline.apply_async`.

**Tenant isolation:** the reused serializer resolves `playbook_requested` via `PlaybookConfig.objects.all()` with no visibility filter, so the tool adds an explicit `visible_for_user(user)` guard on the `playbook` arg (a private playbook of another org is rejected, never resolved). The created `Job.user` is the requesting user.

**Safety boundary (honest):** the code-level guarantee is that no call without `confirm=true` can reach `apply_async`. The human-in-the-loop gate is the system-prompt instruction + the conversation turn boundary (agent returns the plan, the turn ends, the user's next message authorizes `confirm=true`) — not the flag alone. TLP never blocks a launch; it only filters which plugins run.

## Type of change

# Checklist
- [ ] The pull request is for the branch `develop` — *targets the `gsoc-2026/llm-chatbot` umbrella branch (GSoC workflow)*
- [x] Please avoid adding new libraries as requirements whenever it is possible — *no new libraries*
- [x] I have added tests for the feature/bug I solved. All the tests (new and old ones) gave 0 errors — *8 new tests; full `tests.api_app.chatbot_manager` suite 58/58 green locally*
- [x] After submitting the PR, if `DeepSource`/`Django Doctors`/other linters trigger alerts, I have solved them
- [x] I have reviewed and verified any LLM-generated code included in this PR. I used an LLM (Claude Code) to assist and reviewed the full diff (tool logic, the isolation guard, and the tests).

### DIFF
--- a/api_app/chatbot_manager/agent/tools/analyze_observable.py
+++ b/api_app/chatbot_manager/agent/tools/analyze_observable.py
@@ -1,33 +1,16 @@
 # This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
 # See the file 'LICENSE' for copying permission.
 
-from dataclasses import asdict, dataclass, field
 from types import SimpleNamespace
-from typing import List, Optional
+from typing import List
 
+from django.utils.timezone import now
 from langchain_core.tools import tool
 
 from api_app.chatbot_manager.serializers.analyze_observable import AnalyzeObservableResultSerializer
-from api_app.choices import TLP, ScanMode
+from api_app.choices import TLP
 from api_app.playbooks_manager.models import PlaybookConfig
 from api_app.serializers.job import ObservableAnalysisSerializer
-
-
-@dataclass
-class _AnalysisPlan:
-    """Preview of what a confirmed analyze_observable call would launch.
-
-    Computed (non-model) data, so it is a typed dataclass rather than a free-form dict (Matteo's
-    rule); `asdict()` converts it only at the serializer boundary.
-    """
-
-    observable_name: str
-    classification: str
-    tlp: str
-    playbook: Optional[str]
-    analyzers: List[str]
-    connectors: List[str]
-    skipped: List[str] = field(default_factory=list)
 
 
 def _flatten_errors(errors) -> List[str]:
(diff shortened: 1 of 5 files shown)


api_app/chatbot_manager/agent/tools/analyze_observable.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Returns:
+            JSON string with shape {"errors": [...], "confirmation_required": bool,
+            "plan": {...} | null, "job": {...} | null}.
+        """
+        # The reused serializer reads the requesting user from `context["request"].user`; a shim is
+        # enough since the create path only ever reads `.user`.
+        shim = SimpleNamespace(user=user)
+        # FORCE_NEW_ANALYSIS (not the platform default CHECK_PREVIOUS_ANALYSIS): confirm=True always
+        # starts a fresh job rather than possibly returning a cached one -> deterministic contract.
```
mlodic: can you please explain further why the choice of forcing the analysis?

api_app/chatbot_manager/agent/tools/analyze_observable.py
```diff
@@ -0,0 +1,161 @@
...
+from langchain_core.tools import tool
+
+from api_app.chatbot_manager.serializers.analyze_observable import AnalyzeObservableResultSerializer
+from api_app.choices import TLP, ScanMode
+from api_app.playbooks_manager.models import PlaybookConfig
+from api_app.serializers.job import ObservableAnalysisSerializer
+
+
+@dataclass
+class _AnalysisPlan:
```
mlodic: actually here this dataclass is pretty useless considering that it is used only once and after having it initialized it is transformed back into a dict and never used anymore.
Can you see whether you can enforce this structure in the AnalyzeObservableResultSerializer.plan to remove the unstructured dict declaration? In that way this structure can be used appropriately. Otherwise the other option is to just remove it.

Broadly speaking, I don't like to enforce structures in any case. It's the best option to avoid errors, promote future maintainability, provide more clarity...but sometimes it is just not the case.

berardifra: Agreed, it was used once and immediately flattened. Removed the dataclass and enforced the structure where it belongs: plan is now a nested AnalysisPlanSerializer with explicit fields, so there's no more unstructured DictField :)

api_app/chatbot_manager/agent/tools/analyze_observable.py
berardifra: It was a deliberate trade-off, not an oversight. I forced a new analysis to keep the agent's contract predictable: a confirmed request always launches a fresh job, so the assistant never has to explain to the user why it "returned" an analysis from a few hours ago, and the trigger stays deterministic (also easier to reason about and test).
My doubt was exactly what you're pointing at: forcing diverges from the platform default(CHECK_PREVIOUS_ANALYSIS) and from the REST analyze_observable endpoint, and it re-runs analyzers even when an identical analysis just ran -> burning quota for a benefit that's really developer-convenience, not a product requirement. Weighing it again, consistency + dedup win.
So I dropped the scan_mode override and use the default. To keep the UX clarity I wanted from forcing, the envelope now returns reused: true/false, so the agent can still say "started a new analysis" vs "one already ran recently, here it is" , without the duplicate cost.

mlodic (approved):
thanks for the overall considerations on the decision making, that's really helpful.

I think there's nothing else to comment here, good job

Issue #3756, opened 2026-06-09
Chatbot: WebSocket ChatConsumer with token-by-token streaming


berardifra:
The self-hosted LLM chatbot: a Django Channels `ChatConsumer` that streams a chat turn token-by-token over a WebSocket, backed by the `process_chat_message` Celery task on a dedicated chatbot worker. Part of the gsoc-2026/llm-chatbot umbrella.

PR #3757, opened 2026-06-09
[GSoC 2026] Add chatbot WebSocket consumer with token streaming (refs #3756)

berardifra:
Refs #3756. _(Targets the `gsoc-2026/llm-chatbot` umbrella, not `develop`. Using `Refs` (not `Closes`) on purpose — the umbrella isn't `develop`, so the issue is closed manually after merge. Branches off the refreshed umbrella.)_

# Description

Adds the **real-time path** of the chatbot (W6): a Django Channels `ChatConsumer` and an implementation of
the `process_chat_message` Celery task (previously a `NotImplementedError` stub) so a chat turn runs
off-request on a dedicated queue and **streams the answer token-by-token** to the browser over a WebSocket.
The synchronous REST turn (`POST /api/chatbot/sessions/message`, W2) stays as the non-streaming fallback,
unchanged.

```
Browser ──ws/chat/?context_url=…──▶ ChatConsumer.connect()  (joins group chat-<user_id>)
Browser ──{message, session_id?}──▶ ChatConsumer.receive_json()  → ack + process_chat_message.delay(...)
[chatbot worker] agent run  → chat.start → chat.status* / chat.token* → chat.end   (group_send)
                            → chat.error on failure (turn dropped, nothing persisted)
chat.*  ──▶ ChatConsumer.chat_*()  ──▶ send_json  ──▶ Browser
```

- **`ChatConsumer(JsonWebsocketConsumer)`** mirrors `JobConsumer` (sync consumer, `async_to_sync` group
  add/discard, handler methods named from the event `type`). It is **bidirectional** — the inbound frame,
  treated as untrusted, is validated by the existing `MessageRequestSerializer`, the supplied `session_id`
  is resolved scoped to the authenticated user (or a new session is created), the client is acked, and the
  Celery turn is enqueued. The group is **per-user** (`chat-<user_id>`) keyed on `scope["user"].id`: the
  fixed URL `ws/chat/?context_url=…` carries no session id, every event carries `session_id` so a client
  with several sessions/tabs demultiplexes, and a client can only ever receive its own stream. Auth is
  enforced upstream by `WSAuthMiddleware` (anonymous rejected with 1008).

- **Token streaming.** `process_chat_message` runs the ReAct `AgentExecutor` synchronously in the worker
  and attaches a callback **at run level** (`invoke(config={"callbacks": [handler]})`) — not on the LLM —
  so it sees both the LLM token stream and the agent's tool actions. `ChatStreamingCallbackHandler` streams
  only what follows the `Final Answer:` marker (the ReAct Thought/Action scaffolding is hidden) plus a
  lightweight `chat.status` per tool, pushing each via `group_send`. The worker is sync (no running loop),
  so `async_to_sync(channel_layer.group_send)` is the correct bridge — the same precedent as
  `JobConsumer.serialize_and_send_job`.

- **Persistence is the source of truth.** The streamed tokens are a live preview; the worker persists
  `result["output"]` and sends it in `chat.end` (with the assistant `message_id`), which the client treats
  as authoritative — so a tokenization miss can never corrupt the stored/displayed answer. History is
  snapshotted before the turn is written, mirroring the REST ordering. On any failure (Ollama unreachable,
  `SoftTimeLimitExceeded`, etc.) a single generic `chat.error` is emitted and the turn is dropped (nothing
  persisted), matching the sync path.

- **Dedicated chatbot worker (required, not optional).** The `chatbot` queue was declared but **no worker
  consumed it** (the four `celery_*.sh` entrypoints run `-Q …,broadcast,config`, never `chatbot`), so the
  enqueued task would hang silently. This PR adds a `celery_worker_chatbot` service + `celery_chatbot.sh`
  entrypoint (`-Q chatbot,broadcast,config`, `-c 2`, `--time-limit=600`) living in `ollama.override.yml`,
  so the same `--ollama` flag that starts the LLM also starts the worker that uses it; no chatbot worker is
  forced on deployments that don't use the LLM. A `task_routes` entry pins `process_chat_message` to the
  queue regardless of call site. _(Open to folding `chatbot` into the default worker's `-Q` instead if you
  prefer not to add a container — flagging the deployment choice.)_

## Framework-specific notes

- **Run-level callbacks, not on the model.** `on_agent_action` (the tool-status events) fires from the
  executor, not the LLM; binding the handler only to `ChatOllama` would stream tokens but lose the status
  events. `ChatOllama(streaming=True)` is set; callbacks are passed per run.
- **`Final Answer:` is matched as a substring** in an accumulating buffer (re-armed per LLM call), not a
  fixed token tuple — a local model's tokenization of `"Final Answer:"` is not guaranteed.

## Tests

- `test_streaming.py` — the marker heuristic: streams only after `Final Answer:`, marker split across tokens
  still detected, pre-marker tokens never streamed, gate re-arms per iteration, tool-status event.
- `test_consumers.py` — group join keyed on the authenticated user, group leave on disconnect, message
  creates/acks/enqueues, owned session reused, oversized frame rejected, **another user's session rejected**,
  handlers relay the payload, anonymous closed by `WSAuthMiddleware`. Driven by direct calls to the
  consumer's sync methods with a mocked channel layer — `WebsocketCommunicator` hangs here because a sync
  consumer's channel-receive listener is bound to the event loop and the per-test `async_to_sync` loops tear
  it down across loops.
- `test_tasks.py` — persists the exchange and emits `chat.start`/`chat.end` with the assistant id; an agent
  failure emits `chat.error` and drops the turn; a session not owned by the user is rejected before the agent
  is built. Ollama/channel layer fully mocked.

No new dependencies, no model/migration changes.

## Type of change

# Checklist

- [ ] The pull request is for the branch `develop` _(targets the `gsoc-2026/llm-chatbot` umbrella)_
- [x] Please avoid adding new libraries as requirements whenever it is possible _(no new dependencies added)_
- [x] If external libraries/packages with restrictive licenses were added, they were added in the [Legal Notice](https://github.com/certego/IntelOwl/blob/master/.github/legal_notice.md) section _(N/A — no new libraries added)_
- [x] I have added tests for the feature I solved (see `tests` folder); all tests (new and old) gave 0 errors _(74 chatbot tests run locally against a rebuilt test image — the "Build & Tests" CI workflow does not run on umbrella-targeting PRs)_
- [x] After submitting the PR, if `DeepSource`, `Django Doctors` or other third-party linters triggered alerts during CI, I have solved those alerts
- [x] I have reviewed and verified any LLM-generated code included in this PR. **I used an LLM coding assistant while developing this PR, and reviewed all of its output.**

### DIFF
--- a/api_app/chatbot_manager/consumers.py
+++ b/api_app/chatbot_manager/consumers.py
@@ -17,11 +17,10 @@ from asgiref.sync import async_to_sync
 from channels.generic.websocket import JsonWebsocketConsumer
 
 from api_app.chatbot_manager.events import (
-    DETAIL_INVALID_MESSAGE,
-    DETAIL_SESSION_NOT_FOUND,
-    ack_payload,
+    AckEvent,
+    ChatErrorDetail,
+    ErrorEvent,
     chat_group_for_user,
-    error_payload,
 )
 from api_app.chatbot_manager.models import ChatSession
 from api_app.chatbot_manager.serializers.chat import MessageRequestSerializer
@@ -46,13 +45,13 @@ class ChatConsumer(JsonWebsocketConsumer):
         self.group_name = chat_group_for_user(user.id)
         self.accept()
         async_to_sync(self.channel_layer.group_add)(self.group_name, self.channel_name)
-        logger.debug("user %s connected to chat group %s", user, self.group_name)
+        logger.debug(f"user {user} connected to chat group {self.group_name}")
 
     def disconnect(self, close_code) -> None:
         group_name = getattr(self, "group_name", None)
         if group_name:
             async_to_sync(self.channel_layer.group_discard)(group_name, self.channel_name)
-        logger.debug("chat ws disconnected (group=%s, code=%s)", group_name, close_code)
+        logger.debug(f"chat ws disconnected (group={group_name}, code={close_code})")
 
     def receive_json(self, content, **kwargs) -> None:
         """Validate one inbound message, resolve its session, and enqueue the agent turn.
@@ -64,7 +63,7 @@ class ChatConsumer(JsonWebsocketConsumer):
         user = self.scope["user"]
         serializer = MessageRequestSerializer(data=content)
         if not serializer.is_valid():
-            self.send_json(error_payload(None, DETAIL_INVALID_MESSAGE))
+            self.send_json(ErrorEvent(None, ChatErrorDetail.INVALID_MESSAGE.value).to_client())
             return
 
         session_id = serializer.validated_data.get("session_id")
@@ -72,11 +71,11 @@ class ChatConsumer(JsonWebsocketConsumer):
         try:
             session = self._resolve_session(user, session_id)
         except ChatSession.DoesNotExist:
-            self.send_json(error_payload(session_id, DETAIL_SESSION_NOT_FOUND))
+            self.send_json(ErrorEvent(session_id, ChatErrorDetail.SESSION_NOT_FOUND.value).to_client())
             return
 
         # Ack the (possibly newly created) session id before the asynchronous stream begins.
-        self.send_json(ack_payload(session.id))
+        self.send_json(AckEvent(session.id).to_client())
         process_chat_message.delay(session.id, message, user.id)
 
     @staticmethod
--- a/api_app/chatbot_manager/consumers.py
+++ b/api_app/chatbot_manager/consumers.py
@@ -1,0 +1,108 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+"""WebSocket consumer for the chatbot.
+
+Mirrors JobConsumer (api_app/websocket.py): a synchronous JsonWebsocketConsumer that joins a
+channel group on connect and relays group messages to the browser. Unlike JobConsumer it is
+bidirectional — the client sends a chat message over the socket and the consumer enqueues the
+Celery turn — and it groups by user (chat-<user_id>) rather than by resource, because the fixed
+URL ws/chat/?context_url=... carries no session id and a session belongs to exactly one user.
+"""
+
+import logging
+from urllib.parse import parse_qs
+
+from asgiref.sync import async_to_sync
+from channels.generic.websocket import JsonWebsocketConsumer
+
+from api_app.chatbot_manager.events import (
+    DETAIL_INVALID_MESSAGE,
+    DETAIL_SESSION_NOT_FOUND,
+    ack_payload,
+    chat_group_for_user,
+    error_payload,
+)
+from api_app.chatbot_manager.models import ChatSession
+from api_app.chatbot_manager.serializers.chat import MessageRequestSerializer
+from api_app.chatbot_manager.tasks import process_chat_message
+
+logger = logging.getLogger(__name__)
+
+
+class ChatConsumer(JsonWebsocketConsumer):
+    """Streams chat turns to the user and accepts inbound messages to start them.
+
+    Auth is handled upstream by WSAuthMiddleware (anonymous users are rejected before reaching
+    here). The group key is the server-authenticated scope["user"].id, so a client can only ever
+    receive its own stream.
+    """
+
+    def connect(self) -> None:
+        user = self.scope["user"]
+        # context_url (the page the user is on) is captured for later context injection (W9);
+        # it is intentionally unused for now.
+        self.context_url = self._parse_context_url()
+        self.group_name = chat_group_for_user(user.id)
+        self.accept()
+        async_to_sync(self.channel_layer.group_add)(self.group_name, self.channel_name)
+        logger.debug("user %s connected to chat group %s", user, self.group_name)
+
+    def disconnect(self, close_code) -> None:
+        group_name = getattr(self, "group_name", None)
+        if group_name:
+            async_to_sync(self.channel_layer.group_discard)(group_name, self.channel_name)
+        logger.debug("chat ws disconnected (group=%s, code=%s)", group_name, close_code)
+
+    def receive_json(self, content, **kwargs) -> None:
+        """Validate one inbound message, resolve its session, and enqueue the agent turn.
+
+        The frame is untrusted: it is validated/length-capped by MessageRequestSerializer, and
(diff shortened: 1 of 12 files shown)


berardifra:
Design note: the chatbot turn runs on a dedicated `celery_worker_chatbot` rather than sharing the default worker's -Q, to isolate the long ReAct/Ollama turns from the default queue -> confirmed offline as the preferred approach.

berardifra:
Verified the WS pipeline live (daphne + a dedicated `celery_worker_chatbot` + ollama/mistral): handshake + session-cookie auth → `ack` → `chat.start` → `chat.status` streaming → graceful `chat.error` on the soft time limit. Pushed one small follow-up: `chat.status` now reports only **registered** tools, so LangChain's internal `_Exception` parse-recovery pseudo-tool no longer leaks to the client (+ a test). 22 chatbot tests green, ruff clean.

api_app/chatbot_manager/consumers.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def connect(self) -> None:
+        user = self.scope["user"]
+        # context_url (the page the user is on) is captured for later context injection (W9);
+        # it is intentionally unused for now.
+        self.context_url = self._parse_context_url()
+        self.group_name = chat_group_for_user(user.id)
+        self.accept()
+        async_to_sync(self.channel_layer.group_add)(self.group_name, self.channel_name)
+        logger.debug("user %s connected to chat group %s", user, self.group_name)
```
mlodic: minor, use f-format strings if possible

api_app/chatbot_manager/consumers.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def disconnect(self, close_code) -> None:
+        group_name = getattr(self, "group_name", None)
+        if group_name:
+            async_to_sync(self.channel_layer.group_discard)(group_name, self.channel_name)
+        logger.debug("chat ws disconnected (group=%s, code=%s)", group_name, close_code)
```
mlodic: also here

api_app/chatbot_manager/events.py
```diff
@@ -0,0 +1,122 @@
...
+
+from dataclasses import dataclass
+from typing import Optional
+
+# --- Channels group_send "type" -> ChatConsumer handler (dots become underscores) ---
+EVENT_START = "chat.start"
+EVENT_STATUS = "chat.status"
+EVENT_TOKEN = "chat.token"
+EVENT_END = "chat.end"
+EVENT_ERROR = "chat.error"
```
mlodic: could these vars be grouped together in an Enum?

api_app/chatbot_manager/events.py
```diff
@@ -0,0 +1,122 @@
...
+
+# --- client-facing payload "type" labels (what the browser switches on) ---
+CLIENT_ACK = "ack"
+CLIENT_START = "start"
+CLIENT_STATUS = "status"
+CLIENT_TOKEN = "token"
+CLIENT_END = "end"
+CLIENT_ERROR = "error"
```
mlodic: same here

api_app/chatbot_manager/events.py
```diff
@@ -0,0 +1,122 @@
...
+ANSWER_MARKER = "Final Answer:"
+
+# Inbound guardrail: mirrors MessageRequestSerializer(message=CharField(max_length=4096)).
+MAX_INBOUND_MESSAGE_LEN = 4096
+
+# User-facing error details (kept generic on purpose: never leak internals to the client).
+DETAIL_SESSION_NOT_FOUND = "Chat session not found."
+DETAIL_INVALID_MESSAGE = "Invalid message payload."
+DETAIL_TIMEOUT = "The assistant took too long to respond. Please try again."
+DETAIL_UNAVAILABLE = "The assistant is currently unavailable. Please try again."
```
mlodic: also here

api_app/chatbot_manager/events.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def chat_group_for_user(user_id: int) -> str:
+    """Return the per-user channel group that carries this user's chat stream."""
+    return f"{CHAT_GROUP_PREFIX}{user_id}"
+
+
+def _payload(client_type: str, **fields) -> dict:
+    return {"type": client_type, **fields}
```
mlodic: I think you already know here what I am about to ask 🤣 

anyway, not important particuarly but still I noticed a different pattern...here you use a function and a dict but later (OutboundEvent) you use a dataclass. Can we choose just one style?

api_app/chatbot_manager/events.py
```diff
@@ -0,0 +1,122 @@
...
+def token_payload(session_id: int, content: str) -> dict:
+    return _payload(CLIENT_TOKEN, session_id=session_id, content=content)
+
+
+def end_payload(session_id: int, message_id: int, content: str) -> dict:
+    return _payload(CLIENT_END, session_id=session_id, message_id=message_id, content=content)
+
+
+def error_payload(session_id: Optional[int], detail: str) -> dict:
+    return _payload(CLIENT_ERROR, session_id=session_id, detail=detail)
```
mlodic: do you think it could make sense to group all these functions together? 

I mean, here I notice a pattern...the same 5 functions (start,status,...) for different cases (payload, event, chat...)...it seems there's chance to create a base class for those standard functions maybe?

mlodic (commented):
I won't dig much on the core, cause it is quite difficult for me to properly understand all the steps required and this is the reason why you are coding here instead of me 🤣 .
However, still I asked some possible adjusts in the code organizations, sometimes I get a bit lost from the different parts

berardifra:
Thanks for the review! Addressed the code-organization points in `1435d11e`:

- **events.py now uses a single style.** The `*_payload` dict-builders and `OutboundEvent` are gone, replaced by one `ChatEvent` dataclass hierarchy (`Ack/Start/Status/Token/End/Error`), each with `to_client()` (the frame the browser receives) and `as_channel_message()` (the `group_send` wrapper relayed by the consumer). This also removes the duplicated 5-function payload/event pattern you spotted.
- **The loose string constants are grouped into enums:** `ChatEventType` (the frame `type`) and `ChatErrorDetail` (the user-facing error texts).
- **f-string logging** in the consumer (and the task, for consistency).

The wire protocol is unchanged (same `type` + payload), so nothing downstream is affected. 22 chatbot tests green, ruff clean.

mlodic (approved):

Issue #3763, opened 2026-06-10
[GSoC 2026] React chat panel (WebSocket streaming UI)


berardifra:
Umbrella issue for the W7-8 React chat UI that drives the chatbot WebSocket (`ws/chat/`, shipped in #3757) from the IntelOwl frontend.

Two PRs `Refs` this issue:

- **PR A — core streaming chat panel**: Offcanvas drawer toggled from the global header, native WebSocket client with token-by-token streaming (`ack`/`start`/`status`/`token`/`end`/`error`), markdown rendering of assistant answers, capped-backoff reconnect + per-turn watchdogs. No backend changes (consumes the fixed wire protocol as-is).
- **PR B — session management**: session list / new / delete / paginated history reload (`GET /api/chatbot/sessions`, `/sessions/{id}/messages`), reusing the same socket (events demuxed by `session_id`).

Scope: native React only, aligned with the existing IntelOwl frontend (reactstrap / zustand / react-icons) — no new component library.

PR #3765, opened 2026-06-11
[GSoC 2026] Add chat sessions list with backend titles and unlocked navigation

berardifra:
# (Part of #3763) Chat Sessions List + Unlocked Navigation

## Description

Refs #3763

This PR implements the Chat Sessions view (W7 PR3) and introduces two major improvements based on maintainer feedback:

### 1. Session title from the backend (`api_app/chatbot_manager`)

- **Backend**: `GET /sessions` now annotates each session with a `title` — the first user message, truncated to 40 chars via a correlated `Subquery` (no N+1, no extra payload). Only runs on the `list` action.
- **Serializer**: `ChatSessionSerializer` exposes `title` (read-only, nullable).
- **Frontend**: `deriveSessionLabel(session)` now takes a single argument — prefers `session.title`, falls back to formatted `created_at`. The old `sessionLabels` cache, `firstUserPreview()`, and the extra store selectors are all removed.
- **3 new backend tests** verify the annotation, 40-char truncation, and null fallback.

### 2. Navigation unlocked during streaming (`frontend/src/stores` & `useChatWebSocket`)

- **Store**: `switchSession` and `newChat` no longer block on `isStreaming`. Both bump `navEpoch` to signal abandonment.
- **WS hook**: The post-ack and max-turn watchdogs now capture `navEpoch` when armed. If the epoch has changed by the time they fire, they silently expire — no stale error fires onto the new conversation.
- **Component**: `ChatSessionList` removes `disabled={isStreaming}` from the "New chat" button and session rows.
- **Tests**: 5 updated + 4 new frontend tests cover the new behavior (63/63 green). 3 new backend tests (15/15 green).

## Type of change

## Checklist

- [x] The pull request is for the branch `develop` *(GSoC: targets the `gsoc-2026/llm-chatbot` umbrella)*
- [x] If external libraries/packages with restrictive licenses were used, they were added in the Legal Notice section
- [x] Linters (`Black`, `Flake8`, `Isort`) gave 0 errors.
- [x] If changes were made to an existing model/serializer/view, the docs were updated and regenerated

### Important rules

### DIFF
--- a/api_app/chatbot_manager/serializers/chat.py
+++ b/api_app/chatbot_manager/serializers/chat.py
@@ -7,10 +7,12 @@ from ..models import ChatMessage, ChatSession
 
 
 class ChatSessionSerializer(serializers.ModelSerializer):
+    title = serializers.CharField(read_only=True, default=None)
+
     class Meta:
         model = ChatSession
-        fields = ["id", "created_at", "updated_at"]
-        read_only_fields = ["id", "created_at", "updated_at"]
+        fields = ["id", "created_at", "updated_at", "title"]
+        read_only_fields = ["id", "created_at", "updated_at", "title"]
 
 
 class ChatMessageSerializer(serializers.ModelSerializer):
--- a/api_app/chatbot_manager/views.py
+++ b/api_app/chatbot_manager/views.py
@@ -3,6 +3,8 @@
 
 import logging
 
+from django.db.models import OuterRef, Subquery
+from django.db.models.functions import Substr
 from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.generics import get_object_or_404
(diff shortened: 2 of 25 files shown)


berardifra:
I will share some images or video about it, soon...

berardifra:
https://github.com/user-attachments/assets/98743f31-f19a-482d-a35e-4afb2559f273

mlodic:
super useful, great job!

mlodic:
I see that you have the content from the previous PR too here, feel free to close that one so we continue here.

mlodic (commented):
aside from what I said in the other PR, I don't have any major concerns for this, it seems working pretty fine. I won't dig too much in the frontend code

berardifra:
> aside from what I said in the other PR, I don't have any major concerns for this, it seems working pretty fine. I won't dig too much in the frontend code

  Thanks for the careful look!

  On the "Connected" badge: you're right, it's misleading. Right now it only reflects the
  WebSocket transport to Daphne, which stays up even when the chatbot Celery worker (or Ollama)
  is down, so it reads "Connected" while a turn can't actually be served. I'll make the badge
  reflect assistant availability, not just the socket: when the worker doesn't pick up a turn
  (the client already has a post-ack watchdog for exactly this case) the badge flips to an
  explicit "Unavailable" state instead of staying green.

  On "what do you see if the extra containers aren't running" / first-time clarity: to make that
  obvious before sending a message, I'll add a small availability check when the drawer opens, so
  the panel shows a clear "assistant not available -> start the ollama/chatbot services" state up
  front rather than a 20s timeout. That needs a tiny backend health endpoint, so I'll do it as a
  focused follow-up PR right after this one and keep this PR to the badge + the info note below.

  On the "i" info badge: yep, I'll add it next to "Assistant" with the same MdInfoOutline
  + UncontrolledTooltip pattern you use for TLP, with a one-line description (privacy-preserving
  assistant, all inference runs locally via Ollama, query your IntelOwl data in natural language).

  I'll push these to this PR and refresh the demo so the new badge behavior is visible.

berardifra:
Badge now reflects assistant availability: "Unavailable" (+ banner) when the worker isn't serving turns, back to "Connected" on recovery (clip: info tooltip → worker down → recovery). Also added the "i" info tooltip and a hover highlight on the "Assistant" header entry, like Docs. The proactive no-container check (before sending) will follow in a separate PR.

Tell me if there is something u would like to change


https://github.com/user-attachments/assets/8bf4be2c-24cf-4bbe-bf52-4948bedc778b

mlodic (approved):

PR #3764, opened 2026-06-10
[GSoC 2026] Add React chat panel with WebSocket streaming (refs #3763)

berardifra:
# (Part of #3763) React chat panel with WebSocket streaming

## Description

Refs #3763

Adds the frontend chat interface that connects to the chatbot WebSocket consumer
(merged in #3757) and renders the agent's streaming responses in real time.

### Components (`frontend/src/components/chat/`)

- **`ChatPanel`**: reactstrap `Offcanvas` drawer mounted once in `AppMain` (survives
  page navigation). Connection-state badge (Idle → Connecting → Connected → Disconnected).
- **`ChatMessageList`**: scrollable conversation with auto-scroll, live "Thinking…" /
  "Running <tool>…" indicator, and a streaming bubble that renders incrementally.
- **`ChatMessage`**: user bubbles (plain text, right-aligned) and assistant bubbles
  (GFM markdown via `react-markdown` + `remark-gfm`, left-aligned). No `rehype-raw`
  (XSS-safe by default).
- **`ChatComposer`**: textarea + send button, Enter sends / Shift+Enter newline, locked
  while a turn is streaming or the socket is not connected. Client-side 4096 char cap
  mirrors `MessageRequestSerializer`.
- **`chatMarkdown`**: isolated GFM renderer (separate from the shared `markdownToHtml` so
  existing markdown rendering is not affected).

### State (`frontend/src/stores/useChatStore.jsx`)

Zustand store following the existing IntelOwl convention (no non-serializable refs in the
store). Owns: drawer visibility, session id, committed messages, streaming text, current
tool, connection state, error text. Reducers mirror the server event types
(`applyAck/Start/Status/Token/End/Error`).

### Transport (`useChatWebSocket` hook)

- Lazy connect on first drawer open; socket stays alive across open/close.
- Capped exponential backoff reconnect (1s → 30s, max 5 attempts).
- Session demux: `start/status/token/end` frames carry `session_id`, the hook drops frames
  that belong to another tab's session.
- Two watchdog timers: post-ack (20s, catches dead chatbot worker) and max-turn (310s,
  catches dropped socket / hard-killed worker).
- Clean teardown on unmount (NORMAL_CLOSURE 1000).

### Integration points

- `AppHeader`: "Assistant" button (chat bubble icon) toggles the drawer for authenticated users.
- `AppMain` Layout: `<ChatPanel />` mounted alongside `<Toast />`.
[...]

PR #3769, opened 2026-06-12
fix(frontend): re-sync package-lock.json with package.json (npm ci)

berardifra:
# Re-sync frontend `package-lock.json` with `package.json`

## Description

Follow-up to the review thread on #3764, where @mlodic confirmed this should be fixed in a separate PR.

`frontend/package-lock.json` on `develop` is out of sync with `frontend/package.json`, so **`npm ci` fails repo-wide**:

```
npm error `npm ci` can only install packages when your package.json and package-lock.json ... are in sync.
npm error Invalid: lock file's @testing-library/dom@9.3.1 does not satisfy @testing-library/dom@10.4.1
npm error Missing: aria-query@5.3.0 from lock file
npm error Invalid: lock file's picocolors@1.0.1 does not satisfy picocolors@1.1.1
```

Concrete drift:
- the lockfile pins `@testing-library/dom@9.3.1` while `package.json` requires `^10.4.1`;
- the lockfile root entry says `"version": "6.4.0"` (package.json is `6.6.1`);
- `http-proxy-middleware` is under `devDependencies` in the lockfile but under `dependencies` in `package.json`.

**Root cause:** the Docker build uses `npm install` (which tolerates drift and silently rewrites the lockfile), so the desync never surfaced there — but `npm ci` (strict, used for reproducible installs / CI) refuses it.

**Fix:** regenerated the lockfile with the build toolchain image so it conforms to `package.json`:

```bash
cd frontend
docker run --rm -v "$PWD":/app -w /app node:lts-alpine3.21 npm install --package-lock-only
```

- `lockfileVersion` stays **2** (no format churn);
- `package.json` is **untouched** — lockfile-only change;
- net diff is small and targeted (root version → 6.6.1, `@testing-library/dom` → 10.4.1 + its deps, `http-proxy-middleware` placement);
- after this, `npm ci --dry-run` succeeds where it previously errored with `EUSAGE`.

No source/behavior change, no dependency added or version bumped — this only makes the lockfile match the already-declared `package.json`.

## Type of change

## Checklist

- [x] If external libraries/packages with restrictive licenses were used, they were added in the Legal Notice section *(none added)*
- [x] Linters (`Black`, `Flake8`, `Isort`) gave 0 errors. *(no Python changed)*
- [ ] I have added tests for the feature/bug I solved. *(N/A — lockfile-only; verified via `npm ci --dry-run`)*
- [x] If changes were made to an existing model/serializer/view, the docs were updated and regenerated *(N/A)*
- [ ] If the GUI has been modified. *(not modified)*

### Important rules

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "intelowl",
-  "version": "6.4.0",
+  "version": "6.6.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "intelowl",
-      "version": "6.4.0",
+      "version": "6.6.1",
       "dependencies": {
         "@certego/certego-ui": "^0.1.19",
         "@dagrejs/dagre": "^1.1.4",
@@ -19,6 +19,7 @@
         "date-fns-tz": "^3.2.0",
         "flag-icons": "^7.2.3",
         "formik": "^2.4.6",
+        "http-proxy-middleware": "^2.0.6",
         "js-cookie": "^3.0.5",
         "md5": "^2.3.0",
         "prop-types": "^15.8.1",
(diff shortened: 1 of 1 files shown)


mlodic (approved):

mlodic:
thank you!

Issue #3770, opened 2026-06-12
[GSoC 2026] W9: Context injection — make the chatbot page-aware


berardifra:
## Goal

Make the chatbot page-aware so that on a job/investigation detail page the user can say "summarize this" and the agent resolves which entity from the WebSocket `context_url`, without the user typing the ID.

## Scope

- Backend-only: URL→prompt-hint helper, prompt template variable, WebSocket wiring (consumer→task→agent)
- No frontend changes (context_url already sent by the FE)
- Hint-only approach — the agent calls existing user-scoped tools to fetch data; no eager-fetch, no new data-access code

## Deliverables

- [x] `derive_page_context` helper in `api_app/chatbot_manager/agent/context.py`
- [x] `{page_context}` prompt variable + all invoke sites updated
- [x] WebSocket wiring: consumer forwards `context_url`, task derives hint
- [x] Tests (11 new): context helper, task injection, consumer forwarding
- [x] PR linked — #3771

## Links

- Branch: `gsoc-2026/llm-chatbot-context-injection`
- PR: #3771
- Plan: `~/.claude/plans/2026-06-12-chatbot-context-injection-plan.md`

Issue #42, opened 2020-03-22
First Use Clarity


weslambert:
It's not really clear on first use what analyzers are available, or how the platform is to be used.  For example, I try to search an MD5 once logged in (eg. Mimikatz) and get no results.  Do I need to enable some more analyzers, or is there something else I must do to populate a local DB?

weslambert:
I realize there is the JSON config file for analyzers, but from a web perspective, it is a little unclear.

mlodic:
Hi @weslambert, thank you for your interest.

I suggest you to go through the docs. https://intelowl.readthedocs.io/en/latest/
At the moment the interface is very minimal and the better way to use this project is to integrate it via API with other security tools, or using the official client manually (https://github.com/mlodic/pyintelowl).
The single page available from the interface allows the user to look for already launched jobs, nothing more. So it is normal that you did not find anything.

However we are planning to create a fresh new interface for the project (see #13) to allow users to interact with Intel Owl manually more easily.

weslambert:
Thanks, that helps to clarify a good deal.  I was already planning on using the Python client, as I would like to call the API from another tool, but just wanted to test locally before doing so.  I'll proceed with the client/API, and let you know if I have any issues.  Thanks for the quick response!

PR #3771, opened 2026-06-12
[GSoC 2026] Inject current-page context into the chatbot agent (refs #3770)

berardifra:
# Inject current-page context into the chatbot agent

## Description

Refs #3770

W9 — **context injection**. When the user opens the chat drawer from a job or investigation detail page, the agent now knows which entity the user is looking at. The user can say "summarize this" or "what analyzers ran on this?" and the agent resolves the ID from the page URL — no need to type it.

**Architecture: hint-only.** A pure helper maps `context_url` → a compact prompt hint (`"The user is currently viewing job #42 in the IntelOwl UI."`); the consumer forwards `context_url` to the Celery task; the task derives the hint and feeds it to the agent prompt via a new `{page_context}` template variable; the agent then calls the existing user-scoped tools. No eager fetch, no new data access, no frontend change. Multi-tenancy is inherited from the tools; prompt-injection is neutralised by rendering only a validated integer id into a fixed template (the raw URL never enters the prompt).

### Backend — `api_app/chatbot_manager`

- **`agent/context.py`** (new) — `derive_page_context(context_url) -> str`: extracts a job or investigation ID from the URL path, returns a compact hint string, or `""` for non-entity pages. Stdlib-only (`re`, `urllib.parse`), no Django/ORM, importable at module top.
- **`agent/agent.py`** — `{page_context}` appended to the system prompt so the agent sees the current page hint. The prompt template is shared by WS + REST; every invoke site must now supply the variable.
- **`views.py`** — REST invoke passes `page_context=""` (context injection is WS-only for now).
- **`tasks.py`** — `process_chat_message` gains `context_url: str = ""`; derives the hint via `derive_page_context` and injects it into the agent invoke.
- **`consumers.py`** — forwards the existing `self.context_url` (already captured from the WS query string) to the task.

### Tests

- **8 new** `test_context.py` — `SimpleTestCase`: job/investigation URL patterns, non-entity pages, empty/malformed URLs, non-numeric IDs, prompt-injection via query/fragment neutralised.
- **2 new** `test_tasks.py` — `ProcessChatMessageTestCase`: context_url injected as page_context hint; missing context_url yields empty.
- **1 new** `test_consumers.py` — `ChatConsumerTestCase`: context_url forwarded to task; existing `delay` assertions updated to include the 4th arg.
- **`test_agent.py`** — real-executor invoke updated to supply `page_context=""`.
- **Full chatbot suite: 95/95 green** (no regressions).

No new dependencies, no model/migration, no frontend change.

## Type of change

## Checklist

- [x] The pull request is for the branch `develop` *(GSoC: targets the `gsoc-2026/llm-chatbot` umbrella)*
- [x] If external libraries/packages with restrictive licenses were used, they were added in the Legal Notice section *(none added)*
- [x] Linters (`Black`, `Flake8`, `Isort`) gave 0 errors.
- [x] If changes were made to an existing model/serializer/view, the docs were updated and regenerated

### Important rules

### DIFF
--- a/api_app/chatbot_manager/agent/context.py
+++ b/api_app/chatbot_manager/agent/context.py
@@ -11,8 +11,19 @@ from urllib.parse import urlparse
 _JOB_CONTEXT = "The user is currently viewing job #{id} in the IntelOwl UI."
 _INVESTIGATION_CONTEXT = "The user is currently viewing investigation #{id} in the IntelOwl UI."
 
-# Frontend detail routes (frontend/src/components/Routes.jsx): job is /jobs/<id>[/...],
-# investigation is /investigation/<id> (singular). Match the id, ignore any trailing segments.
+# These regexes mirror React Router path definitions in
+# frontend/src/components/Routes.jsx. When those routes change, these patterns
+# must be updated in lockstep — otherwise context injection silently stops
+# recognising job/investigation pages. The coupling is enforced by
+# test_regexes_match_frontend_route_definitions in test_context.py.
+#
+#   /jobs/:id                                   (Routes.jsx:157)
+#   /jobs/:id/:section                          (Routes.jsx:166)
+#   /jobs/:id/:section/:subSection              (Routes.jsx:178)
+#   /jobs/:id/comments                          (Routes.jsx:186)
+#   /investigation/:id                          (Routes.jsx:243)
+#
+# Only the numeric id is captured; any trailing segments are ignored.
 _JOB_RE = re.compile(r"^/jobs/(\d+)(?:/|$)")
 _INVESTIGATION_RE = re.compile(r"^/investigation/(\d+)(?:/|$)")
 
--- a/api_app/chatbot_manager/agent/context.py
+++ b/api_app/chatbot_manager/agent/context.py
@@ -1,0 +1,36 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+import re
+from urllib.parse import urlparse
+
+# Fixed hint templates. context_url is client-supplied, so we render ONLY a validated integer id
+# into these constants — the raw URL text never reaches the prompt. A crafted URL such as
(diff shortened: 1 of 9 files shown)


api_app/chatbot_manager/agent/context.py
```diff
@@ -0,0 +1,36 @@
...
+# into these constants — the raw URL text never reaches the prompt. A crafted URL such as
+# /jobs/42?x=ignore+previous+instructions therefore yields exactly "...job #42..." and nothing
+# else, which neutralises prompt injection via the URL.
+_JOB_CONTEXT = "The user is currently viewing job #{id} in the IntelOwl UI."
+_INVESTIGATION_CONTEXT = "The user is currently viewing investigation #{id} in the IntelOwl UI."
+
+# Frontend detail routes (frontend/src/components/Routes.jsx): job is /jobs/<id>[/...],
+# investigation is /investigation/<id> (singular). Match the id, ignore any trailing segments.
+_JOB_RE = re.compile(r"^/jobs/(\d+)(?:/|$)")
+_INVESTIGATION_RE = re.compile(r"^/investigation/(\d+)(?:/|$)")
```
mlodic, line 28: this poses possible maintenance issues in case these URLs are changed from the code that handle these endpoints. The person working there could not be aware of this. Could you check whether there's a chance to get this info directly from where these routes are created?

berardifra: These are React Router paths defined in `frontend/src/components/Routes.jsx `-> there's no backend URL config to import from (the Django router doesn't know about SPA routes like `/jobs/:id`).
To mitigate the maintenance risk I've added:
- explicit line-number references to Routes.jsx in the comment (lines 157, 166, 178, 186, 243)
- a `test_regexes_match_frontend_route_definitions` coupling test that verifies the regexes against the known route patterns, so if someone renames a route without updating context.py, the test fails immediately

This is the tightest coupling we can achieve without a shared source between the frontend and backend.

mlodic (approved):

Issue #3834, opened 2026-07-01
[GSoC 2026] Run --ollama in CI and gate the slow build to develop→master


berardifra:
## Summary

Make CI exercise the chatbot stack with `--ollama`, and run the heavy **"Startup script launch (Slow)"** step only on the **develop → master** release PR instead of on every PR that targets master.

## Motivation

The chatbot landed in `develop` (#3831). CI never starts the chatbot topology, so a regression in the ollama/`celery_worker_chatbot` wiring (exactly the class of bug fixed in #3829/#3830) would go undetected until a manual run. Running `--ollama` on the release build gives a deployment smoke test at the point it matters (cutting a release), while keeping per-PR develop builds fast (no model pull).

## Blocking prerequisite: `--ollama` doesn't work in `ci` mode

Same root cause as #3829 in `test` mode. `--ollama` adds `docker/ollama.override.yml`, whose `celery_worker_chatbot` is pinned to the pulled release image (`intelowlproject/intelowl:${REACT_APP_INTELOWL_VERSION}`) with no source mount. In `ci` mode every other service is repointed to the freshly built `intelowlproject/intelowl:ci` by `docker/ci.override.yml`, but the chatbot worker is not — so `./start ci up --ollama` would run stale released code (and crash on the missing `celery_chatbot.sh` before a release ships the chatbot). The `test`-mode fix (`docker/test.ollama.override.yml`) is added only for `env_argument == "test"`, never for `ci`.

## Changes

- **New `docker/ci.ollama.override.yml`**: repoint `celery_worker_chatbot` to `intelowlproject/intelowl:ci` + `env_file_app_ci` (mirrors `ci.override.yml`; no source mount).
- **`start`**: register `ci_ollama` in `path_mapping`; in `ci` mode, append `ci.ollama.override.yml` when `--ollama` is set (mirrors the existing `test` block).
- **`.github/workflows/pull_request_automation.yml`**:
  - Slow step gated to `github.base_ref == 'master' && github.head_ref == 'develop'`; command adds `--ollama`.
  - Fast step becomes the exact complement so every other PR (incl. a non-develop head into master) still gets a container up.

## Testing

Validated locally that the `start` compose assembly appends `ci.ollama.override.yml` after `ollama.override.yml` for `ci up --ollama` (and only then), with no regression to `test up --ollama`. Workflow conditions validated on a fork (`develop → master` triggers the Slow+ollama path; other PRs trigger Fast) before opening this PR.

[...]

PR #3836, opened 2026-07-01
[GSoC 2026] fix(chatbot): round up rate limiter retry_after

### LINKED ISSUE #3835: [GSoC 2026] Chatbot rate limiter returns retry_after=0 at end of window (flaky test)
## Describe the bug

`RateLimiter.allow()` (`api_app/chatbot_manager/rate_limit.py`) can return `retry_after == 0` when called in the final second of the rate-limit window. It computes:

```python
elapsed = time.time() % self.window_seconds
retry_after = int(self.window_seconds - elapsed)
```

`int()` truncates toward zero, so when `window_seconds - elapsed` is between 0 and 1 (the last second of the window), `retry_after` becomes `0` — telling the client to retry immediately even though it is still rate-limited. It also always under-estimates the remaining wait.

## Impact

- Wrong `Retry-After` hint returned by `POST /api/chatbot/message` (429) and the WebSocket path at the window boundary.
- Flaky test: `tests.api_app.chatbot_manager.test_rate_limit.RateLimiterAllowTests.test_allow_returns_false_at_limit` asserts `retry_after > 0` and fails ~1.7% of runs (whenever `time.time() % 60` lands in `[59, 60)`). Observed intermittently in CI.

## Fix

Round up instead of truncating: `retry_after = math.ceil(self.window_seconds - elapsed)`. Since `elapsed ∈ [0, window)`, `window - elapsed ∈ (0, window]`, so `ceil` is always ≥ 1 and never under-estimates. Add a regression test that freezes time at the end of the window.

## Notes

Discovered while validating the CI `--ollama` change (#3834) on a fork; unrelated to that change. Present on `develop` since the chatbot merge (#3831).

berardifra:
Closes #3835. allow() could return retry_after=0 in the last second of the window (int() truncation), causing a flaky test. Fixed with math.ceil + regression test. Found while validating #3834 on a fork.

### DIFF
--- a/api_app/chatbot_manager/rate_limit.py
+++ b/api_app/chatbot_manager/rate_limit.py
@@ -8,6 +8,7 @@ shared between the REST view and the WebSocket consumer so a user who sends via
 both paths hits the same bucket.
 """
 
+import math
 import time
 
 from django.core.cache import caches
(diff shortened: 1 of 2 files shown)


mlodic (approved):

mlodic:
I think we can directly merge this to develop as of now instead of the development branch considering this is a fix

Issue #3843, opened 2026-07-03
[GSoC 2026] Chatbot: warm KV-cache tool-selection flip picks the wrong tool ('show my recent jobs' → list_investigations)


berardifra:
## Summary

On the LangChain 1.x chatbot, the query **"show my recent jobs"** picks the **wrong tool when the
model is warm** and returns a **wrong answer**. Warm, `qwen2.5:3b` calls only `list_investigations`
(empty for the test user) and replies *"There are no investigations that match your filters."* — even
though the user has 3 jobs. Cold (model just loaded), the same query runs the correct
`list_investigations` → `search_jobs` path and returns the 3 real jobs.

This is a **correctness / reliability** issue, **not a latency** one (see below). It is a new
manifestation of the tool-calling brittleness first documented in the 2026-06-21 latency benchmark
(**Finding 1**): at `temperature=0` the only difference between warm and cold is Ollama's
**KV-prefix-cache state vs a full recompute**, which tips a borderline tool-selection argmax. PR #3809
(`handle_validation_error` / `on_invalid_tool_args`) made schema-invalid **tool arguments**
recoverable, but it does not touch **which tool** the model selects — so this failure mode is still
open.

## Reproduction (from the 2026-07-03 warm-latency benchmark, `qwen2.5:3b`, `temperature=0`)

Query: **"show my recent jobs"** · user `admin` (3 seeded jobs, 0 investigations).

| condition | tool rounds | tools called | total | answer |
|---|---|---|---|---|
| **warm** (model resident) | **1** | `list_investigations` | ~6.6 s | ❌ *"There are no investigations that match your filters."* (17 chunks) — **wrong** |
| **cold** (model just loaded) | **2** | `list_investigations`, `search_jobs` | ~114.7 s | ✅ the 3 real jobs (174 chunks) — **correct** |

- **Deterministic across all reps**: warm **5/5** → 1 round, wrong answer; cold **3/3** → 2 rounds,
  correct answer. `temperature=0`, so this is reproducible, not a sampling fluke.
- The mechanism is the same as the 2026-06-21 **Finding 1** (there, warm tipped `summarize_job` to emit
  the literal `'<job_id>'`; here, warm tips *"show my jobs"* onto `list_investigations` instead of
  `search_jobs`).
- The system prompt already disambiguates the two tools
  (`api_app/chatbot_manager/agent/system_prompt.txt`: *search_jobs — Use for "show me jobs"*;
[...]

PR #3844, opened 2026-07-04
[GSoC 2026] chatbot: harden jobs-vs-investigations routing + accept null filter args (Refs #3843)

berardifra:
Refs #3843.

## What
Two coupled changes; the second is the null-arg gap the first exposed.

1. **Routing (prompt).** The local 3B agent (`qwen2.5:3b`, temp=0) mis-routed several "jobs" phrasings
   to `list_investigations`. Strengthen the `search_jobs` side of `system_prompt.txt` only (broaden the
   cue with the failing phrasings incl. the verbatim "what jobs do I have?"; one `[Rules]`
   disambiguation line). `list_investigations` cue untouched, so its routing controls are unaffected.
2. **Null filter args (tools).** With routing fixed, the model calls `search_jobs(query=None, ...)`;
   the `query: str = ""` / `status: str = ""` signatures rejected the explicit `None`, so the answer
   was still empty. Widen `query`/`status` to `Optional[str] = None` on `search_jobs` **and**
   `list_investigations` (signature only — the bodies already guard with `if query:` / `if status:`,
   and `visible_for_user(user)` scoping is unchanged).

## Evidence (same-session A/B/C, tool-selection reliability harness)
Correct = expected tool called; "delivered" = the actual job list was returned. Warm n=8, cold n=2.

| phrasing | before: tool / jobs | after: tool / jobs |
|---|---|---|
| **"what jobs do I have?"** | **0/10 / 0/10** | **10/10 / 10/10** |
| "show my recent jobs" | 2/10 / 2/10 | 10/10 / 10/10 |
| "show me my jobs" | 8/10 / 8/10 | 10/10 / 10/10 |
| "list my jobs" / "find my recent jobs" | 10/10 / 10/10 | 10/10 / 10/10 |
| investigations controls; analyze_observable (M-1) / summarize_job controls | pass | pass (unchanged) |

Zero regressions; all controls hold; `analyze_observable` keeps its M-1 `action_required` preview.
171 chatbot unit tests green.

## Data-isolation note
The tool change is a type-only widening of already-optional filter args; the querysets stay scoped by
`visible_for_user(user)` and the bodies are unchanged, so a `null` filter behaves exactly like an
empty one (no widening). New tool tests assert isolation still holds under `query=None` / `status=None`
(another user's RED job / a non-shared investigation stay hidden). `agent-data-isolation-reviewer` was
not spawnable in the authoring session; the review was done inline.

### DIFF
--- a/api_app/chatbot_manager/agent/system_prompt.txt
+++ b/api_app/chatbot_manager/agent/system_prompt.txt
@@ -3,7 +3,7 @@ You are IntelOwl AI, a concise assistant for the IntelOwl threat intelligence pl
 Answer with short, data-driven summaries. Always cite the tools you used at the end.
 
 [Tools — when to use each]
-- search_jobs: find jobs by observable value, tag, or md5. Use for "show me jobs", "find jobs for domain X".
+- search_jobs: find or list the current user's own jobs — by observable value, tag, or md5, or with no filter to list their recent jobs. Use for any question about the user's jobs: "show me my jobs", "list my jobs", "what jobs do I have?", "do I have any jobs?", "find my recent jobs", "find jobs for domain X".
 - get_job_details: full detail of ONE job by id. Use after search_jobs, or when the user asks "details of job #N".
 - summarize_job: high-level summary of ONE job (status, analyzers, findings). Use for "summarize job #N".
 - list_investigations: browse investigations with optional status/name filters. Use for "show my investigations", "what investigations are open?".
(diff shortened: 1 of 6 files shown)


mlodic (approved):

berardifra:
Thanks for the review! CI is green and the branch is up to date with develop. I don't have merge rights on develop — could you squash-merge when you have a moment?

Issue #3892, opened 2026-08-02
[GSoC 2026] Populate the reconciled DataModel verdict for key-free domain/URL analyzers


berardifra:
## Summary

Several key-free domain/URL analyzers already compute a malicious/benign signal on every default scan, but
throw it away because they carry no `DataModel` hook. As a result their verdict never reaches the
visualizer badge, the reconciliation engine, pivots, or the chatbot — `job.data_model.evaluation` stays
empty for observables that were, in fact, evaluated.

This tracks the **core-first** work to close that gap: make those analyzers populate a reconciled
`evaluation` + `reliability` on their `DataModel`, using the mechanism IntelOwl already supports
(`mapping_data_model` + `_do_create_data_model` / `_update_data_model`). Once populated, every consumer —
badge, engine, pivots, and (in a follow-up) the LLM chatbot — reads one verdict, with no semantic fork.

## Motivation

This is **PR B** of the chatbot *result-interpretation* effort (mentor feedback from @mlodic): the analyst
wants an objective reading of whether an observable is malicious. The verdict must be the core's own
reconciled verdict, so the fix belongs in the core, not in a parallel chatbot extractor.

## Scope

Analyzers that run key-free on a default deploy and produce a verdict-shaped signal:

- **10 DNS "malicious detector" analyzers** sharing the `malicious_detector_response` shape — `AdGuard`,
  `Quad9`, `CloudFlare`, `CleanBrowsing`, `UltraDNS`, `DNS4EU`, `Mullvad` (malicious mode), `Spamhaus_WQS`,
  `GoogleSafebrowsing`, `GoogleWebRisk` → `evaluation=malicious` **only on a real hit**.
- **PhishingArmy**, **Phishstats** → `malicious` on a real listing.
- **Phishtank** → `malicious`, higher reliability when community-verified.
- **Tranco** → `trusted` (popularity), the one benign-direction source, reliability banded by rank and
  capped so it never overrides a malicious hit.

Plus a shared `classify()` helper (extracted from the DataModel visualizer) so the badge and the chatbot
resolve identical buckets.

**Safety invariant:** a `$`-prefixed `mapping_data_model` key writes its constant unconditionally, so the
per-analyzer `_do_create_data_model` gate must suppress non-hits (a miss/timeout/failure → no DataModel,
never `trusted`) — otherwise a clean lookup would be stamped MALICIOUS platform-wide.

## Out of scope (follow-ups)

[...]

PR #3893, opened 2026-08-02
[GSoC 2026] core(datamodel): populate the reconciled verdict for key-free domain/URL analyzers

berardifra:
# Description

Populate `job.data_model` with a reconciled **verdict** (`evaluation` + `reliability`) for the key-free
domain/URL analyzers that today produce a malicious/benign signal but throw it away because they carry no
DataModel hook. Once populated, the same verdict is read by the visualizer badge, the reconciliation
engine, pivots, and (in a follow-up chatbot PR) the assistant — one source of truth, no semantic fork.

This is the **core-first** part of the chatbot *result-interpretation* effort (PR B of the split; the
chatbot reader that consumes this lands separately). `Refs #3892`

**What changes**
- **Shared `classify()`** extracted from the DataModel visualizer into `api_app/data_model_manager/classify.py`
  (single source of the five presentation buckets + the `8`/`6` thresholds), with the bucket names in a
  `DataModelVerdictBuckets(Choices)` enum in `data_model_manager/enums.py` whose `TRUSTED`/`MALICIOUS`
  members derive from `DataModelEvaluations`; the visualizer now calls it — behaviour-preserving.
- **10 key-free DNS "malicious detector" analyzers** (`AdGuard`, `Quad9`, `CloudFlare`, `CleanBrowsing`,
  `UltraDNS`, `DNS4EU`, `Mullvad`, `Spamhaus_WQS`, `GoogleSafebrowsing`, `GoogleWebRisk`) contribute
  `evaluation=malicious` **only on a real hit**, via a shared `_do_create_data_model` gate mixin + a
  declarative `mapping_data_model` migration (`0195`).
- **PhishingArmy / Phishstats** — same declarative pattern gated on a real listing (`0196`).
- **Phishtank** — `evaluation=malicious`, reliability **8** verified / **5** unverified (`_update_data_model`).
- **Tranco** — the one benign-direction source: popularity → `evaluation=trusted`, reliability banded by
  rank, with the top 1000 treated as a reliable allowlist (`_update_data_model`).

**The load-bearing safety point:** a `$`-prefixed `mapping_data_model` key writes its constant
*unconditionally* whenever a data model is created (`analyzers_manager/models.py:92-95`), so
`{"$malicious":"evaluation"}` would stamp MALICIOUS on **every clean lookup**. Each analyzer's
`_do_create_data_model` gate is what suppresses non-hits (miss / timeout / failure → **no** data model,
never `trusted`). This is unit-tested per analyzer, incl. the critical clean-lookup case.

**Reliability table (single reviewable source; `reliability` = trust in the source):**

| source (on a hit) | evaluation | reliability | bucket |
|---|---|---|---|
| GoogleSafebrowsing, GoogleWebRisk | malicious | 8 | malicious |
| Spamhaus_WQS | malicious | 7 | malicious |
| AdGuard, Quad9, CloudFlare, CleanBrowsing, UltraDNS, DNS4EU, Mullvad | malicious | 6 | malicious |
| Phishtank (verified) | malicious | 8 | malicious |
| Phishtank (unverified) | malicious | 5 | suspicious |
| PhishingArmy, Phishstats | malicious | 6 | malicious |
| Tranco rank ≤ 1k | trusted | 9 | **trusted** |
| Tranco rank ≤ 10k / ≤ 100k / ranked | trusted | 4 / 3 / 2 | clean |
| any miss / timeout / note | — (no DataModel) | — | — |

Two regimes, following review feedback:

- **Top 1000 (reliability 9 → `trusted`)** — treated as a reliable allowlist, and *intended* to outrank a
  malicious hit under the engine's average-reliability reconciliation. Rationale (@mlodic): in daily
  incident response false positives are the expensive failure mode, because the analyst time they burn
  costs more than the rare true positive they hide, so reliable allowlists must be enforced where
  available. This puts Tranco's top band at the same reliability as the pre-existing `HuntingAbuseAPI`
  allowlist (9).
- **Below the top 1000 (reliability ≤4 → `clean`)** — popularity is weak positive evidence: every malicious
  reliability (≥5) stays above it, so a *popular-but-flagged* domain still resolves to malicious.

The malicious tiers (8/7/6) reflect source authority and can be flattened on request.

**Notes for reviewers (pre-empting two accurate observations from self-review):**
- "Tranco is the only trusted source" is scoped to the verdicts *this PR adds* — `HuntingAbuseAPI` already
  maps `$trusted`/reliability 9 (allowlist, `0158`) and is untouched here; it sits above the malicious
  range by design and does not affect the two regimes above.
- Pre-existing engine mechanics (informational, not a regression): metadata-only DataModels with a `None`
  evaluation default to reliability 5; a *lone* Tranco `trusted` from the lower bands (2–4) can be
  out-averaged by such a group and reconcile to `no evaluation` rather than `clean`. This is orthogonal
  to the malicious guarantee (malicious ≥6 always wins) and is existing `EvaluationEngineModule`
  behaviour, surfaced here for transparency.

## Type of change

- [x] Chore (behaviour-preserving refactor: shared `classify()` extraction).

# Checklist

- [x] A new plugin (analyzer, connector, …) was added or changed, in which case:
    - [x] I created the corresponding `DataModel` population for the changed analyzers following the
      documentation (declarative `mapping_data_model` + `_do_create_data_model`/`_update_data_model` hooks,
      mirroring the URLhaus / HuntingAbuseAPI precedents). *No new analyzer was added — existing key-free
      analyzers gain DataModel hooks; the plugin sub-items about new-analyzer samples/`FREE_TO_USE`/`url`
      attributes do not apply.*
    - [x] Unit tests were added for every changed analyzer with all external calls mocked (no network in
      tests).
- [x] I have inserted the copyright banner at the start of every new file.
- [x] No new libraries were added.
- [x] Linters (`Ruff`) gave 0 errors (`check` + `format --check` clean on all 25 files).
- [x] I have added tests for the change (6 new test modules); all tests (new and old) gave 0 errors locally
  against a rebuilt test image (`makemigrations --check` clean; affected suite green, incl.
  `engines_manager` + `visualizers_manager` regression checks).
- [ ] GUI modified — N/A (the visualizer refactor is backend, behaviour-preserving; no rendered output
  changes).
- [ ] I will address any `DeepSource` / `Django Doctors` alerts raised on the PR.
- [ ] I will address raised Copilot issues.
- [x] **I have used LLMs (Claude Code) to help implement this PR, and I have reviewed and verified the
  generated code** (design, per-analyzer implementation, and tests were reviewed; behaviour verified via the
  local test suite).

### DIFF
--- a/api_app/analyzers_manager/observable_analyzers/tranco.py
+++ b/api_app/analyzers_manager/observable_analyzers/tranco.py
@@ -7,6 +7,16 @@ import requests
 
 from api_app.analyzers_manager import classes
 from api_app.choices import Classification
+from api_app.data_model_manager.enums import DataModelEvaluations
+
+# Popularity is weak positive evidence, so Tranco's reliability caps at 4 —
+# strictly below the malicious detectors (6–8) so a popular-but-flagged domain
+# still reconciles to malicious, and it never reaches the "trusted" bucket (>=8).
+_RANK_TOP = 10_000
+_RANK_POPULAR = 100_000
+_RELIABILITY_TOP = 4
+_RELIABILITY_POPULAR = 3
+_RELIABILITY_RANKED = 2
 
 
 class Tranco(classes.ObservableAnalyzer):
@@ -26,3 +36,18 @@ class Tranco(classes.ObservableAnalyzer):
         response.raise_for_status()
 
         return response.json()
+
+    def _do_create_data_model(self) -> bool:
+        rank = self.report.report.get("rank")
+        return super()._do_create_data_model() and isinstance(rank, int) and rank > 0
+
+    def _update_data_model(self, data_model) -> None:
+        super()._update_data_model(data_model)
+        rank = self.report.report.get("rank")
+        data_model.evaluation = DataModelEvaluations.TRUSTED.value
+        if rank <= _RANK_TOP:
+            data_model.reliability = _RELIABILITY_TOP
+        elif rank <= _RANK_POPULAR:
+            data_model.reliability = _RELIABILITY_POPULAR
+        else:
+            data_model.reliability = _RELIABILITY_RANKED
--- a/api_app/analyzers_manager/observable_analyzers/tranco.py
+++ b/api_app/analyzers_manager/observable_analyzers/tranco.py
@@ -9,11 +9,18 @@ from api_app.analyzers_manager import classes
 from api_app.choices import Classification
 from api_app.data_model_manager.enums import DataModelEvaluations
 
-# Popularity is weak positive evidence, so Tranco's reliability caps at 4 —
-# strictly below the malicious detectors (6–8) so a popular-but-flagged domain
-# still reconciles to malicious, and it never reaches the "trusted" bucket (>=8).
+# A top-1000 rank is treated as a reliable allowlist rather than as weak popularity
+# evidence: reaching the "trusted" bucket (>=8) means such a domain outranks a
+# malicious detector instead of reconciling to malicious. That is deliberate — in daily
+# incident response, false positives are the expensive failure mode, because the
+# analyst time they burn costs more than the rare true positive they hide.
+# Below the top 1000 popularity really is weak evidence, so reliability stays capped at
+# 4, strictly under the malicious detectors (6-8), and a flagged domain still
+# reconciles to malicious.
+_RANK_HIGHLY_TRUSTED = 1_000
 _RANK_TOP = 10_000
 _RANK_POPULAR = 100_000
+_RELIABILITY_HIGHLY_TRUSTED = 9
 _RELIABILITY_TOP = 4
 _RELIABILITY_POPULAR = 3
 _RELIABILITY_RANKED = 2
@@ -45,7 +52,9 @@ class Tranco(classes.ObservableAnalyzer):
         super()._update_data_model(data_model)
         rank = self.report.report.get("rank")
         data_model.evaluation = DataModelEvaluations.TRUSTED.value
-        if rank <= _RANK_TOP:
+        if rank <= _RANK_HIGHLY_TRUSTED:
+            data_model.reliability = _RELIABILITY_HIGHLY_TRUSTED
+        elif rank <= _RANK_TOP:
             data_model.reliability = _RELIABILITY_TOP
         elif rank <= _RANK_POPULAR:
             data_model.reliability = _RELIABILITY_POPULAR
--- a/api_app/data_model_manager/classify.py
+++ b/api_app/data_model_manager/classify.py
@@ -1,0 +1,31 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+from api_app.data_model_manager.enums import DataModelEvaluations
+
+# Presentation buckets — the single source of truth consumed by both the
+# DataModel visualizer and (later) the chatbot, so every surface says the same
+# word for the same (evaluation, reliability) pair.
+BUCKET_TRUSTED = "trusted"
+BUCKET_CLEAN = "clean"
+BUCKET_MALICIOUS = "malicious"
+BUCKET_SUSPICIOUS = "suspicious"
+BUCKET_NO_EVALUATION = "no evaluation"
+
+# Bucket boundaries (verbatim from the pre-existing visualizer logic this
+# function replaced).
+TRUSTED_RELIABILITY_FLOOR = 8
+MALICIOUS_RELIABILITY_FLOOR = 6
+
+
+def classify(evaluation: str | None, reliability: int) -> str:
+    """Map a (evaluation, reliability) pair to one of the five presentation buckets.
+
+    Single source of truth for the bucketing: the DataModel visualizer calls this
+    (and the chatbot will), so the badge and the chat always agree.
+    """
+    if evaluation == DataModelEvaluations.TRUSTED.value:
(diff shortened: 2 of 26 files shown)


api_app/analyzers_manager/observable_analyzers/tranco.py
```diff
@@ -7,6 +7,16 @@
 
 from api_app.analyzers_manager import classes
 from api_app.choices import Classification
+from api_app.data_model_manager.enums import DataModelEvaluations
+
+# Popularity is weak positive evidence, so Tranco's reliability caps at 4 —
+# strictly below the malicious detectors (6–8) so a popular-but-flagged domain
+# still reconciles to malicious, and it never reaches the "trusted" bucket (>=8).
```
mlodic: I have a strong personal opinion about this from my experience as incident responder.
False positives are really a big issue in the daily job and must be filtered out as much as possible. 
While preserving malicious hits should increase the true positives ratio, having too many FPs to handle could result in less incidents handled properly due to analysts' burnout: the result would be worst. This tool original purpose was to aid analysts to sort out information from a lot of data and, broadly speaking ,save time. So it is highly critical to enforce reliable allowlists when available, to avoid unenecessary work.
Tranco may seem weak evidence but has really high positive value in daily operations and I think it should be scored as 9 instead in case the rank is in the top 1000.
Then, we can make it weaker for the other cases, as you already did.

api_app/data_model_manager/classify.py
```diff
@@ -0,0 +1,31 @@
...
+from api_app.data_model_manager.enums import DataModelEvaluations
+
+# Presentation buckets — the single source of truth consumed by both the
+# DataModel visualizer and (later) the chatbot, so every surface says the same
+# word for the same (evaluation, reliability) pair.
+BUCKET_TRUSTED = "trusted"
+BUCKET_CLEAN = "clean"
+BUCKET_MALICIOUS = "malicious"
+BUCKET_SUSPICIOUS = "suspicious"
+BUCKET_NO_EVALUATION = "no evaluation"
```
mlodic: is it possible to reuse the existing evalution enum here?

mlodic (changes requested):
Could you share a screenshot of a working execution of a DNS playbook analysis to show how data models are populated in the raw results of at least one of the changed analyzers?

api_app/analyzers_manager/observable_analyzers/tranco.py
berardifra: Agreed, and thanks for the operational framing. Applied in fbaf00b6: a new top band, rank ≤ 1000 → reliability 9, which clears the trusted floor (8). The lower bands are unchanged (≤10k → 4, ≤100k → 3, ranked → 2), so popularity on its own stays weak evidence exactly as before. 

Worth flagging explicitly, since it reverses something the code used to assert: the original comment justified the cap at 4 by saying Tranco "never reaches the trusted bucket (>=8)", so that a popular-but-flagged domain would still reconcile to malicious. With the new band that is deliberately no longer true for the top 1000 , such a domain now outranks a malicious detector in the engine's reconciliation. I rewrote the comment to state your reasoning instead of the old one, so the next reader sees why the tradeoff is intentional rather than assuming it's a bug. 

`test_rank_bands` now pins both sides of every boundary (1, 1000, 1001, 10000, 10001, 100000, 100001) so a future edit can't silently widen or shrink the allowlist band.

api_app/data_model_manager/classify.py
berardifra: Partly, yes and the part that isn't reusable is worth being explicit about. `DataModelEvaluations` has only trusted and malicious, but there are five buckets. The other three aren't evaluations: clean and suspicious mean "that evaluation, but below its reliability floor", and no evaluation means no analyzer expressed an opinion at all. They describe how a verdict is shown, not what an analyzer concluded, so adding them to `DataModelEvaluations` would let something that isn't a valid evaluation be assigned to `data_model.evaluation.`

So in 48dc6f2e I added `DataModelVerdictBuckets(Choices)` next to `DataModelEvaluations `in `data_model_manager/enums.py` (enums live there rather than in `classify.py`), with the two overlapping members taking their values from `DataModelEvaluations`, so the two vocabularies can't drift apart:

```
class DataModelVerdictBuckets(Choices):
     TRUSTED = DataModelEvaluations.TRUSTED.value
     CLEAN = "clean"
     MALICIOUS = DataModelEvaluations.MALICIOUS.value
     SUSPICIOUS = "suspicious"
     NO_EVALUATION = "no evaluation"
```

All the BUCKET_* module constants are gone; classify() and the DataModel visualizer now use the enum. No migration, it backs no model field.

If you'd rather have all five live in DataModelEvaluations and treat the extra three as valid evaluation values, say so and I'll switch, I went this way to keep the evaluation field's domain closed.

berardifra:
> Could you share a screenshot of a working execution of a DNS playbook analysis to show how data models are populated in the raw results of at least one of the changed analyzers?

Here's a real run of the Dns playbook on `malware.testcategory.com` (job #35, 12/12 analyzers SUCCESS), with the `CloudFlare_Malicious_Detector` report expanded in the raw results:

pr-b-datamodel-dns-job35

The analyzer's report says "malicious": true, and the data_model it produced carries "evaluation": "malicious" with "reliability": 6, the value mapped for this analyzer in migration 0195.

The same screenshot also shows the gate working: Quad9_Malicious_Detector, UltraDNS_Malicious_Detector and DNS4EU_Malicious_Detector all answered "malicious": false on the same domain and produced no data model at all, so a clean lookup is never stamped malicious. The job's reconciled verdict is malicious / 6.

pr-b-datamodel-dns-job35-reconciled

mlodic (approved):
thanks for the efforts and the explanations, wonderful result

Issue #3895, opened 2026-08-03
[GSoC 2026] Give the chatbot an objective verdict on a job, not an LLM opinion


berardifra:
### What

The chatbot cannot tell an analyst what an analysis actually concluded. `summarize_job` returns job
metadata — status, analyzers, timings — and nothing about the findings, so the two job quick-action
chips, "Summarize this job" and "Evaluate results", produce the same answer: the second one only
changes the phrasing, not the data the model sees.

Reported by @mlodic on Discord in July, together with two other issues (the other two are addressed
separately).

### Why it matters

An analyst asking *"is this malicious?"* gets a restatement of the job status. Worse, any
verdict-shaped wording in the reply is the LLM's own opinion: nothing in the tool payload carries
IntelOwl's assessment, so the model has no grounding and the answer can contradict the badge shown
on the very same job page.

### What should happen

The chatbot should report **IntelOwl's own reconciled evaluation** — the same verdict the DataModel
visualizer badge shows — together with the analyzer evidence behind it, and say so honestly when
there is none, instead of inventing one.

Concretely:

- the verdict must come from the platform's reconciliation (`EvaluationEngineModule` +
  the shared `classify()`), never from a chatbot-side scale or an LLM judgement;
- the analyzers that ran should be attributable: which ones support the verdict, which contradict it,
  and which expressed no opinion at all;
- a job with no evaluation (or a `generic` observable, which IntelOwl does not evaluate) must be
  reported as such, with the reason;
- multi-tenancy must be unchanged: the verdict is only ever read for a job the requesting user can
  already see.

### Notes

Depends on #3892 / #3893, which populate the DataModels the verdict is read from — without those,
the key-free domain/URL analyzers produce no evaluation to reconcile.

`Refs #3892`

PR #3898, opened 2026-08-03
[GSoC 2026] chatbot: report IntelOwl's own verdict on a job

berardifra:
## Description

`summarize_job` now returns IntelOwl's **own reconciled verdict** on the job's observable alongside
the metadata it already reported, plus the analyzer evidence behind it. This closes the third of the
bugs reported on Discord in July: "Summarize job" and "Evaluate job" returned identical output
because no chatbot tool ever read the findings.

The verdict is **not** a chatbot-side scale. The headline is obtained by running the platform's own
`EvaluationEngineModule(job).run()` and mapping it through the shared `classify()` helper, so the
chatbot says exactly the word the job-page badge shows. The LLM is never involved in producing it.

### How it works

- **New reader** `api_app/chatbot_manager/evaluation.py` — a pure function `evaluate_job(job)`
  returning a `JobEvaluation` dataclass. It performs **zero `Job` queries**: the caller passes a job
  already scoped through `visible_for_user`, so the tool keeps the single tenancy boundary it had.
- **Live recompute, not the stored scalar.** The engine modules run asynchronously *after* the
  pipeline saves a transient un-reconciled merge, so `job.data_model` can be wrong for a window.
  Recomputing removes that race by construction and guarantees the chatbot cannot diverge from the badge.
- **Evidence partitioning.** The analyzers that ran are split into `supporting` / `contradicting` /
  `silent`, so "we don't know" is always attributable to named analyzers instead of an opaque shrug.
  Attribution goes through `data_model_object_id` rather than the report's GenericForeignKey, which
  would be a per-report N+1 — pinned by a query-count guard.
- **Honest absences.** A GENERIC observable (no DataModel class exists) and a job where no analyzer
  produced an evaluation both return `no evaluation` with an explicit reason, never a fabricated verdict.
- **No 12th tool.** The verdict is folded into `summarize_job`, so no new tool-routing decision is
  introduced (and no repeat of #3843). The two job chips collapse into one, "Summarize & evaluate",
  which keeps sending the already-validated `"Summarize job #{id}"` wording.
- **The headline is echoed into the prose summary** as well as carried structurally. This is a
  deliberate duplication: a live smoke against `qwen2.5:3b` showed the model reproduces the prose
  fields verbatim but paraphrases the structured object away — dropping the reliability and
  reporting contradicting analyzers as silent. A test pins the echo so it cannot silently regress.

### Reliability is truncated, not rounded

The engine averages reliability into a float and stores it through an integer column, and Django's
`IntegerField.get_prep_value` **truncates**. Avg(5, 6) = 5.5 is persisted as **5** (`suspicious`);
rounding would have made the chatbot say `malicious` while the badge said `suspicious`. The reader
truncates to match, and a test pins it by replaying `merge()` and reading the value back from the DB.

### Why the prompt word cap moves from 500 to 600

`test_prompt_under_token_limit` counts **whitespace-separated words**, not tokens. `develop` was
already at **492/500** before this PR: the guard had shrunk to 8 words of headroom and was rejecting
further rules rather than protecting the context window. At ~1.3-1.4 tokens per word here, 600 words
is roughly 850 tokens — about a tenth of Ollama's 8192 window, alongside the tool schemas and the
history. The prompt ships at 552. The docstring now states what is actually being measured.

### Multi-tenancy

`Job.get_user_events_data_model()` resolves user events with the **job owner's** visibility, while
the requester may merely be an org-mate or a viewer of a CLEAR/GREEN job. The reader therefore
exposes only the boolean `analyst_override` — never the event's author, reason or tags. Nothing in
the verdict exposes more than `get_job_details` already does (it returns each report's full body).

## Tests

- `tests/api_app/chatbot_manager/test_evaluation.py` — reader unit tests: partitioning, GENERIC,
  analyst override, boundary bucket, evidence caps, equality with the engine, and purity (no writes).
- `tests/api_app/chatbot_manager/tools/test_summarize_job.py` — envelope, no-evaluation, GENERIC, and
  the tenancy case (an invisible job is indistinguishable from a missing one — no verdict leaks).
- `tests/api_app/chatbot_manager/test_query_counts.py` — N+1 guard in the DataModel dimension.
- `tests/api_app/chatbot_manager/test_agent.py` — pins both former chip intents to the one tool.
- Frontend: `QuickActions.test.jsx` updated; the collapse is pinned by a negative assertion.

**200 chatbot tests green**, `makemigrations --check` clean, `ruff` + `prettier` clean, frontend
chat suite 65/65 green.

The narration of the structured `verdict` by `qwen2.5:3b` was measured end to end against a seeded
job (report and raw logs kept out of the repo): "summarize job #N" and "evaluate the results of job
#N" relay the verdict, the reliability, the named analyzers and the silent count correctly. On the
terse "is job #N malicious?" the model relays the verdict, the reliability and all three counts
correctly but does not enumerate the analyzer names — it never substitutes placeholder names. That
is the prompt's own "one paragraph unless the user asks for a list" rule winning over enumeration.

Refs #3895
(#3893 is merged — this branch is based on `develop` after that merge.)

### DIFF
--- a/api_app/chatbot_manager/agent/system_prompt.txt
+++ b/api_app/chatbot_manager/agent/system_prompt.txt
@@ -5,7 +5,7 @@ Answer with short, data-driven summaries. Always cite the tools you used at the 
 [Tools — when to use each]
 - search_jobs: find or list the current user's own jobs — by observable value, tag, or md5, or with no filter to list their recent jobs. Use for any question about the user's jobs: "show me my jobs", "list my jobs", "what jobs do I have?", "do I have any jobs?", "find my recent jobs", "find jobs for domain X".
 - get_job_details: full detail of ONE job by id. Use after search_jobs, or when the user asks "details of job #N".
-- summarize_job: high-level summary of ONE job (status, analyzers, findings). Use for "summarize job #N".
+- summarize_job: summary AND verdict of ONE job — status, analyzers, and whether the observable is malicious, suspicious, clean or trusted, with the analyzer evidence. Use for "summarize job #N", "evaluate the results of job #N", "is job #N malicious?".
 - list_investigations: browse investigations with optional status/name filters. Use for "show my investigations", "what investigations are open?".
 - get_investigation_tree: full tree of ONE investigation by id with all its jobs. Use after list_investigations.
 - summarize_investigation: high-level summary of ONE investigation (status, job breakdown). Use for "summarize investigation #N".
@@ -26,3 +26,4 @@ Answer with short, data-driven summaries. Always cite the tools you used at the 
 [Response style]
 - One paragraph unless the user asks for a list or breakdown.
 - Mention which tools you called at the end: "(used: search_jobs, summarize_job)".
+- With a verdict: always quote its headline verbatim first, even for a yes/no question, then name the supporting analyzers and the silent count. A disagreeing analyzer is contradicting, not silent. Never invent a score.
--- a/api_app/chatbot_manager/agent/tools/summarize_job.py
+++ b/api_app/chatbot_manager/agent/tools/summarize_job.py
@@ -3,27 +3,35 @@
 
 from langchain_core.tools import tool
 
+from api_app.chatbot_manager.evaluation import evaluate_job
+from api_app.chatbot_manager.serializers.job import SummarizeJobResultSerializer
 from api_app.choices import ReportStatus
+from api_app.models import Job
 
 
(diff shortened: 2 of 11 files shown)


mlodic (approved):

mlodic:
if there's nothing to add, I'll merge this

berardifra:
> if there's nothing to add, I'll merge this

Nothing to add, thanks.

Issue #3910, opened 2026-08-08
CHATBOT_QUEUE setting is silently overridden and can never be customized (settings import-order collision)


Aditya30ag:
## Summary

`CHATBOT_QUEUE` is defined in two different settings modules. Both are wildcard-imported into `intel_owl/settings/__init__.py`, and  `.celery` is imported *after* `.chatbot`, its hardcoded value silently overwrites the operator-configurable one. As a result, the `CHATBOT_QUEUE` environment variable has zero effect at runtime, despite the code implying it's configurable.

## Where

**`intel_owl/settings/chatbot.py`** (line 19):

```python
CHATBOT_QUEUE = secrets.get_secret("CHATBOT_QUEUE", "chatbot")
```

**`intel_owl/settings/celery.py`** (line 21):

```python
CHATBOT_QUEUE = "chatbot"
```

**`intel_owl/settings/__init__.py`** (lines 76 & 79):

```python
from .chatbot import *   # sets CHATBOT_QUEUE from the env var
...
from .celery import *    # OVERWRITES it with the hardcoded literal
```

**`intel_owl/celery.py`** (task routing, consumes the clobbered value):

```python
task_routes={
    "api_app.chatbot_manager.tasks.process_chat_message": {
        "queue": get_queue_name(settings.CHATBOT_QUEUE),  # always "chatbot"
    },
},
```

cgkol2005:
Can i assign ?

mlodic:
@berardifra can you please validate this issue? thanks :)

PR #3911, opened 2026-08-08
Fix/chatbot queue settings collision

Aditya30ag:
# Description

`CHATBOT_QUEUE` was defined in two settings modules: an operator-configurable one in `intel_owl/settings/chatbot.py` (reads from a secret/env var) and a hardcoded literal `"chatbot"` in `intel_owl/settings/celery.py`. Because `settings/__init__.py`
wildcard-imports `.celery` *after* `.chatbot`, the hardcoded literal silently overwrote the configured value, so `intel_owl/celery.py`'s `task_routes` and `CELERY_QUEUES` always used `"chatbot"` regardless of what an operator set via the `CHATBOT_QUEUE` env var.

This PR makes `settings/chatbot.py` the single source of truth: `settings/celery.py` now imports `CHATBOT_QUEUE` from it instead of redefining it, so the setting is actually configurable as the code implies. A regression test verifies `CHATBOT_QUEUE`
set via env var propagates through both `chatbot.py` and `celery.py` (and into `CELERY_QUEUES`).

No behavior change with default configuration `CHATBOT_QUEUE` still defaults to `"chatbot"` everywhere.

Issue #3910 

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not
  work as expected).
- [ ] Chore (refactoring, dependency updates, CI/CD changes, code cleanup, docs-only
  changes).

# Checklist

- [x] I have read and understood the rules about [[how to Contribute] https://intelowlproject.github.io/docs/IntelOwl/contribute/)] https://intelowlproject.github.io/docs/IntelOwl/contribute/) to
  this project
- [ ] A new plugin (analyzer, connector, visualizer, playbook, pivot or ingestor) was added or changed *(not applicable — this is a settings/config fix, no plugin involved)*
- [x] I have inserted the copyright banner at the start of the file (no new files added; existing banners in both modified files were preserved)
- [x] Please avoid adding new libraries as requirements whenever it is possible *(no new dependencies added)*
- [ ] If external libraries/packages with restrictive licenses were added, they were added in the Legal Notice section *(not applicable)*
- [ ] Linters (`Ruff`) gave 0 errors *(pending CI run — not yet executed locally in this environment)*
- [x] I have added tests for the feature/bug I solved (see `tests` folder):
  `tests/api_app/chatbot_manager/test_health.py::ChatbotQueueSettingTestCase::test_chatbot_queue_setting_configured_from_secrets`
- [ ] If the GUI has been modified *(not applicable — backend settings only, no GUI
  changes)*
- [ ] After you had submitted the PR, if `DeepSource`, `Django Doctors` or other
  third-party linters have triggered any alerts during the CI checks, I have solved
  those alerts *(pending CI)*
- [ ] I have addressed raised Copilot issues *(pending review)*
- [x] I have reviewed and verified any LLM-generated code included in this PR. **This
  PR was drafted with the assistance of an LLM (Claude)** — the root-cause analysis,
  fix, and regression test were reviewed and verified by re-tracing the settings
  import chain and reproducing the shadowing behavior in isolation before applying the
  change.

### DIFF
--- a/intel_owl/settings/celery.py
+++ b/intel_owl/settings/celery.py
@@ -5,6 +5,7 @@
 
 from ._util import get_secret
 from .aws import AWS_SQS
+from .chatbot import CHATBOT_QUEUE
 
 RESULT_BACKEND = "django-db"
 BROKER_URL = get_secret("BROKER_URL", None)
@@ -18,7 +19,6 @@ DEFAULT_QUEUE = "default"
 BROADCAST_QUEUE = "broadcast"
 CONFIG_QUEUE = "config"
 
-CHATBOT_QUEUE = "chatbot"
 
 CELERY_QUEUES = get_secret("CELERY_QUEUES", DEFAULT_QUEUE).split(",")
 for queue in [DEFAULT_QUEUE, CONFIG_QUEUE, CHATBOT_QUEUE]:
--- a/tests/api_app/chatbot_manager/test_health.py
+++ b/tests/api_app/chatbot_manager/test_health.py
@@ -88,6 +88,15 @@ class ChatbotHealthTestCase(APITestCase):
         mock_get.return_value = MagicMock(ok=True)
         self.assertFalse(chatbot_health().available)
 
+    def test_chatbot_queue_setting_configured_from_secrets(self):
+        with patch(
+            "intel_owl.secrets.get_secret",
+            side_effect=lambda name, default=None: "custom_queue" if name == "CHATBOT_QUEUE" else default,
+        ):
+            from intel_owl.settings import chatbot
+
+            self.assertEqual(chatbot.CHATBOT_QUEUE, "custom_queue")
+
 
 class ChatHealthViewTestCase(APITestCase):
     URL = "/api/chatbot/health"


berardifra:
Hi @Aditya30ag — @mlodic asked me to validate #3910, so here is the outcome of that. **The diagnosis in your issue is correct and I reproduced it**: `.celery` is wildcard-imported after `.chatbot`, the literal wins, and `CHATBOT_QUEUE` in the environment never reaches `settings.CHATBOT_QUEUE`. Nice catch — it had been silently broken since the setting was introduced.

Your fix direction works too: a single definition, whichever module it lives in, removes the import-order dependency. Three things I think are still missing before it closes the issue.

### 1. The regression test does not exercise the bug

```python
with patch("intel_owl.secrets.get_secret", side_effect=...):
    from intel_owl.settings import chatbot
    self.assertEqual(chatbot.CHATBOT_QUEUE, "custom_queue")
```

`intel_owl.settings.chatbot` is already imported while Django loads the settings, long before the test runs, so `from ... import chatbot` is a cache hit and `CHATBOT_QUEUE` keeps the value computed at boot — patching `get_secret` afterwards cannot change it. I ran the same mechanism (same module, same function, a different import-time constant) inside a running instance:

```
value after patch + re-import : qwen2.5:3b     # not the patched value
assertEqual('custom_queue') would pass?  False
```

So as written the assertion should fail rather than pass — and it went unnoticed because the CI never got to the test job (see the last section). It is also worth noting that this is why `override_settings` cannot be used here either: it would assign the value the test is trying to prove is computed.

What does work is booting a fresh interpreter with the variable set and asserting on what comes out, which also lets you cover the part that matters most — that the value reaches the Celery route, not just the settings module:

```python
env = {**os.environ, "CHATBOT_QUEUE": "my_custom_queue", "DJANGO_SETTINGS_MODULE": "intel_owl.settings"}
subprocess.run([sys.executable, "-c", probe], env=env, ...)
# probe: django.setup(); report settings.CHATBOT_QUEUE, settings.CELERY_QUEUES,
#        app.conf.task_routes[process_chat_message.name]["queue"], and the declared task_queues
```

### 2. The consumer side is still hardcoded

`docker/entrypoints/celery_chatbot.sh` pins the queue in its `-Q` list:

```sh
if [ "$AWS_SQS" = "True" ]; then
  queues="chatbot.fifo,config.fifo"
else
  queues="chatbot,broadcast,config"
fi
```

With the settings fixed but this untouched, `CHATBOT_QUEUE=my_queue` makes Django publish chat turns to `my_queue` while the only worker keeps draining `chatbot` — the turns are queued and never consumed, and `chatbot_health()` reports the queue as unserved. So the setting becomes configurable and immediately broken when configured. `${CHATBOT_QUEUE:-chatbot}` on both branches fixes it (`:-` also covers a blank value, which matters if the Python side falls back on blank too).

### 3. The variable is undocumented

`docker/env_file_app_template` lists `OLLAMA_*`, `CHATBOT_MESSAGE_RETENTION_DAYS`, `CHATBOT_RATE_LIMIT*` but not `CHATBOT_QUEUE`, so operators have no way to discover it.

### Minor

Where the single definition lives is a matter of taste, but there is an argument for `settings/celery.py` rather than `settings/chatbot.py`: it already owns `DEFAULT_QUEUE` / `BROADCAST_QUEUE` / `CONFIG_QUEUE` and the `CELERY_QUEUES` loop, it is marked `# this module must run before the others`, and it imports only `._util` and `.aws` — importing `.chatbot` from it pulls `.cache` into the module that must run first. Not a blocker either way.

### About the CI failure on this PR

Not your fault, and not related to your change: the `linters` job dies in `Perform ShellCheck Analysis` with exit code 127 because `https://raw.githubusercontent.com/CICDToolbox/shellcheck/master/pipeline.sh` now returns 404, and `curl -s` (no `-f`) pipes the `404: Not Found` body into `bash`, which tries to run it as a script. It breaks every PR right now; #3912 / #3913 track the fix. Since `backend-tests` has `needs: linters`, no test has run on this PR at all.

### Disclosure

Before I saw this PR I had already prepared #3919, which fixes the same issue and covers the three points above. I have moved it to draft — I would rather this one landed, and I am happy to hand the pieces over here. @mlodic, your call on which to carry forward; if you prefer #3911, the entrypoint and template changes can be lifted from #3919 (or I can push them here if @Aditya30ag wants).

Aditya30ag:
@berardifra I had no objection to your PR being merged. The purpose was only to resolve the bug.

mlodic:
thanks for the communication, we'll supersede this PR with the new one from @berardifra

Issue #3909, opened 2026-08-08
chatbot: the anti-placeholder prompt rule is probabilistic, not a guarantee


berardifra:
### What

The `[Rules]` line added in #3881 —

```
- Copy analyzer, playbook and job names verbatim from tool results. Never replace them with
  placeholders or bracketed labels like [Analyzer 1] or [Playbook A] — write the real names exactly
  as the tool returned them.
```

— holds in the large majority of answers but is **not a guarantee**. While measuring chatbot
narration for #3898 I recorded one answer that emitted exactly what the rule forbids:

> The playbooks that can analyze this type of observable are:
> - **[Playbook X]**
> - **[Playbook Y]**

### Context of the observation

- model `qwen2.5:3b`, real Ollama, production invocation path (in-process inside `intelowl_uwsgi`);
- one occurrence in roughly 40 measured answers, i.e. ~2-3%;
- it appeared only under a stricter prompt variant that put extra pressure on naming
  (`list the supporting analyzers by name — never just their number`), which was measured and then
  **discarded** — the shipped prompt does not contain it;
- the tool payload involved has no playbook field at all, so the names were not being copied from
  anywhere: the model invented the placeholders outright.

### Why it is worth an issue

The rule was added specifically to stop the bracketed-placeholder output reported on Discord, and it
does work in the ordinary case. But a rule expressed only in the system prompt is probabilistic by
construction: under competing instructions the model can still violate it, and the failure is silent
— a user sees a plausible-looking answer with fabricated names.

### Possible directions (not a proposal, just what the observation suggests)

- a post-generation check that rejects or retries an answer containing `[...]`-style placeholders
  where a tool result provided real names — a guard rather than an instruction;
- or accepting it as a known limitation and documenting it, since the blast radius is a cosmetic
  wrong name rather than a wrong verdict.

### Related

Observed while working on #3898 (chatbot job verdict). **Not caused by it** — that PR's payload
contains no playbook data. Full measurement notes are in the narration-smoke report kept with the
GSoC working material; happy to paste the raw runs here if useful.

PR #3920, opened 2026-08-10
[GSoC 2026] chatbot: post-generation guard against fabricated placeholder names

berardifra:
# Description

`Refs #3909`

The `[Rules]` line added in #3881 tells the model to copy analyzer, playbook and job names verbatim
and never emit bracketed labels like `[Analyzer 1]`. It works in the large majority of answers, but a
rule expressed only in the system prompt is probabilistic by construction — and the failure is
silent: the user sees a plausible answer with fabricated names.

This adds the deterministic backstop @mlodic asked for in #3909: an LLM-free check that runs on the
final answer of every chat turn, on **both** answer paths, and annotates an answer whose bracketed
names look like placeholders instead of letting it through unremarked.

## What it does — and deliberately does not

On detection it logs one warning and appends a single line to the answer:

> _Note: this answer contains bracketed names that look like placeholders rather than real IntelOwl
> names. Treat them as unverified._

It does **not** edit the model's prose and does **not** re-run the model. Two reasons, both
structural:

- **There is no correct name to substitute.** In the recorded failures the tool payload contained no
  playbook data at all — the names were invented outright, not degraded from real ones.
- **A retry would change nothing.** The agent runs at `temperature=0`, so re-running the same input
  reproduces the same answer; any "retry" would have to be a corrective rewrite, i.e. a second
  inference on a CPU-bound local model, which can itself hallucinate.

Removing the offending spans was considered and rejected: in the real case it would leave empty list
bullets under an orphan lead-in (*"The playbooks that can analyze this observable are:"*).

## Honest scope of the claim

Neither archived occurrence was produced by the prompt exactly as it ships today — one came from a
prompt variant that was measured and discarded, the other predates the rule #3881 added. So this is
**defense-in-depth against a measured model tendency, not a fix for a bug reproducible on demand.**
The guard catches the failure; it does not prevent it.

The tendency is not accidental, though: both occurrences belong to the same "playbook" confabulation
family recorded during the #3898 narration measurement (in 4 of 9 answers the model introduced the
analyzer list as *"the playbooks used for this job"*, a word no payload contains). Its root cause —
`system_prompt.txt` advertising `recommend_playbook` while #3881's rule repeats "playbook" — is still
in the shipped prompt.

## The detection criterion, and why it is not `\[.*\]`

A false positive would stamp a warning onto a correct answer, so the criterion is two conditions, not
one regex:

1. a single-line `[...]` span **not** followed by `(` — that lookahead excludes markdown links
   `[text](url)`, which the chat panel renders as real anchors;
2. whose content **starts with** one of the entity nouns the prompt rule protects
   (`analyzer|playbook|job|investigation|observable`).

The start anchor is what keeps out the things that legitimately carry brackets in a threat-intel
product: defanged indicators (`example[.]com`, `1.1.1[.]1`, `hxxp[:]//`), GFM task-list boxes
(`- [x]`), footnote markers (`[1]`), and JSON echoes of a tool result. Every one of those is a test.

Deliberately missed shapes (`[X]`, `[TBD]`, `[insert name]`): precision is preferred over coverage,
and the prompt rule remains the first line of defense. A second family gets added when a real
occurrence justifies it — exactly how this one came to be.

## Measurements

**Offline, 43 archived answers.** The detector was run over every answer archived during the
2026-08-03 narration campaign plus the 9 live answers from the smoke below:

```
scanned 43 archived answer files
  dev3-run2.txt:  ['[Playbook X]', '[Playbook Y]']
  smoke-run4.txt: ['[Playbook names will be listed here if available]']
files flagged: 2
```

Exactly the two known occurrences, with the exact spans, and **zero false positives on the other
41** — real model output containing analyzer names, job ids, verdict headlines and markdown.

Note the shape of the second one: a whole sentence used as a name slot. A detector tuned only on
`[Label N]` would have missed it.

**Live, 3 runs × 3 phrasings.** Real Ollama, `qwen2.5:3b`, production invocation path: **9 answers,
0 guard firings.** Narration is unchanged from the baseline, which is expected by construction — this
PR touches no prompt and no tool.

## What the user sees

Screenshot and a 35s clip below.

The guard fires on ~2-3% of answers and not under the shipped prompt, so it cannot be triggered on
demand for a recording: the conversation in the first scene is **seeded**, and both answer texts in
it are real model output taken verbatim from the measurement archive (the flagged one is the
`[Playbook X]` / `[Playbook Y]` answer from the 2026-08-03 campaign). The rendering, the stored
content and the notice are exactly what a user would get; only the turn that produced them is
simulated.

The second scene is **not** seeded: it is a real turn against live Ollama, and it is deliberately a
name-dense one — *"What analyzers can I use for a domain?"* returns a numbered list of real analyzer
names (`ApiVoid`, `BBOT`, `CheckDMARC`, `Classic_DNS`, `CleanBrowsing_Malicious_Detector`, …). That
is the answer shape a naive `\[.*\]` detector would trip over. No note is added; the recorder asserts
it (`LIVE_TURN_NOTICES: 0`, on each of the three runs recorded).

## Where it hooks in

- `tasks.py` — the Celery/WebSocket turn, before the message is persisted and `chat.end` is emitted.
  The tokens have already streamed by then, but the frontend's `applyEnd` discards the streamed
  buffer and renders `chat.end`'s content, so the guarded text is both what the user reads and what
  is stored.
- `views.py` — the synchronous REST fallback, at the same point in the flow.
- `agent/memory.py` — `DjangoChatMessageHistory` strips the notice when replaying prior turns. The
  notice is an annotation for the user, not model output; without this it would re-enter the prompt,
  where a 3B model may start imitating it.

## Out of scope

- **No prompt change.** Four prompt attempts were measured on 2026-08-03; each traded one failure for
  another, and the fourth actively produced this very bug. Editing the prompt would also force a full
  re-measurement of the narration gate.
- **No structured field** on `ChatMessage` / `EndEvent`. A dedicated flag would allow a warning chip
  in the UI instead of a text line, at the cost of a migration, a serializer change and frontend
  work. Happy to do it as a follow-up if you prefer that treatment.

## Type of change

# Checklist

- [x] Please avoid adding new libraries as requirements whenever it is possible — no new dependency; the new module is stdlib-only (`logging`, `re`)
- [x] I have added tests for the feature/bug I solved. All the tests (new and old ones) gave 0 errors — `tests.api_app.chatbot_manager` 223/223 green
- [x] If the GUI has been modified: no frontend file is touched; the change is visible in the panel because the answer text itself changes, so a screenshot is provided
- [x] I have reviewed and verified any LLM-generated code included in this PR. **LLM usage: yes** — Claude Code was used for the implementation, the detector's test corpus and this description; every line was reviewed, and the criterion was validated against 43 archived real answers and 9 live ones rather than against generated examples.

### DIFF
--- a/api_app/chatbot_manager/agent/memory.py
+++ b/api_app/chatbot_manager/agent/memory.py
@@ -4,6 +4,8 @@
 from langchain_core.chat_history import BaseChatMessageHistory
 from langchain_core.messages import AIMessage, BaseMessage, HumanMessage
 
+from api_app.chatbot_manager.placeholder_guard import strip_notice
+
 
 class DjangoChatMessageHistory(BaseChatMessageHistory):
     """LangChain chat history backed by the Django ChatMessage model.
@@ -27,7 +29,10 @@ class DjangoChatMessageHistory(BaseChatMessageHistory):
             if msg.role == ChatMessage.Role.USER:
                 result.append(HumanMessage(content=msg.content))
             else:
-                result.append(AIMessage(content=msg.content))
+                # The placeholder notice is an annotation added for the user after generation, not
+                # something the model wrote. Replaying it would put it in the model's context, where
+                # a small model may start imitating the format.
+                result.append(AIMessage(content=strip_notice(msg.content)))
         return result
 
     def add_message(self, message: BaseMessage) -> None:
--- a/api_app/chatbot_manager/placeholder_guard.py
+++ b/api_app/chatbot_manager/placeholder_guard.py
@@ -1,0 +1,101 @@
+# This file is a part of IntelOwl https://github.com/intelowlproject/IntelOwl
+# See the file 'LICENSE' for copying permission.
+
+"""Post-generation guard against fabricated bracketed names in a chat answer.
+
+The system prompt tells the model to copy analyzer, playbook and job names verbatim from tool
+results and never emit bracketed labels like ``[Analyzer 1]``. A rule that lives only in the prompt
+is probabilistic by construction: measured answers show the model can still invent a whole section
+(typically a playbook list) and, having no names to copy, fill it with placeholders. This module is
+the deterministic backstop, and it never calls the LLM.
+
+It detects and annotates; it deliberately does not repair. In the observed failures the tool payload
+carried no playbook data at all, so there is no real name to substitute, and rewriting would mean a
+second inference on a local CPU-bound model — a plain retry would not even change the answer, since
+the agent runs at temperature 0.
+
+Precision is preferred over coverage: a false positive would stamp a warning onto a correct answer,
+which is worse than missing a rare shape. The criterion is therefore anchored to the entity nouns
+the prompt rule protects, and the prompt rule remains the first line of defense for the shapes this
+misses (``[X]``, ``[TBD]``, ``[insert name]``).
+"""
+
+import logging
+import re
+
+logger = logging.getLogger(__name__)
+
+# What the notice claims is exactly what the criterion below checks: a bracketed span *shaped* like
+# a name slot. The guard never inspects the tool results, so it cannot assert the name is ungrounded
+# — only that it does not look like a real IntelOwl name.
+PLACEHOLDER_NOTICE = (
+    "_Note: this answer contains bracketed names that look like placeholders rather than real "
+    "IntelOwl names. Treat them as unverified._"
+)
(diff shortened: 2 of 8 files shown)


berardifra:
placeholder-guard


https://github.com/user-attachments/assets/5124ed76-d413-4856-bce1-8fd88e195c9b

mlodic (approved):

PR #3972, opened 2026-08-28
fix: update Job status to FAILED when job_pipeline raises an exception

Aditya30ag:
## Description

When `job.execute()` raises inside `job_pipeline`, the `except` block only marked individual plugin reports as failed it never updated the `Job` row itself.
Since `execute()` sets `job.status = RUNNING` as its first action, and the Celery chain (including `job_set_final_status`) never dispatches on exception, the job gets permanently stuck in `RUNNING`.

**Consequences:**
- Frontend shows an infinite spinner no WebSocket notification is ever sent
- `remove_old_jobs` never cleans the job `finished_analysis_time` stays `NULL`
- Parent `Investigation` status is never updated
- Exception message is silently discarded

**Fix:** In the `except` block, immediately set `job.status = FAILED`, `job.finished_analysis_time = now()`, append the error, persist with `job.save(update_fields=[...])`, push a WebSocket notification, and propagate failure to the parent `Investigation` if one exists.
This matches the pattern already used in `run_plugin` (line 295 of `tasks.py`).

## Type of change

Issue #3653

## Checklist

- [ ] A new plugin (analyzer, connector, visualizer, playbook, pivot or ingestor) was added or changed — **not applicable**
- [x] No new libraries were added as requirements
- [x] I have added tests for the bug I solved (see `tests/` folder). All tests gave 0 errors
- [x] I have reviewed and verified any LLM-generated code included in this PR. I have explicitly stated that I used an LLM (Claude) to assist in identifying the bug, writing the fix, and the verification script. All output was manually reviewed and validated against the real source.

### DIFF
--- a/intel_owl/tasks.py
+++ b/intel_owl/tasks.py
@@ -248,12 +248,18 @@ def job_pipeline(
     job_id: int,
 ):
     from api_app.models import Job
+    from api_app.websocket import JobConsumer
 
     job = Job.objects.get(pk=job_id)
     try:
         job.execute()
     except Exception as e:
         logger.exception(e)
+        job.status = Job.STATUSES.FAILED.value
+        max_error_len = Job._meta.get_field("errors").base_field.max_length
+        job.errors = list(job.errors or []) + [str(e)[:max_error_len]]
+        job.finished_analysis_time = now()
+        job.save(update_fields=["status", "errors", "finished_analysis_time"])
         for report in (
             list(job.analyzerreports.all())
             + list(job.connectorreports.all())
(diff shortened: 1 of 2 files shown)


Aditya30ag:
Hey @berardifra,
Could u please review the changes?

berardifra:
Hi @Aditya30ag, thanks for tagging me.

I opened #3976 with a small branch against `develop` for the same issue, keeping the fix focused and adding a regression test for the `job.execute()` exception path.

I also referenced your PR in the description so the context is preserved.

Aditya30ag:
@berardifra 
Thank you for taking the time to look into this and for opening a PR with a regression test. I really appreciate it.

